PR #5, opened 2018-07-24
Add support to get plans and plan logs

svanharmelen:


plan.go
```diff
@@ -0,0 +1,92 @@
...
+	err = s.client.do(ctx, req, p)
+	if err != nil {
+		return nil, err
+	}
+
+	return p, nil
+}
+
+// Logs retrieves the (partial) logs of a plan.
+func (s *Plans) Logs(ctx context.Context, url string) ([]byte, error) {
```
ryanuber: This seems like it should be a method hung off of the `Plan` struct. The URL is a fully-qualified URL with the protocol, host, etc. all populated in a single string. This way you never actually have to pass in a URL.

plan.go
```diff
@@ -0,0 +1,92 @@
...
+	req, err := s.client.newRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "text/plain")
+
+	var buf bytes.Buffer
+	err = s.client.do(ctx, req, &buf)
```
ryanuber: I think it would be nice if we could return an `io.Reader` here so that the logs could be streamed in. Some plan logs can be very verbose depending on if the user enabled debug or trace logs, and the number of resources in the config. Also wondering about log streaming. It could be nice to use the `io.Reader` to implement this (not necessarily in this PR of course), because we could basically just return the reader, and the user could continue to call `Read()` on it until they get an EOF (which we would return when they reached the end of the log).

plan.go
svanharmelen: That is true, but I like the fact that we are now reusing the existing HTTP client and logic. So (even if the call doesn't require that at the moment) it is actually making an authenticated call using the same HTTP client as was used when creating the API client.

So if anyone is using the client with a custom HTTP client, it will just work as expected (say when using PTFE). Of course we could add the client to the `Plan` struct when returning a `Plan`, but that feels a bit nasty. If we want to make the call from the `Plan` struct, it should contain all the logic to do so IMO.

plan.go
svanharmelen: I played with a couple of options, but it doesn't make the code and using the API easier. The main reason for that is that currently the API _doesn't_ stream the logs. So even if we return an `io.Reader`, you will still need to `ioutil.ReadAll()` to get all the bytes and then trim the bytes that you already showed in the previous iteration (think about "streaming" the logs to the console here).

The best I could come up with is this one:

```go
// Logs retrieves the (partial) logs of a plan.
func (p *Plan) Logs(ctx context.Context) (io.Reader, error) {
	// Create an io.Pipe for piping the logs to a reader.
	r, w := io.Pipe()

	// Close the writer and return if there isn't a log URL.
	if p.LogReadURL == "" {
		w.Close()
		return r, nil
	}

	// Create the request.
	req, err := http.NewRequest("GET", p.LogReadURL, nil)
	if err != nil {
		return nil, err
	}
	req = req.WithContext(ctx)

	// Retrieve the available logs.
	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return nil, err
	}

	// Basic response checking.
	if err := checkResponseCode(resp); err != nil {
		resp.Body.Close()
		return nil, err
	}

	go func() {
		defer w.Close()
		defer resp.Body.Close()
		io.Copy(w, resp.Body)
	}()

	return r, nil
}
```

This looks pretty OK, but with the downside I mentioned before. The application using this will need to diff (trim) the logs before knowing what to display. This makes the reader a bit useless.

So I'm thinking lets keep and merge the code as is for this API call. And then when we get a real streaming endpoint, we can keep this endpoint for retrieving the complete logs and add new `Logs()` method on the `Plan` struct that will use the new streaming capabilities.

plan.go
svanharmelen: Hmm.. I just realized we could just add the HTTP client from the `Client` to the `Plan` (instead of adding the whole client). And then in the `Logs()` method we could check if we have a custom HTTP client and if not just use `http.DefaultClient` (for when someone created a `&Plan{}` manually).

So that would mitigate my earlier comment. Yet I still think it would be a better match to keep this method as is and create a new streaming version on the `Plan` struct when we actually have a streaming API. 

What do you think?

ryanuber: Ah, let me explain!

So, the URL to the plan logs is a totally separate system from TFE itself - Archivist. TFE is able to generate URL's directly to the Archivist service. When requesting Archivist URL's, though, we should *not* be sending the TFE bearer token in the request. It isn't necessary, and we shouldn't send authentication tokens between services when it is not necessary for security reasons.

The log URL is meant to be called directly, with no further headers or other request details. It's fully qualified, as I noted earlier, so the client should only need to make a request to the exact URL given without any other knowledge of the Archivist service.

Fair point about embedding the client in the `Plan` - I'm ok with not doing that for the reasons you mentioned.

ryanuber: I do think, however, that it might be useful to just pass a `Plan` into the `Logs` method in this case, like `Logs(ctx, plan)` (instead of just the URL). That way we can leverage additional details of the plan for things like streaming.

plan.go
ryanuber: We actually do already have a streaming log API! It's what we use in the TFE web interface. You request the log in chunks, and track the last byte index you have seen. That allows you to iterate through the log without re-requesting the whole thing every time. See the streaming options [here](https://github.com/hashicorp/archivist#get-v1objectobject). My thinking was that you could still just call `tfe.Plans.Logs(ctx, plan)`, and then just read until `EOF` to get the full log out. Internally If you use something like `io.Copy` this would already be handled for you, making it extremely easy to stream the log. We don't have to do the full streaming API immediately but I think returning the whole log as an in-memory byte slice is not quite right for a relatively unpredictable log size.

plan.go
svanharmelen: Thanks for the explanation, but I knew all that already :wink: (that's why I said `even if the call doesn't require that at the moment`).

> Fair point about embedding the client in the Plan - I'm ok with not doing that for the reasons you mentioned.

But we do need the `http.Client` so that any custom client will be used instead of a default one (in case anyone passed in a custom `http.Client` with a proxy for example).

> I do think, however, that it might be useful to just pass a Plan into the Logs method in this case, like Logs(ctx, plan) (instead of just the URL). That way we can leverage additional details of the plan for things like streaming.

What additional details do we need for the stream other then the URL?

plan.go
svanharmelen: > We actually do already have a streaming log API!

Ah... That I didn't know! That changes things of course :) Let me take a look at it next week, as I like to do it "first time right". So I prefer to then go ahead and move the `Logs()` method to the `Plan` struct and use the streaming API.

I'll check it out and make an update on the PR 👍

plan.go
svanharmelen: See the other discussion... If we do have a streaming API (which I didn't knew), then I also prefer to move the `Logs()` method. So this discussion becomes irrelevant :)

ryanuber: > But we do need the http.Client so that any custom client will be used instead of a default one (in case anyone passed in a custom http.Client with a proxy for example).

Right - so `Plans` will have the client embedded by way of the `client.http` and related methods you have already built. `Plan` remains a simple data-only struct. You already have it this way, and I totally get why we need the possibly custom HTTP client.

> What additional details do we need for the stream other then the URL?

I think we'll want the plan ID so that we can check if the plan has ended. Then we'll be able to either keep checking for new logs, or stop the streaming and return.

svanharmelen:
Updated the `Logs() method to give users of the method a true streaming experience...

plan.go
```diff
@@ -0,0 +1,172 @@
...
+// LogReader implements io.Reader for streaming plan logs.
+type LogReader struct {
+	client *Client
+	ctx    context.Context
+	offset int64
+	planID string
+}
+
+func (r *LogReader) Read(l []byte) (int, error) {
+	for {
```
ryanuber: You don't need this outer `for` loop - you can just put the `select` out here to block (see below comments).

plan.go
```diff
@@ -0,0 +1,172 @@
...
+			// before checking if there is a new chunk available or that the
+			// plan is finished and we are done reading all chunks.
+			if resp.ContentLength == 0 {
+				resp.Body.Close()
+
+				switch p.Status {
+				case PlanCanceled, PlanErrored, PlanFinished:
+					return 0, io.EOF
+				default:
+					continue
```
ryanuber: You should just be able to return `0, nil` for this case, rather than making `Read` block until more data arrives. The next `Read()` will sleep 500ms anyways. This would clear up some of the body closing logic also.

plan.go
```diff
@@ -0,0 +1,172 @@
...
+	offset int64
+	planID string
+}
+
+func (r *LogReader) Read(l []byte) (int, error) {
+	for {
+		select {
+		case <-r.ctx.Done():
+			return 0, r.ctx.Err()
+		case <-time.After(500 * time.Millisecond):
```
ryanuber: Two things here:

1) This is at the start of the `Read()`, which is a little odd in that it creates an initial "start delay". It seems like this would be better to place after the actual request in case the user wants to do a one-shot `Read()` instead of streaming. What do you think?

2) 500ms is pretty aggressive as a constant delay for streaming - I think the shortest polling interval we have used on the UI side is 1s, and we back off to 3s if no logs are coming in to shed some load. This obviously affects the UX in that users could possibly see more delay, but what do you think about taking a similar approach to guard ourselves from a high volume of API requests?

plan.go
```diff
@@ -0,0 +1,172 @@
...
+			p, err := r.client.Plans.Read(r.ctx, r.planID)
```
ryanuber: To me this feels like something we should probably move out of the "hot path" - this adds significant time to the request and load on TFE to check every time we request the logs from Archivist. I think we could do something better like track the number of requests and just check on every 3rd or 5th one or so.

Also, we should move this below the actual request for the logs, because if that fails then we won't use this result anyways.

svanharmelen:
Updated following our discussion. Please have another look...

plan.go
```diff
@@ -0,0 +1,194 @@
...
+	CreatedAt    time.Time `json:"created-at"`
+	ErroredAt    time.Time `json:"errored-at"`
+	FinishedAt   time.Time `json:"finished-at"`
+	MFAWaitingAt time.Time `json:"mfa_waiting-at"`
+	PendingAt    time.Time `json:"pending-at"`
+	QueuedAt     time.Time `json:"queued-at"`
+	RunningAt    time.Time `json:"running-at"`
+}
+
+// Read a single run by its ID.
```
ryanuber: *plan

svanharmelen: I’ll make sure this is corrected in PR #6 (which updates/fixes some other comments as well)...

PR #6, opened 2018-07-31
Replace structs with interfaces for testabiity

svanharmelen:
This approach allows people using this package to create a mock for a specific service/endpoint.


configuration_version.go
```diff
@@ -11,12 +11,23 @@ import (
...
+// methods that the Terraform Enterprise API supports.
 //
 // TFE API docs:
 // https://www.terraform.io/docs/enterprise/api/configuration-versions.html
-type ConfigurationVersions struct {
+type ConfigurationVersions interface {
+	List(context.Context, string, ConfigurationVersionListOptions) ([]*ConfigurationVersion, error)
+	Create(context.Context, string, ConfigurationVersionCreateOptions) (*ConfigurationVersion, error)
+	Read(context.Context, string) (*ConfigurationVersion, error)
+	Upload(context.Context, string, string) error
```
ryanuber: I think defining the interface without naming the args will make the godoc much less useful. What are the two strings this method wants? It's hard to tell without the names.

We are also losing all of the method godocs, because they now only exist on private structs. I think it's valuable to preserve that in the interface so that users can see the godoc properly.

svanharmelen: Good points, will update that...

svanharmelen:
Added the comments and named the parameters as discussed...

PR #13, opened 2018-09-07
adding delete function

djaboxx:


oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+	}
+	return nil
+}
+
+func (s *oAuthClients) Delete(ctx context.Context, options OAuthCLientDestroyOptions) (*OAuthClient, error) {
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("oauth-clients/%s", url.QueryEscape(*options.CLIENT_ID))
```
svanharmelen: According to the [docs](https://www.terraform.io/docs/enterprise/api/oauth-clients.html#destroy-an-oauth-client), this URL should look like this: `organizations/:organization_name/oauth-clients/:id`

oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+
+	req, err := s.client.newRequest("DELETE", u, s)
```
svanharmelen: Why are you passing `s` to the `newRequest` call? You should either pass an options struct or `nil`.

oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+	if err != nil {
+		return nil, err
+	}
+
+	oc := &OAuthClient{}
```
svanharmelen: You are creating a new empty value and then returning that directly to the caller without unmarshalling any response data. And since this API call does not return any data, we should just omit this.

oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+	err = s.client.do(ctx, req, nil)
+	if err != nil && err.Error() != "resource not found" {
+		return nil, err
+	}
+
+	if err != nil && err.Error() == "resource not found" {
+		return oc, nil
+	}
+
+	return oc, nil
```
svanharmelen: Instead of these custom error checks you can/should just return the result of the `do` call here:

```go
return s.client.do(ctx, req, nil)
```

oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+	u := fmt.Sprintf("oauth-clients/%s", url.QueryEscape(*options.CLIENT_ID))
+
+	req, err := s.client.newRequest("DELETE", u, s)
+	if err != nil {
+		return nil, err
+	}
+
+	oc := &OAuthClient{}
+	err = s.client.do(ctx, req, nil)
+	if err != nil && err.Error() != "resource not found" {
```
svanharmelen: Why would you specifically check for `err.Error() != "resource not found"`?

oauth_client.go
```diff
@@ -125,3 +126,40 @@ func (s *oAuthClients) Create(ctx context.Context, organization string, options
...
+		return nil, err
+	}
+
+	if err != nil && err.Error() == "resource not found" {
```
svanharmelen: I'm afraid I don't think I understand these error checks. Can you share your idea behind this?

svanharmelen:
I'm going to close this PR in favor of PR #15 which (among other changes) adds the endpoint that this PR intended to add as well.

PR #17, opened 2018-09-14
add delete method to oauth client and some example code

scottwinkler:
Add delete method to oauth client. Need this so the tfe provider can have the oauth client as a proper resource.

Closed previous pull request because i was on the wrong branch

### DIFF
--- a/examples/oauth-client/main.go
+++ b/examples/oauth-client/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"context"
+	"log"
+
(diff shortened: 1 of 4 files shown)


svanharmelen:
Hi @scottwinkler! Thanks for your PR, but PR #15 already adds this method, together with a few other missing methods.

So I’m verry sorry, but I’m going to close this PR in favour of PR #15, which I hope to merge later today.

Once this is merged we will update the TFE provider as well, so it uses the new methods.

PR #18, opened 2018-09-14
Add registry module publish/delete

scottwinkler:
This is kind of a tricky one so I did my best. Added support for being able to create registry modules and delete them. Did not implement list, or search or any of those as I figure those could come later. Not sure how to get the tests to run on my machine, and the tests won't work anyways until pull request #15 is approved as this resource requires a vcs connection to properly create. I created an example, and verified that that works. There is an idiosyncrasy with the api itself, in that the relationships for organization returned by the registry-module uses an id for organization that isn't the same as the actual id returned from the organizations endpoint. Also I don't know all the valid statuses so I just included the two I was able to figure out.


svanharmelen:
Thanks @scottwinkler! We actually have a version with more or less the exact same code, but there isn't a PR for that yet as it's not clear how we want to handle all the other registry calls.

The other registry calls have a slightly different format (they don't adhere to the [JSONAPI specifications](http://jsonapi.org/)) so we're still thinking/discussing what the best (i.e. the cleanest and most robust) solution would be.

To that end I'm going to hold of on reviewing this PR until we decided how to incorporate the registry endpoints. Will keep you posted on any progress of course! Thanks!

scottwinkler:
Sounds good. I actually dont think getting the other registry calls done is that important, at least from a customer perspective. What would be more interesting is creating a function that publishes to all the organizations since global modules arent really a thing. What I am thinking is it would list all organizations, try to get an oauth token, and then use that to publish the module. What do you think?

scottwinkler:
Also just having a create and delete would be enough to get the tfe provider updated, if all the attributes are ForceNew and if it doesnt actually do a read to sync the state. Later, when the other registry calls get done, it could implement an actual read. Currently I have to use a script to automate creating modules which isnt extremely elegant right now.

lafentres:
Hi there! 👋 My PR #128 should address this. There are still some endpoints we can't support in go-tfe yet but what I've implemented should cover the create, read, and delete cases.

lafentres:
Hello again 👋 just wanted to let you know that support for this has been merged and released in v0.10.0 :]
I'm going to go ahead and close this now that the release is done.

PR #22, opened 2018-10-05
Add force-cancel method for Runs

ryanuber:
Force-cancel is now [available in TFE](https://www.terraform.io/docs/enterprise/api/run.html#forcefully-cancel-a-run)! This adds a function and some additional fields to support calling the force-cancel operation from this client.

### DIFF
--- a/run_test.go
+++ b/run_test.go
@@ -222,7 +222,7 @@ func TestRunsForceCancel(t *testing.T) {
 				break
 			}
 
-			if i > 5 {
+			if i > 30 {
 				t.Fatal("Timeout waiting for run to be canceled")
 			}
 
--- a/run_test.go
+++ b/run_test.go
@@ -210,14 +210,24 @@ func TestRunsForceCancel(t *testing.T) {
 	t.Run("after a normal cancel", func(t *testing.T) {
 		// Request the normal cancel
 		err := client.Runs.Cancel(ctx, rTest.ID, RunCancelOptions{})
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
-		// Allow time for the async cancel request to operate.
-		time.Sleep(5 * time.Second)
+		for i := 1; ; i++ {
+			// Refresh the view of the run
+			rTest, err = client.Runs.Read(ctx, rTest.ID)
+			require.NoError(t, err)
 
-		// Refresh the view of the run
-		rTest, err = client.Runs.Read(ctx, rTest.ID)
-		assert.NoError(t, err)
+			// Check if the timestamp is present.
+			if !rTest.ForceCancelAvailableAt.IsZero() {
+				break
+			}
+
+			if i > 5 {
+				t.Fatal("Timeout waiting for run to be canceled")
+			}
+
+			time.Sleep(time.Second)
+		}
 
 		t.Run("force-cancel-available-at timestamp is present", func(t *testing.T) {
 			assert.True(t, rTest.ForceCancelAvailableAt.After(time.Now()))
--- a/run_test.go
+++ b/run_test.go
@@ -3,6 +3,7 @@ package tfe
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


run_test.go
```diff
@@ -184,6 +185,64 @@ func TestRunsCancel(t *testing.T) {
...
+	})
+
+	t.Run("user is allowed to force-cancel", func(t *testing.T) {
+		assert.True(t, rTest.Permissions.CanForceCancel)
+	})
+
+	t.Run("after a normal cancel", func(t *testing.T) {
+		// Request the normal cancel
+		err := client.Runs.Cancel(ctx, rTest.ID, RunCancelOptions{})
+		assert.NoError(t, err)
```
svanharmelen: I would change this to `require.NoError` as the rest of the test will not work/succeed if this call fails.

run_test.go
```diff
@@ -184,6 +185,64 @@ func TestRunsCancel(t *testing.T) {
...
+
+		// Allow time for the async cancel request to operate.
+		time.Sleep(5 * time.Second)
+
+		// Refresh the view of the run
+		rTest, err = client.Runs.Read(ctx, rTest.ID)
+		assert.NoError(t, err)
```
svanharmelen: The same for this call. Think this also should be `require.NoError`. But if you choose in use a loop to check the actual state before moving on, this call would not be needed here anymore I guess.

run_test.go
```diff
@@ -184,6 +185,64 @@ func TestRunsCancel(t *testing.T) {
...
+		assert.True(t, rTest.Permissions.CanForceCancel)
+	})
+
+	t.Run("after a normal cancel", func(t *testing.T) {
+		// Request the normal cancel
+		err := client.Runs.Cancel(ctx, rTest.ID, RunCancelOptions{})
+		assert.NoError(t, err)
+
+		// Allow time for the async cancel request to operate.
+		time.Sleep(5 * time.Second)
```
svanharmelen: This looks like something that could end up causing flaky tests. In all the `helper_test.go` cases where I would need a certain state I just created a small loop and checked for the actual state I needed before moving on. Its a few more lines of code but think it would be a bit more robust to use that pattern here as well: https://github.com/hashicorp/go-tfe/blob/master/helper_test.go#L302-L318

run_test.go
ryanuber: Good idea. I updated with a loop as suggested below, but I'll keep the error check, because I don't expect errors. The loop is meant to ride out waiting for a certain state to be achieved, not paper over unexpected errors.

run_test.go
```diff
@@ -210,14 +210,24 @@ func TestRunsForceCancel(t *testing.T) {
...
+			rTest, err = client.Runs.Read(ctx, rTest.ID)
+			require.NoError(t, err)
 
-		// Refresh the view of the run
-		rTest, err = client.Runs.Read(ctx, rTest.ID)
-		assert.NoError(t, err)
+			// Check if the timestamp is present.
+			if !rTest.ForceCancelAvailableAt.IsZero() {
+				break
+			}
```
svanharmelen: What I think we should do here is `if rTest.Status == tfe.RunCanceled` as that means the async task is completed. The the next test will then confirm that a ForceCancelAvailableAt is properly set now the run is canceled.

run_test.go
```diff
@@ -210,14 +210,24 @@ func TestRunsForceCancel(t *testing.T) {
...
+
+			if i > 5 {
```
svanharmelen: Since we're now in a loop I would set this to 30. We check every second so we'll not wait longer then needed, but it could prevent any flaky results when we run the test and TFE has some coincidental peak load at that moment.

run_test.go
ryanuber: That's actually not the case. The `canceled` status means that the run successfully canceled and has now ended.

run_test.go
ryanuber: This is only to paper over the time it takes for Sidekiq to pop the job off the queue and mark the run as canceling (it populates a timestamp). It doesn't need to cover the work going all the way through to the TBW and actually doing the cancel. We could do 30s, I just figured 5 was far more than enough in any normal case and anything longer indicates a serious performance degradation in TFE. Still think we should?

run_test.go
svanharmelen: Ah... So it will remain in the `RunCanceling` status until it's actually canceled. Makes sense now that I think of it 😏

svanharmelen: But... shouldn't it then be `if rTest.Status == tfe.RunCanceling` instead? As that is the target status for the next test if I understand it correctly?

run_test.go
svanharmelen: Guess it doesn't matter that much. I tent to use 30s in tests because I'm fine with some delays and maybe I'm testing this locally on a busy macbook. So if you want to keep it at 5s that fine too.

run_test.go
ryanuber: There's actually not a "canceling" state - the run remains "planning" or "applying" until it reaches a final state such as "canceled". However, once you have initiated a cancel, the API starts returning this timestamp so that you can figure out when a force-cancel would become available, if you need it. Does that make sense?

run_test.go
ryanuber: Makes sense, no reason not to I suppose. I bumped it up.

run_test.go
svanharmelen: That does make sense. At the same time it makes me wonder if a canceling state wouldn’t actually be helpful to have. But that’s not a question we need to answer here. So all good 👍🏻

PR #23, opened 2018-10-09
Add endpoints to get the capacity and queue

svanharmelen:


organization.go
```diff
@@ -242,3 +261,50 @@ func (s *organizations) Delete(ctx context.Context, organization string) error {
...
+	return c, nil
+}
+
+// QueueOptions represents the options for showing the queue.
+type QueueOptions struct {
+	ListOptions
+}
+
+// Queue shows the current run queue of an organization.
+func (s *organizations) Queue(ctx context.Context, organization string, options QueueOptions) (*Queue, error) {
```
ryanuber: I think we might want to call this `RunQueue` to be more clear. I could also see other types of queues being represented in the future (like an ingress queue, or a sentinel queue, etc.)

svanharmelen: 👍

PR #27, opened 2018-10-11
Check for STX and ETX when polling for logs

svanharmelen:
By checking for the STX and ETX characters, we have a way to determine if the process generating the logs is finished.

This prevents the need to poll for the status of a plan or apply during its execution.

The STX and ETX ASCII contol characters are part of the transmission control characters which are intended to structure a data stream.

The Start of Text character (STX) is to mark the start of the textual part of a stream and the End of Text character (ETX) marked the end of the data of a message.

___________________________________________________________________________________________________
➜  go-tfe git:(f-polling) ✗ go test -run=TestLogReader -v
=== RUN   TestLogReader_withMarkers
=== PAUSE TestLogReader_withMarkers
=== RUN   TestLogReader_withoutMarkers
=== PAUSE TestLogReader_withoutMarkers
=== RUN   TestLogReader_withoutEndOfTextMarker
=== PAUSE TestLogReader_withoutEndOfTextMarker
=== CONT  TestLogReader_withMarkers
=== CONT  TestLogReader_withoutEndOfTextMarker
=== CONT  TestLogReader_withoutMarkers
--- PASS: TestLogReader_withMarkers (33.26s)
--- PASS: TestLogReader_withoutMarkers (37.26s)
--- PASS: TestLogReader_withoutEndOfTextMarker (46.89s)
PASS
ok      github.com/hashicorp/go-tfe     47.742s

### DIFF
--- a/logreader.go
+++ b/logreader.go
@@ -1,9 +1,11 @@
 package tfe
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"net/url"
 	"time"
@@ -11,11 +13,24 @@ import (
 
 // LogReader implements io.Reader for streaming logs.
 type LogReader struct {
-	client *Client
-	ctx    context.Context
-	done   func() (bool, error)
-	logURL *url.URL
-	offset int64
+	client      *Client
+	ctx         context.Context
+	done        func() (bool, error)
+	logURL      *url.URL
+	offset      int64
+	reads       int
+	startOfText bool
+	endOfText   bool
+}
+
+// backoff will perform exponential backoff based on the iteration and
+// limited by the provided min and max (in milliseconds) durations.
+func backoff(min, max float64, iter int) time.Duration {
+	backoff := math.Pow(2, float64(iter)/5) * min
+	if backoff > max {
+		backoff = max
+	}
+	return time.Duration(backoff) * time.Millisecond
 }
 
 func (r *LogReader) Read(l []byte) (int, error) {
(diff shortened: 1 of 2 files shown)


logreader.go
```diff
@@ -70,16 +85,33 @@ func (r *LogReader) read(l []byte) (int, error) {
...
 	if written == 0 {
-		done, err := r.done()
-		if err != nil {
-			return 0, err
-		}
-		if done {
-			return 0, io.EOF
+		if (r.startOfText && r.endOfText) || // The logstream finished without issues.
+			(r.startOfText && r.reads%10 == 0) || // The logstream terminated unexpectedly.
+			(!r.startOfText && r.reads%2 == 0) { // The logstream doesn't support STX/ETX.
```
ryanuber: Should this just be `!r.startOfText && r.reads > 1` ?

logreader.go
```diff
@@ -70,16 +85,33 @@ func (r *LogReader) read(l []byte) (int, error) {
...
 	// run is finished and we are done reading all chunks.
 	if written == 0 {
-		done, err := r.done()
-		if err != nil {
-			return 0, err
-		}
-		if done {
-			return 0, io.EOF
+		if (r.startOfText && r.endOfText) || // The logstream finished without issues.
+			(r.startOfText && r.reads%10 == 0) || // The logstream terminated unexpectedly.
```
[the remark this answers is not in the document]
svanharmelen: No, it doesn't. I consider this sort of a health check to verify that we are still running. 

If we did get the STX marker, but did not yet see the ETX marker and we don't get any logs for 10 iterations (~25s), we enter the `if` statement and call the `done` function _once_ to determine if we are done (if the run is in any kind of final state) or not.

If the run isn't in a final state we just continue for another 10 iterations before we check again, unless we received any logs or the ETX marker before that. So even if we don't get any logs for minutes, we will still be waiting and check every ~25 seconds if we are still alive with one API call to the TFE API. 

We could make that period longer (do this every 15 iterations), but I thought that a maximum of 1 call every ~25s would be fine in terms of load? Especially considering that getting no logs at all for more then 10s is actually very rare. What do you think?

logreader.go
svanharmelen: Yeah we can change that, since we now already have the exponential backoff. But it does mean generating twice as much API calls for setups that do not yet run the updated TBW. So not sure if we should do that.

For the UX it will be a little bit better, but the only moment when you will really notice this is when the logs are already done and you are waiting for the binary to exit. If we set this to `r.reads > 1` it will take max 2s, of we set it to `r.reads%2 == 0` it will take max 4s.

Hmm... 4s is pretty long, so let's make this max 2s. then.

PR #41, opened 2018-11-20
Update the rate limit retries

svanharmelen:
After doing some more heavy testing, it turned out that having a higher
value here doesn't actually cause any downsides as the backoff algorithm
doesn't take the attempt number into account. So making this number higher
only makes it much less likely anyone will ever hit the limit.

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -164,8 +164,8 @@ func NewClient(cfg *Config) (*Client, error) {
 			ErrorHandler: retryablehttp.PassthroughErrorHandler,
 			HTTPClient:   config.HTTPClient,
 			RetryWaitMin: 100 * time.Millisecond,
-			RetryWaitMax: 300 * time.Millisecond,
-			RetryMax:     5,
+			RetryWaitMax: 400 * time.Millisecond,
+			RetryMax:     30,
 		},
 	}
 


svanharmelen:
Not quite... The client side rate limiter should prevent any rate limiting issues. But that will only work correctly as long as a user is using his/her own key. When multiple users, or when a CI/CD setup with multiple connected repos, share a key, they will still get rate limit errors when they execute a plan or apply concurrently.

The moment when a rate limit error is received, the retry logic kicks in. The custom backoff algorithm uses the customized `X-RateLimit-Reset` header to see how much milliseconds are left before a new bucket is available. That time is then used as the minimal delay, and a little jitter is added to spread out the calls a bit. The added jitter will be between 100ms and 400ms.

So it can definitely take longer than 15 seconds before we give up. But that's fine (IMO), as that means that during all that time the available buckets we're completely used. So a lot of other calls succeeded and there was still a lot of progress while the user is waiting.

ryanuber:
Makes sense @svanharmelen, and I agree. Thanks! I ask because we made an overly-resilient retrying client in `go-retryablehttp` before, and the result was that clients would sometimes end up retrying for many, many hours, which was quite undesirable. Since these are rate limiting errors, and the fact that we support a `context.Context` almost everywhere, I think you're right that this is fine for this use case. 👍

PR #46, opened 2018-12-18
Fix client cannot delete vcs integration in workspace

linc01n:


workspace_test.go
```diff
@@ -478,3 +478,21 @@ func TestWorkspacesUnassignSSHKey(t *testing.T) {
...
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, _ := createWorkspace(t, client, orgTest)
+
+	t.Run("remove vcs integration", func(t *testing.T) {
+
```
svanharmelen: _nitting_ blank line should be removed

workspace_test.go
```diff
@@ -478,3 +478,21 @@ func TestWorkspacesUnassignSSHKey(t *testing.T) {
...
+		w, err := client.Workspaces.RemoveVCS(ctx, orgTest.Name, wTest.Name)
+		require.NoError(t, err)
+		assert.Equal(t, w.VCSRepo, nil)
```
svanharmelen: I don't think this actually tests what you intent to test as the `createWorkspace` helper function creates a workspace without a VCS connection. So `VCSRepo` will be `nil` already, before calling `RemoveVCSConnection`.

I guess we should create a new helper function called `createWorkspaceWithVCS` that uses the existing `createOAuthToken` to first get a token, and then use when creating the workspace.

Let me know it you need any help with that.

workspace_test.go
```diff
@@ -478,3 +478,21 @@ func TestWorkspacesUnassignSSHKey(t *testing.T) {
...
+	})
+
```
svanharmelen: _nitting_ blank line should be removed

linc01n:
Hi @svanharmelen ,

I will fix the code next Monday. And I'm waiting approval to sign the CLA.

workspace_test.go
svanharmelen: That seems to be right 👍

helper_test.go
```diff
@@ -577,6 +577,44 @@ func createWorkspace(t *testing.T, client *Client, org *Organization) (*Workspac
...
+
+	token, ocTestCleanup := createOAuthToken(t, client, org)
+	w, err := client.Workspaces.Create(ctx, org.Name, WorkspaceCreateOptions{
+		Name: String(randomString(t)),
+		VCSRepo: &VCSRepoOptions{
+			OAuthTokenID: String(token.ID),
+		},
+	})
+
+	if err != nil {
```
svanharmelen: _nitting_ we usually don't insert a blank line between the call and the error check

helper_test.go
```diff
@@ -577,6 +577,44 @@ func createWorkspace(t *testing.T, client *Client, org *Organization) (*Workspac
...
+				"Workspace: %s\nError: %s", w.Name, err)
+		}
+
+		if orgCleanup != nil {
+			orgCleanup()
+		}
+
+		if ocTestCleanup != nil {
+			ocTestCleanup()
+		}
```
svanharmelen: I would swap calling the cleanup functions as the OAuth client is part of the org. So it _should_ be deleted when the org is deleted and so it would error if we try to deleted it after it is already gone.

We could also omit the `ocCleanup()` call, but I like to cleanup the bits in the same order as we created them instead of relying on the last cleanup call to remove everything.

helper_test.go
```diff
@@ -577,6 +577,44 @@ func createWorkspace(t *testing.T, client *Client, org *Organization) (*Workspac
...
+		org, orgCleanup = createOrganization(t, client)
+	}
+
+	ctx := context.Background()
+
+	token, ocTestCleanup := createOAuthToken(t, client, org)
+	w, err := client.Workspaces.Create(ctx, org.Name, WorkspaceCreateOptions{
+		Name: String(randomString(t)),
+		VCSRepo: &VCSRepoOptions{
+			OAuthTokenID: String(token.ID),
```
svanharmelen: We need to specify an `Identifier` as well, but that actually needs to exist in order for the call to succeed. If you could just add a hardcoded string for now, then I will update the test once this is merged so it will take a env var instead 👍

linc01n:
Hi @svanharmelen,

I fixed the code according to the comment. I'm lacking a TFE installation to run the test against. So I'm just guessing the test should pass.

svanharmelen:
@linc01n I added a commit to tweak and fix the tests. So all is good now and we can merge this PR as soon as you've signed the CLA 🎉 Thanks!

svanharmelen:
@linc01n are you able (and willing) to sign the CLA? As without that, I cannot merge this PR... Thanks!

linc01n:
@svanharmelen Sorry for the delay. I'm still waiting reply from legal. I just sent them an email for an update.

linc01n:
@svanharmelen Already signed!

svanharmelen:
LGTM! Thanks @linc01n!

PR #52, opened 2019-01-16
Add Organization Access on teams

chrisarcand:
This adds support for the new Organization Access feature on teams (https://www.terraform.io/docs/enterprise/users-teams-organizations/teams.html#managing-organization-access). As a side effect, I've also added Team.Update for updating teams as well as fixed an erroneous validation that kept users from being able to specify a team name with a space in it.

### DIFF
--- a/team.go
+++ b/team.go
@@ -24,6 +24,9 @@ type Teams interface {
 	// Read a team by its ID.
 	Read(ctx context.Context, teamID string) (*Team, error)
 
+	// Update a team by its ID.
+	Update(ctx context.Context, teamID string, options TeamUpdateOptions) (*Team, error)
+
 	// Delete a team by its ID.
 	Delete(ctx context.Context, teamID string) error
 }
@@ -41,16 +44,24 @@ type TeamList struct {
 
 // Team represents a Terraform Enterprise team.
 type Team struct {
-	ID          string           `jsonapi:"primary,teams"`
-	Name        string           `jsonapi:"attr,name"`
-	Permissions *TeamPermissions `jsonapi:"attr,permissions"`
-	UserCount   int              `jsonapi:"attr,users-count"`
+	ID                 string              `jsonapi:"primary,teams"`
+	Name               string              `jsonapi:"attr,name"`
+	OrganizationAccess *OrganizationAccess `jsonapi:"attr,organization-access"`
+	Permissions        *TeamPermissions    `jsonapi:"attr,permissions"`
+	UserCount          int                 `jsonapi:"attr,users-count"`
 
 	// Relations
 	Users []*User `jsonapi:"relation,users"`
 }
 
-// TeamPermissions represents the team permissions.
+// OrganizationAccess represents the team's permissions on its organization
+type OrganizationAccess struct {
+	ManagePolicies    bool `json:"manage-policies"`
+	ManageWorkspaces  bool `json:"manage-workspaces"`
+	ManageVCSSettings bool `json:"manage-vcs-settings"`
+}
+
+// TeamPermissions represents the current user's permissions on the team.
 type TeamPermissions struct {
 	CanDestroy          bool `json:"can-destroy"`
 	CanUpdateMembership bool `json:"can-update-membership"`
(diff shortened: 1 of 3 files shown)


team_test.go
```diff
@@ -139,6 +136,59 @@ func TestTeamsRead(t *testing.T) {
...
 }
 
+func TestTeamsUpdate(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	teamTest, teamTestCleanup := createTeam(t, client, orgTest)
```
svanharmelen: `teamTest` is called `tmTest` everywhere else, so I guess we should do that here as well to keep things consistent.

team.go
```diff
@@ -149,6 +160,49 @@ func (s *teams) Read(ctx context.Context, teamID string) (*Team, error) {
 	return t, nil
 }
 
+// TeamUpdateOptions represents the options for updating a team.
+type TeamUpdateOptions struct {
+	// For internal use only!
+	ID string `jsonapi:"primary,teams"`
+
+	// New name for the team
+	Name *string `jsonapi:"attr,name,omitempty"`
```
svanharmelen: If `Name` is indeed required (see comment below), then we should remove `omitempty`

team.go
```diff
@@ -149,6 +160,49 @@ func (s *teams) Read(ctx context.Context, teamID string) (*Team, error) {
...
+
+	// The team's organization access
+	OrganizationAccess *OrganizationAccess `jsonapi:"attr,organization-access,omitempty"`
+}
+
+func (o TeamUpdateOptions) valid() error {
+	if !validString(o.Name) {
+		return errors.New("name is required")
+	}
```
svanharmelen: Are you sure `Name` is required? That would mean that if you want to update the org access, you also have to fill in the existing name?

team_test.go
```diff
@@ -139,6 +136,59 @@ func TestTeamsRead(t *testing.T) {
...
+
+	t.Run("with valid options", func(t *testing.T) {
+		options := TeamUpdateOptions{
+			Name: String("Totally Cool Team"),
+			OrganizationAccess: &OrganizationAccess{
+				ManagePolicies:    false,
+				ManageVCSSettings: true},
+		}
+
+		team, err := client.Teams.Update(ctx, teamTest.ID, options)
```
svanharmelen: `team` => `tm` to keep it consistent

team_test.go
```diff
@@ -139,6 +136,59 @@ func TestTeamsRead(t *testing.T) {
...
+			)
+			assert.Equal(t,
+				options.OrganizationAccess.ManageVCSSettings,
+				item.OrganizationAccess.ManageVCSSettings,
+			)
+		}
+	})
+
+	t.Run("when the team does not exist", func(t *testing.T) {
+		tm, err := client.Teams.Update(ctx, "nonexisting", TeamUpdateOptions{Name: String("Cool Team")})
```
svanharmelen:
```suggestion
		tm, err := client.Teams.Update(ctx, "nonexisting", TeamUpdateOptions{
				Name: String("foo bar"),
		})
```

svanharmelen: So line length and team name are updated by this ^^

team.go
chrisarcand: No, not required, my bad.

chrisarcand:
Eh, I think some other things might have to change before this can be merged, influenced by work in the TFE provider. WIP for now.

team_test.go
```diff
@@ -132,13 +129,69 @@ func TestTeamsRead(t *testing.T) {
...
+		require.NoError(t, err)
+
+		for _, item := range []*Team{
+			tm,
+			refreshed,
+		} {
+			assert.Equal(t, *options.Name, item.Name)
+			assert.Equal(t,
+				options.OrganizationAccess.ManagePolicies,
+				Bool(item.OrganizationAccess.ManagePolicies),
```
svanharmelen: These are better written as:

```
	*options.OrganizationAccess.ManagePolicies,
	item.OrganizationAccess.ManagePolicies,
```

So dereferencing the existing pointer instead of creating a new one (which is the dereferenced in the assert function).

team_test.go
```diff
@@ -132,13 +129,69 @@ func TestTeamsRead(t *testing.T) {
...
+		tm, err := client.Teams.Update(ctx, "nonexisting", TeamUpdateOptions{
+			Name: String("foo bar"),
+		})
+		assert.Nil(t, tm)
+		assert.Equal(t, err, ErrResourceNotFound)
+	})
+
+	t.Run("without a valid team ID", func(t *testing.T) {
+		tm, err := client.Teams.Update(ctx, badIdentifier, TeamUpdateOptions{})
+
```
svanharmelen: You can drop this blanc line

PR #56, opened 2019-02-15
Add timestamp fields for planned runs

eculver:
For planned runs, there are a few extra timestamp fields that were missing. I'm not sure when these were added but, like the others, if they are not returned by the API, they will just be 0 values.


svanharmelen:
Thanks @eculver! LGTM!

PR #64, opened 2019-03-21
Allow creation of variables with no value

justincampbell:
An empty string and null are both valid values for a variable according to the [API docs](https://www.terraform.io/docs/enterprise/api/variables.html) and testing done by @sudomateo.

A null value in the API results in an empty string being stored and returned from the server.

### DIFF
--- a/variable.go
+++ b/variable.go
@@ -129,9 +129,6 @@ func (o VariableCreateOptions) valid() error {
 	if !validString(o.Key) {
 		return errors.New("key is required")
 	}
-	if !validString(o.Value) {
-		return errors.New("value is required")
-	}
 	if o.Category == nil {
 		return errors.New("category is required")
 	}
--- a/variable_test.go
+++ b/variable_test.go
@@ -107,6 +107,18 @@ func TestVariablesCreate(t *testing.T) {
 		assert.EqualError(t, err, "key is required")
 	})
 
+	t.Run("when options has an empty key", func(t *testing.T) {
+		options := VariableCreateOptions{
+			Key:       String(""),
+			Value:     String(randomString(t)),
+			Category:  Category(CategoryTerraform),
+			Workspace: wTest,
+		}
+
+		_, err := client.Variables.Create(ctx, options)
+		assert.EqualError(t, err, "key is required")
+	})
+
 	t.Run("when options is missing value", func(t *testing.T) {
 		options := VariableCreateOptions{
 			Key:       String(randomString(t)),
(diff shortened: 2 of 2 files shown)


validations.go
```diff
@@ -9,7 +9,7 @@ var reStringID = regexp.MustCompile(`^[a-zA-Z0-9\-\._]+$`)
 
 // validString checks if the given input is present and non-empty.
 func validString(v *string) bool {
-	return v != nil && *v != ""
+	return v != nil
```
sudomateo: This is more thinking out loud than an actual review since I'm still learning golang and learning the Terraform provider structure.

I think removing the check for a non-empty string may let the `key` attribute of the resource be set to an empty string.
Instead, should the `validString` function call be removed from the `value` attribute [here](https://github.com/hashicorp/go-tfe/blob/master/variable.go#L128-L134)?

Again this is more for learning purposes rather than actual feedback.

justincampbell: Good point! Will reach out.

justincampbell:
I paired with @sudomateo on this to remove the check entirely, as the API does not require a value to be passed in.

svanharmelen:
According to the API docs `data.attributes.value` is a required field as it doesn't have a default: `Properties without a default value are required`. So before we change this logic here, I would first like to know if allowing an empty value is actually a bug in the API instead of the docs. And it would be nice to understand the use case why one would want to create empty variables.

chrisarcand:
Was typing a reply, but Sander is mind-linked with me and said exactly what I was going to say 😄 Let's make sure this isn't a bug first, and if it isn't the docs seem to indicate the opposite so let's fix those too.

sudomateo:
The only use case I can think of is a producer creating a skeleton workspace that a consumer would then use. But that producer can just set variables to `CHANGE_ME` or something. I'm in agreement to confirm that this isn't an API bug first and then make the appropriate changes afterwards.

justincampbell:
My 2 cents: We need to allow empty string values in the API. I don't think it matters if we coerce `null` to `""` or not, but on the API some people may already be using that behavior so I'd be wary of changing it.

I propose we update the API docs to say that the default is `""`, and either merge this or disallow a nil Value.

beekus:
Although empty/nil variable values don't make sense to me, we should check out how TF OSS deals with empty variables and make a decision based off of that.

chrisarcand:
[Terraform 0.12 will have null value support](https://github.com/hashicorp/terraform/issues/5471#issuecomment-393904735) and after consulting the team it sounds like accepting `""` is a good idea as TF continues to get better empty/nil support in the future (and as mentioned, it would be a non-breaking change for us, too) 👍

PR #67, opened 2019-04-09
Implement Admin Runs List API

linc01n:
I'm not sure is this how hashicorp wants to implement the Admin API.

I think I should improve the test a bit. But the `q` parameter test is a bit tricky to test. I don't have a TFE instance to run my test against.

### DIFF
--- a/admin_run.go
+++ b/admin_run.go
@@ -1,0 +1,48 @@
+package tfe
+
+import (
+	"context"
+)
+
+// Compile-time proof of interface implementation.
+var _ AdminRuns = (*adminRuns)(nil)
+
+// AdminRuns describes all the admin run related methods that the Terraform Enterprise
+// API supports.
+//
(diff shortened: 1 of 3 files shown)


svanharmelen:
Thanks for your PR @linc01n. Unfortunately (at least for now) we do not want to add support for the admin endpoints to this package.

We are thinking about and discussing possible architectures for adding the admin endpoints in the future, but we haven't made any decisions yet. One option is to make an `Admin` service on the client struct, so all admin related endpoints are defined in the same location (e.g. `client.Admin.Runs.List()`). Another option is to make a dedicated client for all the admin endpoints.

So I'm very sorry, but for now I'm going to close your PR until we've settled on an architecture and know how we want to support the admin endpoints. Thanks again for your PR and I hope that you understand our motivation for not merging this right now.

PR #72, opened 2019-04-15
new runs api update

thrashr888:


helper_test.go
```diff
@@ -357,6 +361,59 @@ func createPlannedRun(t *testing.T, client *Client, w *Workspace) (*Run, func())
...
+
+		if i > 30 {
+			rCleanup()
+			t.Fatal("Timeout waiting for run to be planned")
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	return r, func() {
```
mwhooker: Maybe can put this in a call to defer, or put the infinite loop above in a goroutine? Is there prior art you can look at? this looks like an off pattern to me

cost_estimation_test.go
```diff
@@ -0,0 +1,65 @@
...
+		p, err := client.CostEstimations.Read(ctx, rTest.CostEstimation.ID)
+		require.NoError(t, err)
+
+		logReader, err := client.CostEstimations.Logs(ctx, p.ID)
+		require.NoError(t, err)
+
+		logs, err := ioutil.ReadAll(logReader)
+		require.NoError(t, err)
+
+		assert.Contains(t, string(logs), "1 to add, 0 to change, 0 to destroy")
```
svanharmelen: We should check for for something cost estimation specific here, to make sure we have received cost estimation output as well.

cost_estimation.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	return &LogReader{
+		client: s.client,
+		ctx:    ctx,
+		done:   done,
+		logURL: u,
+	}, nil
+}
```
svanharmelen: I think the big question here is if this log will be a streaming log like the plan and apply logs are, of that it will be more like the policy-check logs. Looking at how things work, I think it is closer to the latter which means this will not give the desired result.

cost_estimation_test.go
thrashr888: You're right. What's here is kinda dumb.

thrashr888: I don't think we want to test for anything specific yet since it will be changing. Checking for read errors is good, though.

helper_test.go
```diff
@@ -262,7 +262,7 @@ func createOAuthToken(t *testing.T, client *Client, org *Organization) (*OAuthTo
 func createOrganization(t *testing.T, client *Client) (*Organization, func()) {
 	ctx := context.Background()
 	org, err := client.Organizations.Create(ctx, OrganizationCreateOptions{
-		Name:  String(randomString(t)),
+		Name:  String("go-" + randomString(t)),
```
thrashr888: this is new. it namespaces the test organizations so they're easier to find & cleanup. maxchars is 40 so this barely fits.

helper_test.go
```diff
@@ -342,19 +342,18 @@ func createPlannedRun(t *testing.T, client *Client, w *Workspace) (*Run, func())
...
 
-		if r.Status == RunPlanned || r.Status == RunPolicyChecked || r.Status == RunPolicyOverride {
-			break
+		switch r.Status {
+		case RunPlanned, RunCostEstimated, RunPolicyChecked, RunPolicyOverride:
+			return r, rCleanup
 		}
 
-		if i > 30 {
+		if i > 45 {
```
thrashr888: support an extra run stage. it got too close to the full 30s during my local testing. 45 feels safer, depending on the speed of your laptop w/ containers.

helper_test.go
svanharmelen: I don't know if we really need this, but if we do I suggest to prefix it with `tst-` as it seems more common and can be used by other tests as well.

I took the liberty to update this in the PR.

PR #75, opened 2019-05-09
Add workspace filtering fields

alisdair:
I have not yet figured out how to set up an environment to run the tests, so this is untested.


alisdair:
Tests pass locally, and manual verification using the client with a modified example confirmed that all's well.

PR #76, opened 2019-05-22
Remove mentions of cost estimation pre-alpha

davidcelis:


run.go
```diff
@@ -53,8 +53,6 @@ const (
 	RunApplying           RunStatus = "applying"
 	RunCanceled           RunStatus = "canceled"
 	RunConfirmed          RunStatus = "confirmed"
-	RunCostEstimated      RunStatus = "cost_estimated"
-	RunCostEstimating     RunStatus = "cost_estimating"
```
svanharmelen: I think we should keep these states as in some cases they can be returned right?

helper_test.go
```diff
@@ -343,7 +343,7 @@ func createPlannedRun(t *testing.T, client *Client, w *Workspace) (*Run, func())
 		}
 
 		switch r.Status {
-		case RunPlanned, RunCostEstimated, RunPolicyChecked, RunPolicyOverride:
```
svanharmelen: I think we should keep this state as in some cases it can be returned right?

run.go
thrashr888: I think that, for now, these are not possible to see in production. If they end up changing, we'll have to remove/replace them anyway, which might be easier to PR as a group.

helper_test.go
thrashr888: See other comment.

run.go
svanharmelen: Check 👍🏻

PR #77, opened 2019-06-13
Add support for VCS policy sets

vancluever:
This adds support for the upcoming VCS policy sets feature.

This was a minimal change, requiring only the extra options in
PolicySetCreateOptions and PolicySet.

### DIFF
--- a/policy_set.go
+++ b/policy_set.go
@@ -28,10 +28,13 @@ type PolicySets interface {
 	// Update an existing policy set.
 	Update(ctx context.Context, policySetID string, options PolicySetUpdateOptions) (*PolicySet, error)
 
-	// Add policies to a policy set.
+	// Add policies to a policy set. This function can only be used
+	// when there is no VCS repository associated with the policy set.
 	AddPolicies(ctx context.Context, policySetID string, options PolicySetAddPoliciesOptions) error
 
-	// Remove policies from a policy set.
+	// Remove policies from a policy set. This function can only be
+	// used when there is no VCS repository associated with the policy
+	// set.
 	RemovePolicies(ctx context.Context, policySetID string, options PolicySetRemovePoliciesOptions) error
 
 	// Add workspaces to a policy set.
@@ -61,7 +64,9 @@ type PolicySet struct {
 	Name           string    `jsonapi:"attr,name"`
 	Description    string    `jsonapi:"attr,description"`
 	Global         bool      `jsonapi:"attr,global"`
+	PoliciesPath   string    `jsonapi:"attr,policies-path,omitempty"`
 	PolicyCount    int       `jsonapi:"attr,policy-count"`
+	VCSRepo        *VCSRepo  `jsonapi:"attr,vcs-repo"`
 	WorkspaceCount int       `jsonapi:"attr,workspace-count"`
 	CreatedAt      time.Time `jsonapi:"attr,created-at,iso8601"`
 	UpdatedAt      time.Time `jsonapi:"attr,updated-at,iso8601"`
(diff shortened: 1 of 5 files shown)


vancluever:
I might change the test around after taking a bit of a look at our CI tomorrow to see what we are using for a test fixture for VCS integrations. For now, I've populated every option - even the branch - the supplied repo needs to be able to ingress policies at the branch and subdirectory given in the test.

policy_set.go
```diff
@@ -61,6 +64,8 @@ type PolicySet struct {
 	Name           string    `jsonapi:"attr,name"`
 	Description    string    `jsonapi:"attr,description"`
 	Global         bool      `jsonapi:"attr,global"`
+	PoliciesPath   string    `jsonapi:"attr,policies-path,omitempty"`
+	VCSRepo        *VCSRepo  `jsonapi:"attr,vcs-repo,omitempty"`
```
svanharmelen: I try to keep a (sort of) alphabetical order here, so I would prefer to reorder the new fields just a little:

```
	ID             string    `jsonapi:"primary,policy-sets"`
	Name           string    `jsonapi:"attr,name"`
	Description    string    `jsonapi:"attr,description"`
	Global         bool      `jsonapi:"attr,global"`
	PoliciesPath   string    `jsonapi:"attr,policies-path,omitempty"`
	PolicyCount    int       `jsonapi:"attr,policy-count"`
	VCSRepo        *VCSRepo  `jsonapi:"attr,vcs-repo,omitempty"`
	WorkspaceCount int       `jsonapi:"attr,workspace-count"`
	CreatedAt      time.Time `jsonapi:"attr,created-at,iso8601"`
	UpdatedAt      time.Time `jsonapi:"attr,updated-at,iso8601"`
```

Also the `omitempty` tags should be removed from the new fields in this struct.

svanharmelen: In terms of consistency `PolicyPath` would probably be a better name, but I guess that ship has sailed already :wink:

policy_set.go
```diff
@@ -115,6 +120,18 @@ type PolicySetCreateOptions struct {
...
+	// files and directories outside of this sub-path will be ignored.
+	// This option may only be specified when a VCS repo is present.
+	PoliciesPath *string `jsonapi:"attr,policies-path,omitempty"`
+
+	// VCS repository information. When present, the policies and
+	// configuration will be sourced from the specified VCS repository
+	// instead of being defined within the policy set itself. Note that
+	// this option is mutually exclusive with the Policies option and
+	// both cannot be used at the same time.
+	VCSRepo *VCSRepoOptions `jsonapi:"attr,vcs-repo,omitempty"`
```
svanharmelen: Same thing about the ordering and are you sure you need the full `VCSRepoOptions` object here?

vancluever: Yeah, I double-checked on this. We are using the same bits on the backend here, so in essentiality, it's the same type. All of the options are available to the policy set request as seen in [the documentation](https://github.com/hashicorp/terraform-website/blob/39f0af766d655f5b938696371d4df0c232756b79/content/source/docs/enterprise/api/policy-sets.html.md#request-body). I was thinking of spinning off the VCS repo stuff into its own file, but it doesn't necessarily seem to make sense at this time as it's such a minor amount of code.

policy_set.go
vancluever: This has been updated now!

policy_set.go
svanharmelen: > I was thinking of spinning off the VCS repo stuff into its own file, but it doesn't necessarily seem to make sense at this time as it's such a minor amount of code

Fully agree 👍

policy_set.go
```diff
@@ -61,7 +64,9 @@ type PolicySet struct {
 	Name           string    `jsonapi:"attr,name"`
 	Description    string    `jsonapi:"attr,description"`
 	Global         bool      `jsonapi:"attr,global"`
+	PoliciesPath   string    `jsonapi:"attr,policies-path,omitempty"`
 	PolicyCount    int       `jsonapi:"attr,policy-count"`
+	VCSRepo        *VCSRepo  `jsonapi:"attr,vcs-repo,omitempty"`
```
svanharmelen: The ordering satisfies my ODC now, but the tags still contain `omitempty` 😉

vancluever: I thought this was copied from `WorkSpace` but I guess not, will update!

vancluever:
> The code present looks good, but I think we'll want to add to the `valid()` functions for `PolicySetCreateOptions` and `PolicySetUpdateOptions` so that it matches what's going on in the provider (namely that `VCSRepo` and `Policies` conflict but that one of them is required)

I actually don't think that this is needed?

Reasoning: We aren't currently requiring that `Policies` be a non-nil or non-zero-length value, so I'm assuming if both are absent, the API will perform some default behavior (I guess a legacy policy set)?

The reason why I added the inclusivity check in the `tfe_policy_set` resource is because `policy_ids` was previously set to `Required`. In retrospect I'm wondering if that was actually needed either as we could have just defaulted to an empty set.

svanharmelen:
I'm not sure it we need to add these checks in the `valid()` method as we could also rely on the API response (which I assume does the same checks). But on the other hand is also wouldn't hurt to check things before making the actual API call. I'll leave that one for you to decide...

I updated the code (nitpicking stuff mostly) to make it fit in nicely and be consistent with the rest of the code in this package. So when its decided how to update (if to update) the `valid()` method, this should be good to be merged.

PR #78, opened 2019-06-26
Hit the new TFE ping endpoint

alisdair:
Stop filling the logs with 404s due to hitting an invalid URL.

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -32,7 +32,8 @@ const (
 	DefaultAddress = "https://app.terraform.io"
 	// DefaultBasePath on which the API is served.
 	DefaultBasePath = "/api/v2/"
-	PingEndpoint    = "ping"
+	// No-op API endpoint used to configure the rate limiter
+	PingEndpoint = "ping"
 )
 
 var (
--- a/tfe.go
+++ b/tfe.go
@@ -32,6 +32,7 @@ const (
 	DefaultAddress = "https://app.terraform.io"
 	// DefaultBasePath on which the API is served.
 	DefaultBasePath = "/api/v2/"
+	PingEndpoint    = "ping"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


tfe_test.go
```diff
@@ -16,7 +16,7 @@ func TestClient_newClient(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/vnd.api+json")
 		w.Header().Set("X-RateLimit-Limit", "30")
-		w.WriteHeader(404) // We query the configured base URL which should return a 404.
+		w.WriteHeader(204) // We query the configured ping URL which should return a 404.
```
mwhooker:
```suggestion
		w.WriteHeader(204) // We query the configured ping URL which should return a 204.
```

tfe.go
```diff
@@ -32,6 +32,7 @@ const (
 	DefaultAddress = "https://app.terraform.io"
 	// DefaultBasePath on which the API is served.
 	DefaultBasePath = "/api/v2/"
+	PingEndpoint    = "ping"
```
svanharmelen: Maybe we should add a small comment to explain what this is used for?

svanharmelen: Uuh... Did you test how this works against a PTFE instance that doesn't offer the `/ping` endpoint? Or did that endpoint already exist? If not it would be good to double check if the required headers are returned in that case (I think they are not returned correctly when getting a 301 or 302).

alisdair: > Maybe we should add a small comment to explain what this is used for?

👍

> Did you test how this works against a PTFE instance that doesn't offer the `/ping` endpoint?

It returns a 404, just like the previous request to `/api/v2/` did. The headers are there.

PR #84, opened 2019-09-14
Adding support for configuring operations

sudomateo:
Attempting to add the ability to manage execution mode via the SDK. Looking for some feedback and also help with adding this to tests.

### DIFF
--- a/workspace_test.go
+++ b/workspace_test.go
@@ -90,6 +90,7 @@ func TestWorkspacesCreate(t *testing.T) {
 			AutoApply:           Bool(true),
 			FileTriggersEnabled: Bool(true),
 			QueueAllRuns:        Bool(true),
+			Operations:          Bool(true),
 			TerraformVersion:    String("0.11.0"),
 			TriggerPrefixes:     []string{"/modules", "/shared"},
 			WorkingDirectory:    String("bar/"),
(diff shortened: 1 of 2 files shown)


sudomateo:
Thanks for the feedback. I added this to existing tests, but I didn't write a new test as I would need some more guidance there.

svanharmelen:
When/if all the checks pass (I can no longer test this easily myself), this should be good to go 👍

workspace_test.go
```diff
@@ -376,6 +383,7 @@ func TestWorkspacesUpdateByID(t *testing.T) {
 			assert.Equal(t, *options.AutoApply, item.AutoApply)
 			assert.Equal(t, *options.FileTriggersEnabled, item.FileTriggersEnabled)
 			assert.Equal(t, *options.QueueAllRuns, item.QueueAllRuns)
+			assert.Equal(t, *options.Operations, item.Operations)
```
svanharmelen: In order to keep things (more or less) consistent, this should be moved above `QueueAllRuns` everywhere in this file 😏

sudomateo: Oops! Clearly I can't alphabet. Going to make the changes soon. I'm also in the process of working with a developer regarding the CI builds.

PR #93, opened 2019-12-05
Retry 500 server errors by default after successful initialization

alisdair:
This new default means that the TFE provider will retry requests which result in 5xx errors, but still fail on initialization if the instance is unreachable. This should improve robustness to intermittent network failures, without making it more difficult to diagnose incorrectly configured networks or TFE instances.

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -175,10 +175,11 @@ func NewClient(cfg *Config) (*Client, error) {
 
 	// Create the client.
 	client := &Client{
-		baseURL:      baseURL,
-		token:        config.Token,
-		headers:      config.Headers,
-		retryLogHook: config.RetryLogHook,
+		baseURL:           baseURL,
+		token:             config.Token,
+		headers:           config.Headers,
+		retryLogHook:      config.RetryLogHook,
+		retryServerErrors: true,
 	}
 
 	client.http = &retryablehttp.Client{
--- a/tfe.go
+++ b/tfe.go
@@ -98,13 +98,12 @@ func DefaultConfig() *Config {
 // Client is the Terraform Enterprise API client. It provides the basic
 // connectivity and configuration for accessing the TFE API.
 type Client struct {
-	baseURL           *url.URL
-	token             string
-	headers           http.Header
-	http              *retryablehttp.Client
-	limiter           *rate.Limiter
-	retryLogHook      RetryLogHook
-	retryServerErrors bool
+	baseURL      *url.URL
+	token        string
+	headers      http.Header
+	http         *retryablehttp.Client
+	limiter      *rate.Limiter
+	retryLogHook RetryLogHook
 
 	Applies                    Applies
 	ConfigurationVersions      ConfigurationVersions
@@ -183,7 +182,7 @@ func NewClient(cfg *Config) (*Client, error) {
 
 	client.http = &retryablehttp.Client{
 		Backoff:      client.retryHTTPBackoff,
-		CheckRetry:   client.retryHTTPCheck,
+		CheckRetry:   client.retryHTTPCheckNoServerErrors,
 		ErrorHandler: retryablehttp.PassthroughErrorHandler,
 		HTTPClient:   config.HTTPClient,
 		RetryWaitMin: 100 * time.Millisecond,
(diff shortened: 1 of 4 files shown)


svanharmelen:
Is there a reason to change this default behavior? This could be annoying when you have something configured incorrectly (mainly thinking about locally installed enterprise setups). In those cases it will take a while (maybe 10+ minutes) before you will get an error.

After the first initial (if successful) call to the backend, this will be enabled. So wonder what the idea is behind changing this.

alisdair:
Thanks for the feedback, Sander! That's a very good point about the down side of this—I noticed the same thing in local testing.

> Is there a reason to change this default behavior? 

We're seeing occasional 502 errors when running the TFE provider inside Terraform Cloud runs. Because retry is disabled for 5xx errors, this means that a single 502 in a long provider run causes the entire run to be cancelled. While we're tracking down and fixing the server side of the problem, we thought it would be helpful to enable retries at the client.

> After the first initial (if successful) call to the backend, this will be enabled. 

Ah, I hadn't spotted that. That behaviour seems very reasonable, but I'm not sure how to achieve the same thing with the provider. I'll investigate that approach before proceeding here.

svanharmelen:
Instead of directly returning [here](https://github.com/terraform-providers/terraform-provider-tfe/blob/master/tfe/provider.go#L197) you could change this to something like:

```
tfeClient, err := tfe.NewClient(cfg)
if err != nil {
  return nil, fmt.Errorf("could not create client: %v", err)
}

tfeClient.RetryServerErrors(true)

return tfeClient, nil
```

That (or something alike) should enable retrying on 500 errors right after the initial setup of the client.

alisdair:
I've been working on an alternative approach for this today, but after seeing behaviour in testing I'm not sure it's necessary.

We already make a request to the API as part of client initialization, hitting the "ping" endpoint to establish initial values for rate limiting. If we simply enable server-error-retries in the constructor after this happens, we should get what we want: fast-fail for incorrectly set-up TFE instances, and request retry for any subsequent server errors.

I pushed this idea as 32d8ceb and I'll update the PR description. @svanharmelen, I'd appreciate your thoughts!

svanharmelen:
@alisdair I guess this should work as well, but it gives users of this client just a little bit less flexibility for when they want to start/enable this behavior. For example in Terraform we currently only enable the behavior after the client is fully initialized which also includes a check on the entitlements. But I'm not sure if changing the behavior to this would be a problem in that case.

alisdair:
@chrisarcand Thanks for the suggestion! I wasn't aware of the updates to retryablehttp. I think f4812e8 implements what you were suggesting.

tfe_test.go
```diff
@@ -316,14 +316,14 @@ func TestClient_retryHTTPCheck(t *testing.T) {
...
-			checkErr: connErr,
+			checkOK:  true,
+			checkErr: nil,
 		},
 		"err-with-server-errors": {
 			err:               connErr,
 			retryServerErrors: true,
 			checkOK:           true,
-			checkErr:          connErr,
+			checkErr:          nil,
```
svanharmelen: Why did these test change? That seems to indicate the behavior changed as well? I might have to look at bit deeper at this one to understand it, but I wouldn't have expected these changes.

tfe.go
```diff
@@ -227,22 +230,33 @@ func NewClient(cfg *Config) (*Client, error) {
...
 // will retry both rate limit (429) and server (>= 500) errors.
 func (c *Client) retryHTTPCheck(ctx context.Context, resp *http.Response, err error) (bool, error) {
-	if ctx.Err() != nil {
-		return false, ctx.Err()
-	}
-	if err != nil {
-		return c.retryServerErrors, err
-	}
-	if resp.StatusCode == 429 || (c.retryServerErrors && resp.StatusCode >= 500) {
+	retry, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
```
svanharmelen: Are you sure it's save to not check this error here?

tfe.go
```diff
@@ -227,22 +230,33 @@ func NewClient(cfg *Config) (*Client, error) {
...
+	if !retry && resp != nil && resp.StatusCode == 429 {
 		return true, nil
 	}
-	return false, nil
+
+	return retry, err
+}
+
+func (c *Client) retryHTTPCheckNoServerErrors(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	retry, err := c.retryHTTPCheck(ctx, resp, err)
```
svanharmelen: Are you sure it's save to not check this error here?

alisdair:
This is getting too involved. Instead, I'm just changing the usage of go-tfe in the provider, as suggested earlier: https://github.com/terraform-providers/terraform-provider-tfe/pull/109

PR #94, opened 2019-12-09
adding in private key for ado_server only

veverkap:
This adds in PrivateKey to the OAuthClientCreateOptions struct to allow users to include the RSA private key associated with the ado server.

This is only valid for Azure DevOps Server VCS provider and is validated.

### DIFF
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -122,6 +122,9 @@ type OAuthClientCreateOptions struct {
 
 	// The VCS provider being connected with.
 	ServiceProvider *ServiceProviderType `jsonapi:"attr,service-provider"`
+
+	// Private key associated with this vcs provider - only available for ado_server
+	PrivateKey *string `jsonapi:"attr,private-key"`
 }
 
 func (o OAuthClientCreateOptions) valid() error {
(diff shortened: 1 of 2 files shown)


oauth_client_test.go
```diff
@@ -224,3 +224,88 @@ func TestOAuthClientsDelete(t *testing.T) {
 		assert.EqualError(t, err, "invalid value for OAuth client ID")
 	})
 }
+
+func TestOAuthClientsCreateOptionsValid(t *testing.T) {
+
```
svanharmelen:
```suggestion
```

oauth_client.go
```diff
@@ -122,6 +122,9 @@ type OAuthClientCreateOptions struct {
 
 	// The VCS provider being connected with.
 	ServiceProvider *ServiceProviderType `jsonapi:"attr,service-provider"`
+
+	// Private key associated with this vcs provider - only available for ado_server
+	PrivateKey *string `jsonapi:"attr,private-key"`
```
svanharmelen: Not blocking, but these attributes are ordered alphabetically...

PR #95, opened 2019-12-09
remove spurious mod

veverkap:


thrashr888:
what did you run to get this result? `go mod tidy`?

veverkap:
Yes. I'm not sure how I got this spurious mod added in the first place.  

It was added in https://github.com/hashicorp/go-tfe/commit/a5eea2aff06e96800f3d249d6e035c70ba056aee but all I did in that PR (AFAIK) was move the property to alphabetize it.

PR #96, opened 2019-12-12
fixing nil check for validString check

veverkap:
So, o.PrivateKey references a memory address it seems, not a string value so it's never nil.


veverkap:
> Can we add test coverage for this?

So, um, there WAS test coverage but it didn't work.  

https://github.com/hashicorp/go-tfe/blob/veverkap/ado_bugfix/oauth_client_test.go#L285-L309

It's possible I'm doing it wrong.

veverkap:
> Can we add test coverage for this?

I added in a better test per your suggestion @alisdair

PR #97, opened 2019-12-17
Update variable API and add parameter API

cam-stitt:
We have recently performed some work to simplify the API for interacting with variables, as well as adding the policy set parameters API for interacting with Sentinel.

This work updates the existing variable API to use the nested workspace routing as well as adds the client for interacting with the Parameter API.

Ideally, it would be nice to merge the website docs prior to releasing this code.

- [Docs PR](https://github.com/hashicorp/terraform-website/pull/1060)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Currently the following endpoints are supported:
 - [x] [Organization Memberships](https://www.terraform.io/docs/cloud/api/organization-memberships.html)
 - [x] [Organization Tokens](https://www.terraform.io/docs/enterprise/api/organization-tokens.html)
 - [x] [Policies](https://www.terraform.io/docs/enterprise/api/policies.html)
+- [x] [Policy Set Parameters](https://www.terraform.io/docs/enterprise/api/policy-set-params.html)
 - [x] [Policy Sets](https://www.terraform.io/docs/enterprise/api/policy-sets.html)
 - [x] [Policy Checks](https://www.terraform.io/docs/enterprise/api/policy-checks.html)
 - [ ] [Registry Modules](https://www.terraform.io/docs/enterprise/api/modules.html)
@@ -40,7 +41,6 @@ Currently the following endpoints are supported:
 - [x] [Team Tokens](https://www.terraform.io/docs/enterprise/api/team-tokens.html)
 - [x] [Teams](https://www.terraform.io/docs/enterprise/api/teams.html)
 - [x] [Workspace Variables](https://www.terraform.io/docs/enterprise/api/workspace-variables.html)
-- [x] [Policy Set Parameters](https://www.terraform.io/docs/enterprise/api/policy-set-params.html)
 - [x] [Workspaces](https://www.terraform.io/docs/enterprise/api/workspaces.html)
 - [ ] [Admin](https://www.terraform.io/docs/enterprise/api/admin/index.html)
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ Currently the following endpoints are supported:
 - [x] [Team Memberships](https://www.terraform.io/docs/enterprise/api/team-members.html)
 - [x] [Team Tokens](https://www.terraform.io/docs/enterprise/api/team-tokens.html)
 - [x] [Teams](https://www.terraform.io/docs/enterprise/api/teams.html)
-- [x] [Variables](https://www.terraform.io/docs/enterprise/api/variables.html)
+- [x] [Workspace Variables](https://www.terraform.io/docs/enterprise/api/workspace-variables.html)
+- [x] [Policy Set Parameters](https://www.terraform.io/docs/enterprise/api/policy-set-params.html)
 - [x] [Workspaces](https://www.terraform.io/docs/enterprise/api/workspaces.html)
 - [ ] [Admin](https://www.terraform.io/docs/enterprise/api/admin/index.html)
 
--- a/helper_test.go
+++ b/helper_test.go
@@ -117,6 +117,36 @@ func createNotificationConfiguration(t *testing.T, client *Client, w *Workspace)
 	}
 }
 
+func createPolicySetParameter(t *testing.T, client *Client, ps *PolicySet) (*PolicySetParameter, func()) {
+	var psCleanup func()
+
+	if ps == nil {
+		ps, psCleanup = createPolicySet(t, client, nil, nil, nil)
+	}
+
+	ctx := context.Background()
+	v, err := client.PolicySetParameters.Create(ctx, ps.ID, PolicySetParameterCreateOptions{
+		Key:      String(randomString(t)),
+		Value:    String(randomString(t)),
+		Category: Category(CategoryPolicySet),
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return v, func() {
+		if err := client.PolicySetParameters.Delete(ctx, ps.ID, v.ID); err != nil {
+			t.Errorf("Error destroying variable! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"Parameter: %s\nError: %s", v.Key, err)
+		}
+
+		if psCleanup != nil {
+			psCleanup()
+		}
+	}
+}
+
 func createPolicySet(t *testing.T, client *Client, org *Organization, policies []*Policy, workspaces []*Workspace) (*PolicySet, func()) {
 	var orgCleanup func()
 
@@ -671,36 +701,6 @@ func createVariable(t *testing.T, client *Client, w *Workspace) (*Variable, func
 
 		if wCleanup != nil {
 			wCleanup()
-		}
-	}
-}
-
-func createParameter(t *testing.T, client *Client, ps *PolicySet) (*Parameter, func()) {
-	var psCleanup func()
-
-	if ps == nil {
-		ps, psCleanup = createPolicySet(t, client, nil, nil, nil)
-	}
-
-	ctx := context.Background()
-	v, err := client.Parameters.Create(ctx, ps.ID, ParameterCreateOptions{
-		Key:      String(randomString(t)),
-		Value:    String(randomString(t)),
-		Category: Category(CategoryPolicySet),
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	return v, func() {
-		if err := client.Parameters.Delete(ctx, ps.ID, v.ID); err != nil {
-			t.Errorf("Error destroying variable! WARNING: Dangling resources\n"+
-				"may exist! The full error is shown below.\n\n"+
-				"Parameter: %s\nError: %s", v.Key, err)
-		}
-
-		if psCleanup != nil {
-			psCleanup()
 		}
 	}
 }
--- a/helper_test.go
+++ b/helper_test.go
@@ -653,18 +653,17 @@ func createVariable(t *testing.T, client *Client, w *Workspace) (*Variable, func
 	}
 
 	ctx := context.Background()
-	v, err := client.Variables.Create(ctx, VariableCreateOptions{
-		Key:       String(randomString(t)),
-		Value:     String(randomString(t)),
-		Category:  Category(CategoryTerraform),
-		Workspace: w,
+	v, err := client.Variables.Create(ctx, w.ID, VariableCreateOptions{
+		Key:      String(randomString(t)),
+		Value:    String(randomString(t)),
+		Category: Category(CategoryTerraform),
 	})
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	return v, func() {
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -39,7 +39,8 @@ Currently the following endpoints are supported:
 - [x] [Team Memberships](https://www.terraform.io/docs/enterprise/api/team-members.html)
 - [x] [Team Tokens](https://www.terraform.io/docs/enterprise/api/team-tokens.html)
 - [x] [Teams](https://www.terraform.io/docs/enterprise/api/teams.html)
-- [x] [Variables](https://www.terraform.io/docs/enterprise/api/variables.html)
+- [x] [Workspace Variables](https://www.terraform.io/docs/enterprise/api/workspace-variables.html)
+- [x] [Policy Set Parameters](https://www.terraform.io/docs/enterprise/api/policy-set-params.html)
```
svanharmelen: Maybe keep these (except for the `Admin` Entry) ordered? That helps keeping the list easy readable and searchable.

helper_test.go
```diff
@@ -648,6 +647,36 @@ func createVariable(t *testing.T, client *Client, w *Workspace) (*Variable, func
 	}
 }
 
+func createParameter(t *testing.T, client *Client, ps *PolicySet) (*Parameter, func()) {
```
svanharmelen: These helper functions are also (more or less) ordered...

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+// Compile-time proof of interface implementation.
+var _ Parameters = (*parameters)(nil)
+
+// Parameters describes all the parameter related methods that the Terraform
+// Enterprise API supports.
+//
+// TFE API docs: https://www.terraform.io/docs/enterprise/api/policy-set-params.html
+type Parameters interface {
```
svanharmelen: I know it's long, but for clarity I would call this `PolicySetParameters`

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// Update values of an existing parameter.
+	Update(ctx context.Context, policySetID string, parameterID string, options ParameterUpdateOptions) (*Parameter, error)
+
+	// Delete a parameter by its ID.
+	Delete(ctx context.Context, policySetID string, parameterID string) error
+}
+
+// parameters implements Parameters.
+type parameters struct {
```
svanharmelen: I know it's long, but for clarity I would call this `policySetParameters`

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	client *Client
+}
+
+// ParameterList represents a list of parameters.
+type ParameterList struct {
```
svanharmelen: I know it's long, but for clarity I would call this `PolicySetParameterList`

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	*Pagination
+	Items []*Parameter
+}
+
+// Parameter represents a Terraform Enterprise parameter.
+type Parameter struct {
```
svanharmelen: I know it's long, but for clarity I would call this `PolicySetParameter`

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	Value     string       `jsonapi:"attr,value"`
+	Category  CategoryType `jsonapi:"attr,category"`
+	Sensitive bool         `jsonapi:"attr,sensitive"`
+
+	// Relations
+	PolicySet *PolicySet `jsonapi:"relation,configurable"`
+}
+
+// ParameterListOptions represents the options for listing parameters.
+type ParameterListOptions struct {
```
svanharmelen: I know it's long, but for clarity I would call this `PolicySetParameterListOptions`

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+
+	u := fmt.Sprintf("policy-sets/%s/parameters", policySetID)
+	req, err := s.client.newRequest("GET", u, &options)
+	if err != nil {
+		return nil, err
+	}
+
+	vl := &ParameterList{}
```
svanharmelen:
```suggestion
	pl := &ParameterList{}
```

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	err = s.client.do(ctx, req, vl)
+	if err != nil {
+		return nil, err
+	}
+
+	return vl, nil
+}
+
+// ParameterCreateOptions represents the options for creating a new parameter.
+type ParameterCreateOptions struct {
```
svanharmelen: Same about the name...

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	// Make sure we don't send a user provided ID.
+	options.ID = ""
+
+	u := fmt.Sprintf("policy-sets/%s/parameters", url.QueryEscape(policySetID))
+	req, err := s.client.newRequest("POST", u, &options)
+	if err != nil {
+		return nil, err
+	}
+
+	v := &Parameter{}
```
svanharmelen:
```suggestion
	p := &Parameter{}
```

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, errors.New("invalid value for parameter ID")
+	}
+
+	u := fmt.Sprintf("policy-sets/%s/parameters/%s", url.QueryEscape(policySetID), url.QueryEscape(parameterID))
+	req, err := s.client.newRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	v := &Parameter{}
```
svanharmelen:
```suggestion
	p := &Parameter{}
```

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	// Make sure we don't send a user provided ID.
+	options.ID = parameterID
+
+	u := fmt.Sprintf("policy-sets/%s/parameters/%s", url.QueryEscape(policySetID), url.QueryEscape(parameterID))
+	req, err := s.client.newRequest("PATCH", u, &options)
+	if err != nil {
+		return nil, err
+	}
+
+	v := &Parameter{}
```
svanharmelen:
```suggestion
	p := &Parameter{}
```

parameter.go
```diff
@@ -0,0 +1,230 @@
+package tfe
```
svanharmelen: I would call this file `policy_set_parameter.go` to make it explicit what it is. Only `parameter` seems very generic.

tfe.go
```diff
@@ -129,6 +129,7 @@ type Client struct {
 	Users                      Users
 	Variables                  Variables
 	Workspaces                 Workspaces
+	Parameters                 Parameters
```
svanharmelen: Please order to according to the existing ordering and maybe rename it as well?

tfe.go
```diff
@@ -220,6 +221,7 @@ func NewClient(cfg *Config) (*Client, error) {
 	client.Users = &users{client: client}
 	client.Variables = &variables{client: client}
 	client.Workspaces = &workspaces{client: client}
+	client.Parameters = &parameters{client: client}
```
svanharmelen: Please order to according to the existing ordering and maybe rename it as well?

variable.go
```diff
@@ -43,6 +43,7 @@ type CategoryType string
 const (
 	CategoryEnv       CategoryType = "env"
 	CategoryTerraform CategoryType = "terraform"
+	CategoryPolicySet CategoryType = "policy-set"
```
svanharmelen: Please order to according to the existing ordering...

variable.go
```diff
@@ -61,33 +62,29 @@ type Variable struct {
 	Sensitive bool         `jsonapi:"attr,sensitive"`
 
 	// Relations
-	Workspace *Workspace `jsonapi:"relation,workspace"`
+	Workspace *Workspace `jsonapi:"relation,configurable"`
```
svanharmelen: Interesting, are you sure this works as expected?

variable.go
```diff
@@ -61,33 +62,29 @@ type Variable struct {
...
 
 func (o VariableListOptions) valid() error {
-	if !validString(o.Organization) {
-		return errors.New("organization is required")
-	}
-	if !validString(o.Workspace) {
-		return errors.New("workspace is required")
-	}
 	return nil
 }
```
svanharmelen: If this no longer checks anything, the method has to reason to exist. So in that case I would remove this all together.

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+// Parameter represents a Terraform Enterprise parameter.
+type Parameter struct {
+	ID        string       `jsonapi:"primary,vars"`
+	Key       string       `jsonapi:"attr,key"`
+	Value     string       `jsonapi:"attr,value"`
+	Category  CategoryType `jsonapi:"attr,category"`
+	Sensitive bool         `jsonapi:"attr,sensitive"`
+
+	// Relations
+	PolicySet *PolicySet `jsonapi:"relation,configurable"`
```
svanharmelen: Does `configurable` work as expected here?

parameter.go
```diff
@@ -0,0 +1,230 @@
...
+	// For internal use only!
+	ID string `jsonapi:"primary,vars"`
+
+	// The name of the parameter.
+	Key *string `jsonapi:"attr,key"`
+
+	// The value of the parameter.
+	Value *string `jsonapi:"attr,value,omitempty"`
+
+	// Whether this is a Terraform or environment parameter.
```
svanharmelen: Is that comment still true?

parameter_test.go
```diff
@@ -0,0 +1,285 @@
...
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	psTest, _ := createPolicySet(t, client, orgTest, nil, nil)
+
+	vTest1, _ := createParameter(t, client, psTest)
+	vTest2, _ := createParameter(t, client, psTest)
```
svanharmelen: I understand that this is all copy/pasted, but maybe rename all these to `pTest1`?

parameter_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	t.Run("without list options", func(t *testing.T) {
+		vl, err := client.Parameters.List(ctx, psTest.ID, ParameterListOptions{})
```
svanharmelen:
```suggestion
		pl, err := client.Parameters.List(ctx, psTest.ID, ParameterListOptions{})
```

Goes for the whole file of course.

parameter.go
cam-stitt: I believe so, yes. My tests are passing as I expect and I am unsure of any additional ways to ensure that this is working.

krantzinator:
Nice test coverage, thank you :raised_hands:

variable.go
svanharmelen: Seems you didn't answer this question yet. Can you share the reasoning for changing this as I noticed you used `configurable` in the new `PolicySetParams` struct as well?

README.md
```diff
@@ -39,7 +40,7 @@ Currently the following endpoints are supported:
 - [x] [Team Memberships](https://www.terraform.io/docs/enterprise/api/team-members.html)
 - [x] [Team Tokens](https://www.terraform.io/docs/enterprise/api/team-tokens.html)
 - [x] [Teams](https://www.terraform.io/docs/enterprise/api/teams.html)
-- [x] [Variables](https://www.terraform.io/docs/enterprise/api/variables.html)
+- [x] [Workspace Variables](https://www.terraform.io/docs/enterprise/api/workspace-variables.html)
```
svanharmelen: Ah, they really changed the concept behind this in the API. Nice... Guess it should be ordered below `Workspace` though 😉

variable.go
thrashr888: The relation name was changed in the API since it can be either workspaces or policy sets.

svanharmelen: Check... I'm a bit surprised that it works as expected in that case, but cool if it does 👍

cam-stitt: Sorry, I thought I had replied! The models behind the scene utilize a polymorphic relationship, which is based on a "configurable" type that can either be a workspace or policy set, as @thrashr888 said.

svanharmelen: Yeah, no problem. But I just wonder if this works as expected as I think to remember that the JSONAPI package doesn't work with polymorphic types (as it's all in Go right). But if it does and all the tests pass, then I guess I'm wrong and this is fine.

README.md
cam-stitt: I was simply following the alphabetical ordering. Plural `Workspaces` goes after `Workspace Variables` the same as `Teams` went after `Team ...`.

svanharmelen: My bad, your right 👍

variable.go
cam-stitt: I think it's most likely because of the way that the API is interacted with. The `configurable` isn't sent in the create/update requests and is actually handled within the API according to the route.

cam-stitt: I just added a couple of assertions to ensure that the policy set id is set and retrieved as expected.

PR #98, opened 2020-01-07
Add organization memberships resource

beekus:
This PR adds support for [Organization Memberships](https://www.terraform.io/docs/cloud/api/organization-memberships.html) to go-tfe.

### DIFF
--- a/organization_membership.go
+++ b/organization_membership.go
@@ -22,6 +22,12 @@ type OrganizationMemberships interface {
 	// Create a new organization membership with the given options.
 	Create(ctx context.Context, organization string, options OrganizationMembershipCreateOptions) (*OrganizationMembership, error)
 
+	// Read an organization membership by ID
+	Read(ctx context.Context, organizationMembershipID string) (*OrganizationMembership, error)
+
+	// Read an organization membership by ID with options
+	ReadWithOptions(ctx context.Context, organizationMembershipID string, options OrganizationMembershipReadOptions) (*OrganizationMembership, error)
+
 	// Delete an organization membership by its ID.
 	Delete(ctx context.Context, organizationMembershipID string) error
 }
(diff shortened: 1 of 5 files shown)


beekus:
I'm gonna do some more work on this and then re-open it.

beekus:
I've added support for the Read action, which is needed for the tfe-provider.

organization_membership.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// List all the organization memberships of the given organization.
+func (s *organizationMemberships) List(ctx context.Context, organization string, options OrganizationMembershipListOptions) (*OrganizationMembershipList, error) {
+	if !validStringID(&organization) {
+		return nil, errors.New("invalid value for organization")
+	}
+
+	u := fmt.Sprintf("organizations/%s/organization-memberships", url.QueryEscape(organization))
+	req, err := s.client.newRequest("GET", u, &options)
```
alisdair: I'm a bit surprised that we don't specify a default `include` here, but the `OrganizationMembership` struct has relationships to `User` and `Teams`. Won't those be partially-populated structs unless the user sets `include` correctly? I don't think that's broken behaviour, but it feels potentially surprising.

I had a quick look around for precedent in the library, and all I could find was this TeamMember example: https://github.com/hashicorp/go-tfe/blob/8b656f1e4c1b48738f180e9b171907f2dbc2c3c0/team_member.go#L44-L48

organization_membership.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+// Read an organization membership by its ID.
+func (s *organizationMemberships) Read(ctx context.Context, organizationMembershipID string) (*OrganizationMembership, error) {
+	return s.ReadWithOptions(ctx, organizationMembershipID, OrganizationMembershipReadOptions{})
```
alisdair: Same thought here—do we want the default options to include user and team?

beekus:
I went back and forth on including stuff by default, but eventually settled on giving the consumer of this package the option to specify, mainly because we have these nested structs in other places and don't automatically include things in those cases. The argument for having a default includes, to me, is really if we are looking at this as *only* for our use in the tfe-provider, or if we should look at this as more a generic client.

PR #100, opened 2020-01-23
Workspace variable descriptions

acespacecase:
Workspace variables now have descriptions (`terraform` and `env` categories) and this is the first step in exposing this functionality to the tfe provider.

Variable tests passing locally:
Screen Shot 2020-02-03 at 11 38 03 AM

### DIFF
--- a/variable_test.go
+++ b/variable_test.go
@@ -127,7 +127,7 @@ func TestVariablesCreate(t *testing.T) {
 		}
 
 		_, err := client.Variables.Create(ctx, wTest.ID, options)
-		assert.EqualError(t, err, "Invalid Attribute\n\nDescription is too long (maximum is 512 characters)")
+		assert.Error(t, err)
 	})
 
 	t.Run("when options is missing value", func(t *testing.T) {
--- a/variable_test.go
+++ b/variable_test.go
@@ -67,6 +67,7 @@ func TestVariablesCreate(t *testing.T) {
 			Key:      String(randomString(t)),
 			Value:    String(randomString(t)),
 			Category: Category(CategoryTerraform),
+			Description: String(randomString(t)),
 		}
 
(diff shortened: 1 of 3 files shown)


variable_test.go
```diff
@@ -93,9 +114,22 @@ func TestVariablesCreate(t *testing.T) {
...
+	t.Run("when options has a too-long description", func(t *testing.T) {
+		options := VariableCreateOptions{
+			Key:         String(randomString(t)),
+			Value:       String(randomString(t)),
+			Description: String("tortor aliquam nulla facilisi cras fermentum odio eu feugiat pretium nibh ipsum consequat nisl vel pretium lectus quam id leo in vitae turpis massa sed elementum tempus egestas sed sed risus pretium quam vulputate dignissim suspendisse in est ante in nibh mauris cursus mattis molestie a iaculis at erat pellentesque adipiscing commodo elit at imperdiet dui accumsan sit amet nulla facilisi morbi tempus iaculis urna id volutpat lacus laoreet non curabitur gravida arcu ac tortor dignissim convallis aenean et tortor"),
+			Category:    Category(CategoryTerraform),
+		}
+
+		_, err := client.Variables.Create(ctx, wTest.ID, options)
+		assert.EqualError(t, err, "Invalid Attribute\n\nDescription is too long (maximum is 512 characters)")
```
lafentres: i think we might not want to check for the exact error message returned from the api here.

from what i can tell, the existing tests only check for a specific error message for error messages that go-tfe is returning. i think this makes sense because if the error message changes in the api, this test will break. 

i do think you are right about checking this case though. i think you could check that an error is returned from the api, and not check specifically what the error message is. there are a few other tests that do that. the workspace tests for "when an error is returned from the api" or the user test "with invalid email address" do this.

PR #104, opened 2020-02-21
Feature: Admin/Users list

DeviaVir:
Hi! I needed the admin/users endpoint to automatically sync all users of an organization to a team in a cronjob, so I've added it here.

I've verified this works as you'd expect.

Though, I did notice that not every admin token is created equally- I am admin but my token kept returning `unauthorized`, I asked another admin to create a token and that did work. Not sure what that was about.

### DIFF
--- a/admin_users.go
+++ b/admin_users.go
@@ -1,0 +1,75 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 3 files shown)


chrisarcand:
Added in #189

PR #114, opened 2020-04-06
Fixup tests

leetrout:
This PR has 3 minor fixups:

- Organizations are not automatically in a trial upon creation so for CI (and most (all?) local test cases trial expiration will be null).
- The user update runs against the user running the tests so they will likely have a confirmed email so we fall back from `UnconfirmedEmail` to `Email` and failing if for some reason both are empty.
- Capacity is not available from the API so skipping for now (cc @beekus)

### DIFF
--- a/organization_test.go
+++ b/organization_test.go
@@ -204,6 +204,7 @@ func TestOrganizationsDelete(t *testing.T) {
 }
 
 func TestOrganizationsCapacity(t *testing.T) {
+	t.Skip("Capacity queues are not available in the API")
 	client := testClient(t)
 	ctx := context.Background()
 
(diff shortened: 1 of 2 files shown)


organization_test.go
```diff
@@ -280,6 +282,7 @@ func TestOrganizationsEntitlements(t *testing.T) {
 }
 
 func TestOrganizationsRunQueue(t *testing.T) {
+	t.Skip("Capacity queues are not available in the API")
```
lafentres, line 285: should the skip message here be different than the one above in the tests about capacity? or are capacity and run queues related?

totally fine if you don't have an answer for this, but are there plans to bring these things back in the api? just want to understand if we should eventually work toward removing or restructuring these tests.

beekus: The are related. They should be brought back eventually, but I think we've struggled with how queue things should be exposed in the UI, so the necessary API bits have been neglected as well. It doesn't seem like a high priority at the moment though.

PR #118, opened 2020-05-02
Team: Include option to be able to fetch relations

DeviaVir:
It is currently not possible to fetch relations (Users, AuthTokens) since you can't pass them in `Include`.  Comes down to a user of this API not being able to fetch a team's members.

It is not well documented [here](https://www.terraform.io/docs/cloud/api/teams.html) but it is documented [here](https://www.terraform.io/docs/cloud/api/admin/users.html#available-related-resources) and I verified it works.

### DIFF
--- a/team.go
+++ b/team.go
@@ -72,6 +72,8 @@ type TeamPermissions struct {
 // TeamListOptions represents the options for listing teams.
 type TeamListOptions struct {
 	ListOptions
+
+	Include string `url:"include"`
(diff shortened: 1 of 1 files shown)


chrisarcand:
The documentation you're looking for it at the bottom of that Teams page: https://www.terraform.io/docs/cloud/api/teams.html#available-related-resources

Thanks!

PR #129, opened 2020-07-30
Allow setting run comment

mwudka:
## Description

Almost every TFE API endpoint uses [jsonapi](https://jsonapi.org/), but a handful of endpoints expect plain JSON objects. Previously, requests to those endpoints would fail because the request would be formatted as a jsonapi request, and the endpoint expected a plain JSON object.

This change extends the request logic to inspect the tags of the request body object. If the fields are tagged with `jsonapi`, it uses the existing logic. Otherwise, it serializes the request body as a plain JSOn object.

## Testing plan

1. Ensure that applying/discarding runs with a comment works and the comment appears
1. Ensure that locking a workspace with a comment works and the comment appears

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -440,11 +440,18 @@ func (c *Client) newRequest(method, path string, v interface{}) (*retryablehttp.
 		reqHeaders.Set("Content-Type", "application/vnd.api+json")
 
 		if v != nil {
-			modelValue := reflect.ValueOf(v).Elem()
+
+			var modelType reflect.Type
+			if reflect.TypeOf(v).Kind() == reflect.Slice {
+				modelType = reflect.TypeOf(v).Elem().Elem()
+			} else {
+				modelType = reflect.ValueOf(v).Elem().Type()
+			}
+
 			jsonApiFields := 0
 			jsonFields := 0
-			for i := 0; i < modelValue.NumField(); i++ {
-				structField := modelValue.Type().Field(i)
+			for i := 0; i < modelType.NumField(); i++ {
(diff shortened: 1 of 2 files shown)


mwudka:
Far from silly, that question is one I wrestled with very much! I think there are two helpful ways to think about the change: conceptually good, and whether it preserves the good existing behavior.

For the former, the main change is essential to (pardon pseudocode) go from this:
```
request_to_send = json_api_serialize(request_object)
```
to this:
```
request_to_send = request_object_has_jsonapi_tags ? json_api_serialize(request_object) : plain_json_serialize(request_object)
```
I believe that is a good change to make, because only the top-level object should affect serialization. As a though experiment, if someone had a top-level object with all json tags that had a sub-struct with jsonapi tags they'd get garbage. But I think that is OK, because it doesn't really make sense to wrap a jsonapi-serialized thing inside a plain json thing; it'd be like embedding a SOAP envelope inside a different transport. Conversely, if someone had a top-level request object with jsonapi tags that contained regular json then everything is fine, which makes sense because it is totally reasonable to embed little plain json objects inside a larger request.

When viewed through the lense of whether the existing cases are broken, it helps to consider only the top-level objects. All but two of those use only jsonapi tags, so all of them will go through the old code path. Only the two structs that are tagged with plain json tags go through the new code path. Those cases were already broken, so at worst I've done now harm (I hope).

In short, the thing that spooked me is that I changed the broken structs, and then did a search for any `json` tags and found tons of hits. However, my mood swiftly rebounded when I realized that they were all child objects of regular `jsonapi` tags.

PR #143, opened 2020-10-02
Add AgentPoolID and Execution mode attributes to workspace

koikonom:


chrisarcand:
Closing in favor of a clean branch; this was also used in a demo and I want to avoid any issues with a ref being overwritten.

PR #144, opened 2020-10-05
README updates

chrisarcand:
## Description

Several README updates, including updating the test badge to the new CI platform we're using. Generally, cuts down on the overwhelming amount of info we're putting in this single doc by cutting out testing to it's own thing, generalizing the API support list (which is constantly out of date anyway), etc etc. Emphasizes that this is for TFC as well as Enterprise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Terraform Enterprise Go Client
 ==============================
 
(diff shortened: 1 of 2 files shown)


cam-stitt (approved):
Nice and clean.

PR #153, opened 2020-11-17
Add support for TFC Agent Pools

chrisarcand:
## Description

This adds support for [Terraform Cloud Agents](https://www.terraform.io/docs/cloud/workspaces/agent.html); specifically, managing [Agent Pools](https://www.terraform.io/docs/cloud/api/agents.html) and configuring them for use in a Terraform Cloud workspace. 

### TODO

- [x] Should add inline documentation concerning the fact that this is specific to Terraform Cloud (not Terraform Enterprise)

### DIFF
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -151,10 +151,21 @@ type AgentPoolUpdateOptions struct {
 	Name *string `jsonapi:"attr,name"`
 }
 
+func (o AgentPoolUpdateOptions) valid() error {
+	if o.Name != nil && !validStringID(o.Name) {
+		return errors.New("invalid value for name")
+	}
+	return nil
+}
+
 // Update an agent pool by its ID.
 func (s *agentPools) Update(ctx context.Context, agentPoolID string, options AgentPoolUpdateOptions) (*AgentPool, error) {
 	if !validStringID(&agentPoolID) {
 		return nil, errors.New("invalid value for agent pool ID")
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
 	}
 
 	// Make sure we don't send a user provided ID.
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -42,10 +42,13 @@ type AgentPoolList struct {
 	Items []*AgentPool
 }
 
-// AgentPool represents a Terraform Enterprise agent pool.
+// AgentPool represents a Terraform Cloud agent pool.
 type AgentPool struct {
 	ID   string `jsonapi:"primary,agent-pools"`
 	Name string `jsonapi:"attr,name"`
+
+	// Relations
+	Organization *Organization `jsonapi:"relation,organization"`
 }
 
 // AgentPoolListOptions represents the options for listing agent pools.
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -21,8 +21,14 @@ type AgentPools interface {
 	// Create a new agent pool with the given options.
 	Create(ctx context.Context, organization string, options AgentPoolCreateOptions) (*AgentPool, error)
 
-	// Read a agentpool by its ID.
-	Read(ctx context.Context, agentpoolID string) (*AgentPool, error)
+	// Read a agent pool by its ID.
+	Read(ctx context.Context, agentPoolID string) (*AgentPool, error)
+
+	// Update an agent pool by its ID.
+	Update(ctx context.Context, agentPool string, options AgentPoolUpdateOptions) (*AgentPool, error)
+
+	// Delete an agent pool by its ID.
+	Delete(ctx context.Context, agentPoolID string) error
 }
 
 // agentPools implements AgentPools.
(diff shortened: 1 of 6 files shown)


agent_pool.go
```diff
@@ -0,0 +1,190 @@
...
+	// A name to identify the agent pool.
+	Name *string `jsonapi:"attr,name"`
+}
+
+func (o AgentPoolCreateOptions) valid() error {
+	if !validString(o.Name) {
+		return errors.New("name is required")
+	}
+	return nil
+}
```
lafentres, line 92: an additional check to see if `name` contains any invalid characters would be good here. i think we do this for policy sets and workspaces, which have essentially the same naming convention. 
```suggestion
func (o AgentPoolCreateOptions) valid() error {
	if !validString(o.Name) {
		return errors.New("name is required")
	}
	if !validStringID(o.Name) {
		return errors.New("invalid value for name")
	}
	return nil
}
```

agent_pool.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// A new name to identify the agent pool.
+	Name *string `jsonapi:"attr,name"`
+}
+
+// Update an agent pool by its ID.
+func (s *agentPools) Update(ctx context.Context, agentPoolID string, options AgentPoolUpdateOptions) (*AgentPool, error) {
+	if !validStringID(&agentPoolID) {
+		return nil, errors.New("invalid value for agent pool ID")
+	}
```
lafentres, line 157: i think it would be good to validate the name in the options like we do for the create options.

agent_pool.go
```diff
@@ -0,0 +1,190 @@
...
+	Name string `jsonapi:"attr,name"`
+}
+
+// AgentPoolListOptions represents the options for listing agent pools.
+type AgentPoolListOptions struct {
+	ListOptions
+}
+
+// List all the agent pools of the given organization. Note this currently is
+// limited to a single pool per organization (API enforced).
```
lafentres, line 57: nit: remove the part about single agent pool
```suggestion
// List all the agent pools of the given organization.
```

agent_pool.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (o AgentPoolCreateOptions) valid() error {
+	if !validString(o.Name) {
+		return errors.New("name is required")
+	}
+	return nil
+}
+
+// Create a new agent pool with the given options. Note only a single pool is
+// allowed per organization (API enforced).
```
lafentres, line 95:
```suggestion
// Create a new agent pool with the given options.
```

agent_pool.go
```diff
@@ -0,0 +1,190 @@
...
+// TFE API docs: https://www.terraform.io/docs/cloud/api/agents.html
+type AgentPools interface {
+	// List all the agent pools of the given organization.
+	List(ctx context.Context, organization string, options AgentPoolListOptions) (*AgentPoolList, error)
+
+	// Create a new agent pool with the given options.
+	Create(ctx context.Context, organization string, options AgentPoolCreateOptions) (*AgentPool, error)
+
+	// Read a agent pool by its ID.
+	Read(ctx context.Context, agentPoolID string) (*AgentPool, error)
```
lafentres, line 25: do you think it is worth adding support for the `sort` and `includes` params on `List` and the `includes` param on `Read`?
edit: the more i think about it, probably not right now. this is mostly to support the provider work and those things are necessary for that and can be added later

workspace.go
```diff
@@ -165,9 +168,19 @@ type WorkspaceCreateOptions struct {
...
+	AgentPoolID *string `jsonapi:"attr,agent-pool-id,omitempty"`
+
 	// Whether to automatically apply changes when a Terraform plan is successful.
 	AutoApply *bool `jsonapi:"attr,auto-apply,omitempty"`
 
+	// Which execution mode to use. Valid values are remote, local, and agent.
+	// When set to local, the workspace will be used for state storage only.
+	// This value must not be specified if operations is specified.
+	ExecutionMode *string `jsonapi:"attr,execution-mode,omitempty"`
+
```
lafentres, line 183: nit: it might be good to mark the `Operations` attribute as deprecated in comments of the `Create` and `Update` options.

lafentres (approved):
everything looks good to me! i had a few minor comments/suggestions but that's all 🎉 🥳

agent_pool.go
chrisarcand: That's what `validStringID` does 👍🏻 Not that the name implies that in any way 😑 

https://github.com/hashicorp/go-tfe/blob/72179c6bdb6484b1550110cb3ac444e929466edf/validations.go#L15-L19

lafentres: we're currently only using validString here though right? 
i was suggesting we add validStringID as well (mostly because keeping them both separate give slightly better error messages and we do it like this elsewhere)

agent_pool.go
chrisarcand: Agreed

PR #161, opened 2021-01-12
Allow access to JSON version of state, and allow access to output variables that aren't strings

sfc-gh-ahess:
This PR represents two changes to the go-tfe library:
1. It provides a method for retrieving the JSON state of a workspace
2. It provides a mechanism for accessing state outputs that are not strings.  The built-in StateVersionOutputs assumes all outputs are strings, and panics when an array or other non-string output is encountered, which Terraform does support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/hashicorp/go-tfe
+module github.com/sfc-gh-ahess/go-tfe
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/plan.go
+++ b/plan.go
@@ -1,6 +1,7 @@
 package tfe
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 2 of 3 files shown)


chrisarcand (changes requested):
Hello, and thanks for looking at this!

This PR includes two very unrelated features; for the sake of review and historical context, could you please split off one of these to a separate pull request? Thanks.

---

Reviewing the state version output part only: The existing behavior should never have been done that way and this can be considered a bug; your additional `ReadSimpler()` needn't be added in an attempt to keep backwards compatibility. I believe `StateVersionOutput.Value` should implement [the Unmarshaler interface](https://golang.org/pkg/encoding/json/#Unmarshaler) and properly allow for more types than just strings, using a normal `Read()` call.

sfc-gh-ahess:
Going to create a two new PRs to separate these issues.

sfc-gh-ahess:
> Reviewing the state version output part only: The existing behavior should never have been done that way and this can be considered a bug; your additional `ReadSimpler()` needn't be added in an attempt to keep backwards compatibility. I believe `StateVersionOutput.Value` should implement [the Unmarshaler interface](https://golang.org/pkg/encoding/json/#Unmarshaler) and properly allow for more types than just strings, using a normal `Read()` call.

@chrisarcand : I feel like this is beyond the scope of a minor bugfix for me to make -- the unmarshaler interface would solve the retrieving of the data, but does not solve the issue of where to store it.  Does 'Value' remain an 'interface{}' ? Or do we create several value fields for every possible output type?  ValueInt? ValueString? ValueArray?

I think I will implement the latter to keep with strong typing.

PR #164, opened 2021-01-14
Add support for retrieving the JSON version of the plan

sfc-gh-ahess:
Implements the API function "JSON Output" for the plan and returns it as a byte array.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -1,6 +1,7 @@
 package tfe
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -22,6 +23,9 @@ type Plans interface {
 
 	// Logs retrieves the logs of a plan.
 	Logs(ctx context.Context, planID string) (io.Reader, error)
+	
+	// Provide the JSON of a plan as a string, for external policy checking.
+	JsonOutput(ctx context.Context, planID string) ([]byte, error)
 }
 
 // plans implements Plans.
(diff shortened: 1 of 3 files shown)


plan.go
```diff
@@ -134,3 +138,27 @@ func (s *plans) Logs(ctx context.Context, planID string) (io.Reader, error) {
...
+	if !validStringID(&planID) {
+		return nil, errors.New("invalid value for plan ID")
+	}
+
+	u := fmt.Sprintf("plans/%s/json-output", url.QueryEscape(planID))
+	// TODO: FInd a way to not have to use our own code to do this - maybe extend the API?
+	req, err := s.client.newRequest("GET", u, nil)
+	if err != nil {
+		return nil, fmt.Errorf("tfe: unable to get json plan %v", err)
+	}
```
radditude, line 154: When handling request errors elsewhere, we generally return the error from the API directly:

```
	if err != nil {
		return nil, err
	}
```

Is there a special reason we need extra error text here?

plan.go
```diff
@@ -134,3 +138,27 @@ func (s *plans) Logs(ctx context.Context, planID string) (io.Reader, error) {
...
+
+
+// Provide the JSON version of the plan from the TFE API.
+func (s *plans) JsonOutput(ctx context.Context, planID string) ([]byte, error) {
+	if !validStringID(&planID) {
+		return nil, errors.New("invalid value for plan ID")
+	}
+
+	u := fmt.Sprintf("plans/%s/json-output", url.QueryEscape(planID))
+	// TODO: FInd a way to not have to use our own code to do this - maybe extend the API?
```
radditude, line 150: I'm not quite sure I understand what this TODO comment is suggesting - could you clarify?

radditude (changes requested):
Thank you for splitting these out!

I'd love to add some tests for `Plans.JsonOutput`. You can see an example of tests for a similar method in [`TestStateVersionDownload`](https://github.com/hashicorp/go-tfe/blob/master/state_version_test.go#L386), though I wouldn't recommend comparing the full output against a static fixture file like we do there, since plan output will vary depending on the Terraform version used in the workspace.

Instead, I'd just check that the output is not empty and contains some string representative of JSON plan output, like `"address": "null_resource.foo"`, to confirm that retrieving the plan output was successful.

plan.go
sfc-gh-ahess: I've updated this to conform to your design preferences.

plan.go
sfc-gh-ahess: Sorry, it was left over from before, when I was doing it a different way.  I've removed it now.

sfc-gh-ahess:
I'm just implementing a missing part of your API here, unfortunately it is beyond my scope of work to do any additional development on this project.  Please feel free to take my code and add any additional changes that you require.

radditude:
Thanks for contributing! I'm going to close this out in favor of #187.

PR #165, opened 2021-01-14
Support for Terraform outputs of types other than String.

sfc-gh-ahess:
As requested by @chrisarcand, just updating the 'Read' and 'Value' to use the unmarshaler interface.

Background:
Terraform supports output types other than 'string'. For example, this Terraform declaration produces an output type of 'array':

output "instance_ips" {
  value = aws_instance.ubuntu.*.private_ip
}
resulting output:

instance_ips: [
  "10.1.2.3",
  "10.1.2.4",
  "10.1.2.5" 
]
This is why the State Output contains a "type" field, which will be "array" in this case.

However, the tfe-go library defines the State Version Output as:

type StateVersionOutput struct {
	ID        string `jsonapi:"primary,state-version-outputs"`
	Name      string `jsonapi:"attr,name"`
	Sensitive bool   `jsonapi:"attr,sensitive"`
	Type      string `jsonapi:"attr,type"`
	Value     string `jsonapi:"attr,value"`
}
Note that Value is 'string' only.

When .Read is called on an Output of type 'array', the Read call will fail with an error, as the jsonapi is not able to unmarshal the array value into this struct.

Here's an example API response from Terraform with an array:

{
  "data": {
    "id": "wsout-abcdef",
    "type": "state-version-outputs",
    "attributes": {
      "name": "instance_ips",
      "sensitive": false,
      "type": "array",
      "value": [
        "10.180.69.26",
        "10.180.69.11",
        "10.180.69.37"
      ]
    },
    "links": {
      "self": "/api/v2/state-version-outputs/wsout-abcdef"
    }
  }
}

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/hashicorp/go-tfe
+module github.com/sfc-gh-ahess/go-tfe
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/sfc-gh-ahess/go-tfe
+module github.com/hashicorp/go-tfe
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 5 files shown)


sfc-gh-ahess:
Sorry, forgot one thing.

Issue #158, opened 2020-12-18
StateVersionOutput expects all outputs to be strings, and fails when other output types exist


sfc-gh-ahess:
Terraform supports output types other than 'string'.  For example, this Terraform declaration produces an output type of 'array':

```
output "instance_ips" {
  value = aws_instance.ubuntu.*.private_ip
}
```

resulting output:
```
instance_ips: [
  "10.1.2.3",
  "10.1.2.4",
  "10.1.2.5" 
]
```

This is why the State Output contains a "type" field, which will be "array" in this case.

However, the tfe-go library defines the State Version Output as:
```
type StateVersionOutput struct {
	ID        string `jsonapi:"primary,state-version-outputs"`
	Name      string `jsonapi:"attr,name"`
	Sensitive bool   `jsonapi:"attr,sensitive"`
	Type      string `jsonapi:"attr,type"`
	Value     string `jsonapi:"attr,value"`
}
```

Note that Value is 'string' only. 

When .Read is called on an Output of type 'array', the Read call will fail with an error, as the jsonapi is not able to unmarshal the array value into this struct.

Here's an example API response from Terraform with an array:

```
{
  "data": {
    "id": "wsout-abcdef",
    "type": "state-version-outputs",
    "attributes": {
      "name": "instance_ips",
      "sensitive": false,
      "type": "array",
      "value": [
        "10.180.69.26",
        "10.180.69.11",
        "10.180.69.37"
      ]
    },
    "links": {
      "self": "/api/v2/state-version-outputs/wsout-abcdef"
    }
  }
}
```

Thanks!
-alex

rorsten:
I'm hitting this as well, with a state output that's an integer. And I can think of some workspaces where the outputs are maps. IIRC in the case of maps there's also a schema field? I would be quite satisfied with getting something from cty or even a json.RawMessage of the field.

sfc-gh-ahess:
I've fixed this in a fork of the repo, you can see my PR here:
https://github.com/hashicorp/go-tfe/pull/161

omarismail:
Hey @sfc-gh-ahess, we know this has been a long standing issue and apologies for the delay. We have addressed this problem in #246. Now the type of `StateVersionOutput.Value` is an `interface{}` and you can type cast it into anything you'd like.

PR #166, opened 2021-01-14
Support for State outputs that aren't just strings

sfc-gh-ahess:
Updated to use Unmarshaler interface as requested by @chrisarcand 

Terraform supports output types other than 'string'. For example, this Terraform declaration produces an output type of 'array':

output "instance_ips" {
  value = aws_instance.ubuntu.*.private_ip
}
resulting output:

instance_ips: [
  "10.1.2.3",
  "10.1.2.4",
  "10.1.2.5" 
]
This is why the State Output contains a "type" field, which will be "array" in this case.

However, the tfe-go library defines the State Version Output as:

type StateVersionOutput struct {
	ID        string `jsonapi:"primary,state-version-outputs"`
	Name      string `jsonapi:"attr,name"`
	Sensitive bool   `jsonapi:"attr,sensitive"`
	Type      string `jsonapi:"attr,type"`
	Value     string `jsonapi:"attr,value"`
}
Note that Value is 'string' only.

When .Read is called on an Output of type 'array', the Read call will fail with an error, as the jsonapi is not able to unmarshal the array value into this struct.

Here's an example API response from Terraform with an array:

{
  "data": {
    "id": "wsout-abcdef",
    "type": "state-version-outputs",
    "attributes": {
      "name": "instance_ips",
      "sensitive": false,
      "type": "array",
      "value": [
        "10.180.69.26",
        "10.180.69.11",
        "10.180.69.37"
      ]
    },
    "links": {
      "self": "/api/v2/state-version-outputs/wsout-abcdef"
    }
  }
}

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/hashicorp/go-tfe
+module github.com/sfc-gh-ahess/go-tfe
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/sfc-gh-ahess/go-tfe
+module github.com/hashicorp/go-tfe
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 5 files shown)


omarismail:
Hey @sfc-gh-ahess , thank you for attempting this PR and we know this has been a long standing issue. This has been addressed in #246. Now the type of `StateVersionOutput.Value` is an `interface{}` and you can type cast it into anything you'd like. I also commented on the [issue you opened up](https://github.com/hashicorp/go-tfe/issues/158#issuecomment-902650552) as well.

PR #176, opened 2021-02-18
Rename Apply file

omarismail:
## Description

Renaming the `apply` file to correspond to the `/api/v2/applies` endpoint. 

[API Docs](https://www.terraform.io/docs/cloud/api/applies.html)


radditude:
I'm not sure I'm entirely following the reasoning for this change! We use the singular for all resources in this SDK - [`plan.go`](https://github.com/hashicorp/go-tfe/blob/master/plan.go), for instance - even though the actual path for the API endpoint/docs generally uses the plural (https://www.terraform.io/docs/cloud/api/plans.html for plans).

Is there some reason `applies` should be its own special case that uses a different naming convention?

omarismail:
@radditude you're right, I didn't look through all the file names and saw a couple others that were plural (ex `ip-ranges`), and quickly assumed there wasn't consistency. But looks like singular is the pattern.

Closing this out.

radditude:
Ah, yeah, `ip-ranges` is the odd one out since it isn't actually a CRUD resource!

PR #177, opened 2021-02-19
Extracted errors to separte file, and cleaned up import structures

omarismail:
## Description

The purpose of this PR is to clean up a bit of error handling. 

Having dedicated errors variables instead of using hard coded strings avoids doing [shotgun surgery](https://refactoring.guru/smells/shotgun-surgery). If we want to improve the error handling (a later followup to this), we will need to change strings in every location. But now, by having one variable, we can change that one string and it applies everywhere.

This change does three things:
1. creates an errors.go file that holds these variables.
2. Moves the errors from `tfe.go` into `errors.go`.
3. find all and replace where the errors for validation are used. 
4. Drive by clean up in making sure the imports for all files follows the pattern of: first is standard library and the following that is external pacakges.
5. fix some formatting indentation 

## Follow up

This does not change all errors, just a few of them. A follow up PR will go through all errors and organize them in this manner. 

## Testing plan

Run the tests and make sure they all pass. 

## Output from tests

_verifying tests_

### DIFF
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -59,7 +59,7 @@ type AgentPoolListOptions struct {
 // List all the agent pools of the given organization.
 func (s *agentPools) List(ctx context.Context, organization string, options AgentPoolListOptions) (*AgentPoolList, error) {
 	if !validStringID(&organization) {
-		return nil, errors.New("invalid value for organization")
+		return nil, ErrInvalidOrg
 	}
 
 	u := fmt.Sprintf("organizations/%s/agent-pools", url.QueryEscape(organization))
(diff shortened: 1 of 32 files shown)


chrisarcand (commented):
* Is there a reason you moved the assert functions over to `Error` rather than `EqualError`?
* Let's close this PR and have you create a branch in the origin repository and reopen so that CI can run.

omarismail:
> Is there a reason you moved the assert functions over to Error rather than EqualError

It looked cleaner to compare errors directly, instead of doing something like
```
```

But i'm looking at the source of `assert.Equal`, and I'm trying to test, in Go Playground, the byte comparison and/or `reflect` usage, and am coming up short.

On second thought, I think the verboseness of using `EqualError(t, err, ErrInvalidOrg.Error())` seems more apparent then the type checking in `assert.Equal`. 

Will close this PR and open in origin, and also update the assert!

PR #182, opened 2021-02-24
Oauth: Add missing uuid and secret fields

samm-git:
## Description

My intention is to extend TFE provider to add missing secret and uuid fields. This would allow me to automate vcs connection creation for the TFE orgs. Should also be relevant to TFC users. 

## Testing plan

1. Create VCS connection to the github.com provider using TFE GUI.
2. Terraform UI will generate UUID in advance and show it to user as part of oauth callback
3. There will be input-field to enter GH secret

Sample API request provided below:

```
{
  "data": {
    "attributes": {
      "name": "Example GH",
      "created-at": null,
      "callback-url": null,
      "oauth-token-string": null,
      "connect-path": null,
      "key": "XXXXX",
      "secret": "XXXXXXXX",
      "rsa-public-key": null,
      "uuid": "82213e23-3d48-4061-8960-3c01ef3e2d6e",
      "service-provider": "github",
      "service-provider-display-name": null,
      "api-url": "https://api.github.com",
      "http-url": "https://github.com"
    },
    "relationships": {
      "organization": {
        "data": {
          "type": "organizations",
          "id": "sampl-org-gh"
        }
      }
    },
    "type": "oauth-clients"
  }
}
```

At the moment it is not possible to do this programmatically via go-tfe, as `uuid` and `secrets` fields are missing.

Also i found that API documentation includes secret only on PATCH method and not the POST one, however for POST it is working as well (tested with `curl` and also visible from TFE UI requests)
## External links

- [TFE API documentation](https://www.terraform.io/docs/cloud/api/oauth-clients.html)
- PR to update documentation: https://github.com/hashicorp/terraform-website/pull/1635

### DIFF
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -129,7 +129,7 @@ type OAuthClientCreateOptions struct {
 	Secret *string `jsonapi:"attr,secret"`
 
 	// UUID of the vcs connection. If empty - automatically generated.
-	Uuid *string `jsonapi:"uuid,secret"`
+	Uuid *string `jsonapi:"attr,uuid"`
 
 	// The VCS provider being connected with.
 	ServiceProvider *ServiceProviderType `jsonapi:"attr,service-provider"`
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -70,6 +70,8 @@ type OAuthClient struct {
 	CreatedAt           time.Time           `jsonapi:"attr,created-at,iso8601"`
 	HTTPURL             string              `jsonapi:"attr,http-url"`
 	Key                 string              `jsonapi:"attr,key"`
+	Secret              string              `jsonapi:"attr,secret"`
+	Uuid                string              `jsonapi:"attr,uuid"`
 	RSAPublicKey        string              `jsonapi:"attr,rsa-public-key"`
 	ServiceProvider     ServiceProviderType `jsonapi:"attr,service-provider"`
 	ServiceProviderName string              `jsonapi:"attr,service-provider-display-name"`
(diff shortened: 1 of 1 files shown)


samm-git:
@mwhooker any chance to get this merged?

chrisarcand:
Hi!

> This would allow me to automate vcs connection creation for the TFE orgs.

This is already possible today without these fields. In your automation, create an OAuth application in your VCS provider and provide an OAuth token here, and you should be good to go.

These particular fields (`secret` and `uuid`) are present in the API primarily to support Terraform Cloud's user interface. Both `uuid` and `secret` are not fields that are ever returned in an OAuth client response that would be useful for 3rd party clients (`uuid` doesn't exist, and `secret` is always `null`).

Using this API endpoint and client - rather than using these fields which are used in the process of creating a GitHub/GitLab/Azure DevOps Server/etc OAuth Application - you provide a pre-generated OAuth token string instead (that's what `private-key` and `oauth-token-string` are for).

Thanks!

samm-git:
Hi @chrisarcand , i still cant get how to use it in automation, sorry. 

My intention is to configure connection up to the stage where user have to go to interface and finish connection from his account. So i must use secret field, if i will not do it - tfe will not allow user to establish connection.

I am TFE customer and will be happy to discuss this usecase if possible. Thank you.

samm-git:
@chrisarcand just to make it very clear - my intention is to provide half-configured vcs connection, so user will click "connect" in the org. Goal is to make sure that vcs link is established from user account, not from system one. As org shared between many teams and we have to make sure that user can use it only for repositories he have access to.

chrisarcand:
Your use case of wanting a half-configured VCS connection is interesting! A part I'm a little blurry on is that if the user has access to hit that 'Connect' button in the org, it means they have permission to manage VCS settings in which case they can create their own OAuth client from the start - meaning you could automate the creation of the OAuth app in the VCS provider for them using their token, making the resulting OAuth secret not sensitive (because they created the app) and allowing you to further automate using that generated secret to create the OAuth client in TFC.

I will say that as it stands now that API is not considered stable (that's why it's not documented) and as I mentioned above is in service of the TFC user interface only; it needs to be stable for inclusion here in the client and in documentation.

As a TFE customer we would _love_ to discuss your usecase more, and I highly encourage you to open a support ticket to get this work triaged accordingly or help in crafting a detailed alternative solution like I described above.

Thank you!

PR #183, opened 2021-02-26
Admin Workspaces API

omarismail:
## Description

This PR adds the [Admin Workspaces API](https://www.terraform.io/docs/cloud/api/admin/workspaces.html).

The current implemented methods are List, Read, and Delete

## Testing plan

Run tests against an Terraform Enterprise Instance.

## Output from tests

```
$ TFE_TOKEN=<token> TFE_ADDRESS=<adrress-of-tfe-instance>  ENABLE_TFE=1 go test -v ./...  -timeout=30m
...
PASS
ok  	github.com/hashicorp/go-tfe	485.784s

```

## External links

- [Asana](https://app.asana.com/0/1199201948575144/1199972599424003)
- [Admin Workspace API Docs](https://www.terraform.io/docs/cloud/api/admin/workspaces.html)

### DIFF
--- a/admin_workspace.go
+++ b/admin_workspace.go
@@ -17,11 +17,11 @@ type AdminWorkspaces interface {
 	// List all the workspaces within an organization.
 	List(ctx context.Context, options AdminWorkspaceListOptions) (*AdminWorkspaceList, error)
 
-	// Read a workspace by its name.
-	Read(ctx context.Context, workspace string) (*AdminWorkspace, error)
+	// Read a workspace by its ID.
+	Read(ctx context.Context, workspaceID string) (*AdminWorkspace, error)
 
-	// Delete a workspace by its name.
-	Delete(ctx context.Context, workspace string) error
+	// Delete a workspace by its ID.
+	Delete(ctx context.Context, workspaceID string) error
 }
 
 // adminWorkspaces implements AdminWorkspaces.
@@ -46,6 +46,7 @@ type AdminWorkspaceListOptions struct {
 	ListOptions
 
 	// A query string (partial workspace name) used to filter the results.
+	// https://www.terraform.io/docs/cloud/api/admin/workspaces.html#query-parameters
 	Query *string `url:"q,omitempty"`
 
 	// A list of relations to include. See available resources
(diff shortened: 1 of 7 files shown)


admin_workspace.go
```diff
@@ -0,0 +1,114 @@
...
+type adminWorkspaces struct {
+	client *Client
+}
+
+// AdminWorkspaces represents a Terraform Enterprise admin workspace.
+type AdminWorkspace struct {
+	ID      string   `jsonapi:"primary,workspaces"`
+	Name    string   `jsonapi:"attr,name"`
+	Locked  bool     `jsonapi:"attr,locked"`
+	VCSRepo *VCSRepo `jsonapi:"attr,vcs-repo"`
```
chrisarcand, line 37: It looks like the API only returns the `identifier` for the this and not all the fields of this type will be populated; should we use some other simple type besides `*VCSRepo` to denote this, or would we be boxing ourselves in a corner doing something like `map[string]interface{}`? Whatcha think?

chrisarcand (changes requested):
Looks great! The `include` error is my only blocker.

errors.go
```diff
@@ -12,6 +12,9 @@ var (
 	// ErrResourceNotFound is returned when a receiving a 404.
 	ErrResourceNotFound = errors.New("resource not found")
 
+	// ErrInvalidIncludeParam is returned when an 'include' parameter is invalid.
+	ErrInvalidIncludeParam = errors.New("bad request\n\nInvalid include parameter")
```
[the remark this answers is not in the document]
omarismail: Very fair point. I'll move it over to the test

admin_workspace.go
omarismail: Good catch! Looking at the serializers in TFC, `vcs_repo` is only used once for all Admin endpoints, so I think just nesting the struct here would do. 

```
// AdminWorkspaces represents a Terraform Enterprise admin workspace.
type AdminWorkspace struct {
	ID      string `jsonapi:"primary,workspaces"`
	Name    string `jsonapi:"attr,name"`
	Locked  bool   `jsonapi:"attr,locked"`
	VCSRepo *struct {
		Identifier string `json:"identifier"`
	} `jsonapi:"attr,vcs-repo"`
}
```

chrisarcand (approved):
Looks great!! ✅ :shipit:

PR #185, opened 2021-03-04
Add Admin Settings API

omarismail:
## Description

This PR adds all the [Admin Settings API](https://www.terraform.io/docs/cloud/api/admin/settings.html).

It implements the API for each Settings:
* General.Read
* General.Update
* CostEstimation.Read
* CostEstimation.Update
* SAML.Read
* SAML.Update
* SAML.RevokeIdpCert
* SMTP.Read
* SMTP.Update
* Twilio.Read
* Twilio.Update
* Twilio.Verify
* Customization.Read
* Customization.Update

## Testing plan

1. Pull down and run the tests 

## Output from tests

```
$ TFE_TOKEN=<token> TFE_ADDRESS=<address>  ENABLE_TFE=1 go test -v ./...  -timeout=30m
...
PASS
ok  	github.com/hashicorp/go-tfe	510.513s
```

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/admin/settings.html)


petems:
Awesome, thanks @omarismail! 🤝

admin_organization_test.go
```diff
@@ -22,8 +22,10 @@ func TestAdminOrganizations_List(t *testing.T) {
 		adminOrgList, err := client.Admin.Organizations.List(ctx, AdminOrganizationListOptions{})
 		require.NoError(t, err)
 
-		assert.Equal(t, true, adminOrgItemsContainsName(adminOrgList.Items, org.Name))
-		assert.Equal(t, 1, adminOrgList.CurrentPage)
+		// Given that org creation occurs on every test, the ordering is not
+		// gauranteed. It may be that the `org` created in this test does not appear
```
chrisarcand, line 26: Ah nice, good catch.

```suggestion
		// guaranteed. It may be that the `org` created in this test does not appear
```

admin_setting.go
```diff
@@ -0,0 +1,461 @@
...
+// Compile-time proof of interface implementation.
+var _ AdminSettings = (*adminSettings)(nil)
+
+// AdminSettings describes all the admin settings related methods that the Terraform Enterprise API supports.
+// Note that admin settings are only available in Terraform Enterprise.
+//
+// TFE API docs: https://www.terraform.io/docs/cloud/api/admin/settings.html
+type AdminSettings interface {
+	// GetGeneral returns the general settings
+	GetGeneral(ctx context.Context) (*AdminGeneralSetting, error)
```
chrisarcand, line 16: I think `Read*` for getter funcs might be a little more consistent with the rest of the API. In spelunking history for one of your other PRs I actually noted this was changed from `Fetch` intentionally, too!

admin_setting.go
```diff
@@ -0,0 +1,461 @@
...
+
+	// VerifyTwilio verifies Twilio settings.
+	VerifyTwilio(ctx context.Context, options AdminTwilioSettingsVerifyOptions) error
+
+	// GetCustomization returns the customization settings.
+	GetCustomization(ctx context.Context) (*AdminCustomizationSetting, error)
+
+	// UpdateCustomization updates the customization settings.
+	UpdateCustomization(ctx context.Context, options AdminCustomizationSettingsUpdateOptions) (*AdminCustomizationSetting, error)
+}
```
chrisarcand, line 57: Random thought, take or leave it: Does it make sense, especially as more and more stuff will undoubtedly be added to this, to further differentiate areas by one more layer of structing? `Client.Admin.Settings.SAML.Read`, `Client.Admin.Settings.SAML.RevokeIdpCert`, etc? I'm torn really and could go either way, but thought I'd bring it up to ponder about at least.

chrisarcand: One thing lending to this organization is something I didn't notice before, which is that the "Settings API" is actually a collection of a bunch of different ones 🤔 `admin/cost-estimation-settings`, `admin/saml-settings`, etc.

admin_setting.go
```diff
@@ -0,0 +1,461 @@
...
+
+// AdminGeneralSetting represents a the general settings in Terraform Enterprise.
+type AdminGeneralSetting struct {
+	ID                            string `jsonapi:"primary,general-settings"`
+	LimitUserOrganizationCreation bool   `jsonapi:"attr,limit-user-organization-creation"`
+	APIRateLimitingEnabled        bool   `jsonapi:"attr,api-rate-limiting-enabled"`
+	APIRateLimit                  int    `jsonapi:"attr,api-rate-limit"`
+	SendPassingStatusesEnabled    bool   `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
+	AllowSpeculativePlansOnPR     bool   `jsonapi:"attr,allow-speculative-plans-on-pull-requests-from-forks"`
```
chrisarcand, line 71: Seems like there's quite a few settings missing from this list; the documentation is well out of date 😕 You'll want to double check what's actually there and compare that to here.

admin_setting.go
```diff
@@ -0,0 +1,461 @@
...
+type AdminCostEstimationSetting struct {
+	ID                  string `jsonapi:"primary,cost-estimation-settings"`
+	Enabled             bool   `jsonapi:"attr,enabled"`
+	AWSAccessKeyID      string `jsonapi:"attr,aws-access-key-id"`
+	AWSAccessKey        string `jsonapi:"attr,aws-secret-key"`
+	GCPCredentials      string `jsonapi:"attr,gcp-credentials"`
+	AzureClientID       string `jsonapi:"attr,azure-client-id"`
+	AzureClientSecret   string `jsonapi:"attr,azure-client-secret"`
+	AzureSubscriptionID string `jsonapi:"attr,azure-subscription-id"`
+	AzureTenantID       string `jsonapi:"attr,azure-tenant-id"`
```
chrisarcand, line 127: I think this (and Options) is missing `aws-instance-profile-enabled` ?

chrisarcand (changes requested):
Lookin' good!

I skimmed a lot and had enough feedback to stop and revisit in more detail a second time around; the biggest bit is to double check all these APIs to see if they're all covered in the structs; I think a lot of the documentation around these APIs is woefully out of date 😩

admin_setting.go
omarismail: Thats a very fair point. I think thats a better approach. I'll update

admin_setting.go
omarismail: Will do

omarismail: Added

admin_setting.go
omarismail: Added

admin_setting_smtp.go
```diff
@@ -0,0 +1,106 @@
...
+func (o AdminSMTPSettingsUpdateOptions) valid() bool {
+	if !validString(o.Auth) {
+		return false
+	}
+
+	validSMTPAuthType := map[string]int{
+		string(SMTPAuthNone):  1,
+		string(SMTPAuthPlain): 1,
+		string(SMTPAuthLogin): 1,
+	}
```
chrisarcand, line 102: Prefer `map[SMTPAuthType]struct{}` constructed outside the function.

* Key type can more specifically use the type you declared for these values
* Value type, whenever you're doing a validation like this and don't actually care about the keys, should be a `struct{}{}`. No space needs to be allocated for those values (ints do, empty structs don't), and (though this is getting more pedantic, I know) you then can't assign any int, potentially abusing the map beyond its intended purpose.
* Defined outside the function, it doesn't need to be constructed every invocation.

chrisarcand: Alternatively, if you're ok with the lack of O(1)....

```go
switch *o.Auth {
case SMTPAuthNone:
case SMTPAuthPlain:
case SMTPAuthLogin:
    return true
}
return false
```

Either way!

chrisarcand (approved):
Two quick nits and this is ready to merge!

tfe.go
```diff
@@ -138,6 +138,7 @@ type Admin struct {
 	Workspaces        AdminWorkspaces
 	Runs              AdminRuns
 	TerraformVersions AdminTerraformVersions
+  Settings          *AdminSettings
```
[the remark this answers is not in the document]
omarismail: Oops, I don't know how that one got through haha.

admin_setting_smtp.go
omarismail: Good point on the use of `struct{}{}` as value. I'll update

PR #186, opened 2021-03-05
Add Admin Terraform Versions API

omarismail:
## Description

This PR adds all the [Admin TerraformVersions API](https://www.terraform.io/docs/cloud/api/admin/terraform-versions.html).

It implements:
* List: lists all terraform versions
* Read: reads a specific version
* Create: creates a terraform version
* Update: updates a specific terraform version
* Delete: deletes a terraform version

## Testing plan

1. Pull down the code and run the tests.

## Output from tests

```
$ TFE_TOKEN=<token> TFE_ADDRESS=<address>  ENABLE_TFE=1 go test -v ./...  -timeout=30m
PASS
ok  	github.com/hashicorp/go-tfe	533.924s

```

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/admin/terraform-versions.html)

### DIFF
--- a/admin_terraform_version.go
+++ b/admin_terraform_version.go
@@ -62,14 +62,14 @@ type AdminTerraformVersionsList struct {
 }
 
 // List all the terraform versions.
-func (s *adminTerraformVersions) List(ctx context.Context, options AdminTerraformVersionsListOptions) (*AdminTerraformVersionsList, error) {
-	req, err := s.client.newRequest("GET", "admin/terraform-versions", &options)
+func (a *adminTerraformVersions) List(ctx context.Context, options AdminTerraformVersionsListOptions) (*AdminTerraformVersionsList, error) {
+	req, err := a.client.newRequest("GET", "admin/terraform-versions", &options)
 	if err != nil {
 		return nil, err
 	}
 
 	awl := &AdminTerraformVersionsList{}
-	err = s.client.do(ctx, req, awl)
+	err = a.client.do(ctx, req, awl)
 	if err != nil {
 		return nil, err
 	}
@@ -78,19 +78,19 @@ func (s *adminTerraformVersions) List(ctx context.Context, options AdminTerrafor
 }
 
 // Read a terraform version by its ID.
-func (s *adminTerraformVersions) Read(ctx context.Context, id string) (*AdminTerraformVersion, error) {
+func (a *adminTerraformVersions) Read(ctx context.Context, id string) (*AdminTerraformVersion, error) {
 	if !validStringID(&id) {
 		return nil, ErrInvalidTerraformVersionID
 	}
 
 	u := fmt.Sprintf("admin/terraform-versions/%s", url.QueryEscape(id))
-	req, err := s.client.newRequest("GET", u, nil)
+	req, err := a.client.newRequest("GET", u, nil)
 	if err != nil {
 		return nil, err
 	}
 
 	tfv := &AdminTerraformVersion{}
-	err = s.client.do(ctx, req, tfv)
+	err = a.client.do(ctx, req, tfv)
 	if err != nil {
 		return nil, err
 	}
@@ -121,18 +121,18 @@ func (o AdminTerraformVersionCreateOptions) valid() error {
 }
 
 // Create a new terraform version.
-func (s *adminTerraformVersions) Create(ctx context.Context, options AdminTerraformVersionCreateOptions) (*AdminTerraformVersion, error) {
+func (a *adminTerraformVersions) Create(ctx context.Context, options AdminTerraformVersionCreateOptions) (*AdminTerraformVersion, error) {
 	if err := options.valid(); err != nil {
 		return nil, err
 	}
 
-	req, err := s.client.newRequest("POST", "admin/terraform-versions", &options)
+	req, err := a.client.newRequest("POST", "admin/terraform-versions", &options)
 	if err != nil {
 		return nil, err
 	}
 
 	tfv := &AdminTerraformVersion{}
-	err = s.client.do(ctx, req, tfv)
+	err = a.client.do(ctx, req, tfv)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


admin_terraform_version.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return tfv, nil
+}
+
+// AdminTerraformVersionCreateOptions for creating a terraform version.
+// https://www.terraform.io/docs/cloud/api/admin/terraform-versions.html#request-body
+type AdminTerraformVersionCreateOptions struct {
```
chrisarcand, line 103: Every CreateOptions type in this project has a `ID` field (`string`) "for internal use only!"

...I've never actually have a clue what it's for, though, so I went on a spelunking session and [found when this was added and discussion around it](https://github.com/hashicorp/go-tfe/pull/1#discussion_r200041679)

Can you dig in to that and report back if that actually matters anymore? I don't immediately see why `jsonapi` needs this field at all. If it's not clear and we want to just move forward, I'd just add it for consistency's sake and we can revisit separately and remove them _all_ if it makes sense (it's always been marked as internal private API only that we can totally rip out if it doesn't make sense).

admin_terraform_version.go
```diff
@@ -0,0 +1,203 @@
...
+	Type     *string `jsonapi:"primary,terraform-versions"`
```
chrisarcand, line 104: This is actually so awkward (in the existing API, I mean) that I'd go so far as to remove it entirely here in the client. The type here is a vestige of HashiCorp Atlas, when these 'tools' could be more than just Terraform.

As it's very purpose-built for Terraform now (it's even encoded in the route as `terraform-versions`), I think we should just remove this and provide it implicitly to the API (which I'm surprised doesn't just default to `terraform-versions`), not requiring the client user to have to think about a thing they have to provide and literally can't be anything _but_ `terraform-versions` 😆 

I'd either just document on the field that it's ignored and assign the required value in the functions when using the struct, or do the actual request with some sort of private struct that includes it (leaving it out of this public interface); whatcha think?

admin_terraform_version.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (o AdminTerraformVersionCreateOptions) valid() error {
+	if validStringID(o.Type) {
+		if *o.Type == "terraform-version" {
+			return nil
+		}
+	}
+
+	return ErrInvalidTerraformVersionType
+}
```
chrisarcand, line 121: (With the above feedback, stuff like this can go away)

chrisarcand (changes requested):
Sooo excited for this one.

admin_terraform_version.go
```diff
@@ -0,0 +1,203 @@
...
+type AdminTerraformVersion struct {
+	ID        string `jsonapi:"primary,terraform-versions"`
+	Version   string `jsonapi:"attr,version"`
+	URL       string `jsonapi:"attr,url"`
+	Sha       string `jsonapi:"attr,sha"`
+	Official  bool   `jsonapi:"attr,official"`
+	Enabled   bool   `jsonapi:"attr,enabled"`
+	Beta      bool   `jsonapi:"attr,beta"`
+	Usage     int    `jsonapi:"attr,usage"`
+	CreatedAt string `jsonapi:"attr,created-at"`
```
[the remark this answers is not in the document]
omarismail: Oops, right, will update.

admin_terraform_version.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// List all the terraform versions.
+func (a *adminTerraformVersions) List(ctx context.Context, options AdminTerraformVersionsListOptions) (*AdminTerraformVersionsList, error) {
+	req, err := a.client.newRequest("GET", "admin/terraform-versions", &options)
+	if err != nil {
+		return nil, err
+	}
+
+	awl := &AdminTerraformVersionsList{}
```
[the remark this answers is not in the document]
omarismail: copy pasta 😅

admin_terraform_version.go
omarismail: Here are a few findings.

1. Checked a number of resources (Organizations, Variables, Workspaces, StateVersions) and when we create those resources, we do *not* utilize the `id` of the request body.
2. The *only* requirement for creating a [resource according to the jsonapi spec](https://jsonapi.org/format/#crud-creating) is a *type* field.
3. The only time `id`s are being used is when the [client wants to generate a unique identifier](https://jsonapi.org/format/#crud-creating-client-ids) for that resource.

**Recommendation**

Since we are *not* using the `id` field internally as part of creating resources, and they are *not* required by JSON:API, I'd recommend we remove them all. I can create a ticket for this. And seeing how this followup ticket would be a quick effort, are you strongly in favor of adding it here for the sake of consistency even though it will be removed shortly afterwords? I'm not sure it adds any consistency-value.

chrisarcand: Ahh nice work digging in to this. 

> they are not required by JSON:API

They are, aren't they? Don't you still need some sort of public field to add the fieldtag that makes the type be sent properly? Or are you saying you see a way where we can do that, just _not_ as a part of `CreateOptions`/`UpdateOptions`? 

> are you strongly in favor of adding it here

Nope, whatever makes sense to you! Doesn't matter.

omarismail: > They are, aren't they? Don't you still need some sort of public field to add the fieldtag

Yes, that is correct. The requirement is that there is some struct attribute that is the place where the `json:primary,<resource>` resides. 

I'm going to make it `Type` and go with that. That will also answer the issue below about validating type. 

Also going to make a ticket in Asana to come back and fix this id/type stuff for all the other APIs.

admin_terraform_version.go
omarismail: Per the [conversation above](https://github.com/hashicorp/go-tfe/pull/186#discussion_r589703113), I think its better to keep these here, but not use a pointer, that way it is present as a default null value and that way its included in the `jsonapi` as part of marshalling.

admin_terraform_version.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+
+	return tfv, nil
+}
+
+// AdminTerraformVersionUpdateOptions for updating terraform version.
+// https://www.terraform.io/docs/cloud/api/admin/terraform-versions.html#request-body
+type AdminTerraformVersionUpdateOptions struct {
+	Type     string  `jsonapi:"primary,terraform-versions"`
```
chrisarcand, line 133: I really like what you landed on here, with `Type` used in `*Options` but leaving the necessary tag as `ID` in the struct (since jsonapi does say it [should be used on the primary key](https://github.com/google/jsonapi#primary)) Nice work!

One thing I'd consider adding in your followup is a docstring indicating what this is for in each `*Options` (something far more helpful than "Internal use only!"); maybe even just `See /link/to/readme` 🤷

chrisarcand (approved):

admin_terraform_version.go
omarismail: 😄 .

> adding in your followup is a docstring indicating what this is for

👍 will do

PR #190, opened 2021-03-10
Change create/update options in resources to match JSON:API Type field

omarismail:
## Description

Currently a lot of the Create/UpdateOptions structs for various resources have an `ID` attribute that represents the primary field in JSON:API. This is misleading, as [JSON:API indicates](https://jsonapi.org/format/#crud-creating) that this field is a `Type` field. 

This PR is to make the fields more accurately represent the JSON:API spec. It also removes the frequent `options.ID = ""` as that is not necessary.

## Testing plan

No changes. Run all the tests locally.

## Output from tests

```
$ 
...
```

## External links

- [JSON:API](https://jsonapi.org/format/#crud-creating)

### DIFF
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -78,8 +78,9 @@ func (s *agentPools) List(ctx context.Context, organization string, options Agen
 
 // AgentPoolCreateOptions represents the options for creating an agent pool.
 type AgentPoolCreateOptions struct {
-	// For internal use only!
-	ID string `jsonapi:"primary,agent-pools"`
+	// Type is the required field as part of JSON:API.
+	// https://jsonapi.org/format/#crud-creating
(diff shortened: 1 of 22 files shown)


chrisarcand (commented):
Love this, thanks for digging in to it. 

I think even this message we might want to make very clear to users unfamiliar that this field is literally not a settable field or their concern whatsoever, it just a public field that needs to be present.

```go
// Type is a public field utilized by JSON:API to set the resource type via the field tag.
// It is not a user-defined value and does not need to be set.
```

Maybe?

omarismail:
> I think even this message we might want to make very clear to users unfamiliar that this field is literally not a settable field or their concern whatsoever, it just a public field that needs to be present.

Makes sense, I like this wording. Updated.

PR #192, opened 2021-03-11
Add new workspace fields

mwhooker:
## Description

Adds support for new fields in the workspace object

## Example Output

```
(*tfe.Workspace)(0xc00016a000)({
...,
 UpdatedAt: (time.Time) 2020-12-16 18:05:40.569 +0000 UTC,
 ResourceCount: (int) 2,
 ApplyDurationAverage: (time.Duration) 41s,
 PlanDurationAverage: (time.Duration) 25s,
 PolicyCheckFailures: (int) 0,
 RunFailures: (int) 1,
 RunsCount: (int) 6,
})

// spew.Dump of workspaces.Readme:

([]uint8) (len=11 cap=1536) {
 00000000  23 20 74 66 74 65 73 74  0a 0a 31                 |# tftest..1|
}
```

## Testing plan

Added a unit test, but it won't work because this feature is not generally available yet. It should pass once it is.

The fields have been tested manually against an existing workspace with real data.

### DIFF
--- a/workspace.go
+++ b/workspace.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/url"
+	"strings"
 	"time"
 )
 
@@ -24,6 +26,9 @@ type Workspaces interface {
 
 	// Read a workspace by its name.
 	Read(ctx context.Context, organization string, workspace string) (*Workspace, error)
+
+	// Readme gets the readme of a workspace by its ID.
+	Readme(ctx context.Context, workspaceID string) (io.Reader, error)
 
 	// ReadByID reads a workspace by its ID.
 	ReadByID(ctx context.Context, workspaceID string) (*Workspace, error)
@@ -97,12 +102,31 @@ type Workspace struct {
 	TriggerPrefixes      []string              `jsonapi:"attr,trigger-prefixes"`
 	VCSRepo              *VCSRepo              `jsonapi:"attr,vcs-repo"`
 	WorkingDirectory     string                `jsonapi:"attr,working-directory"`
+	UpdatedAt            time.Time             `jsonapi:"attr,updated-at,iso8601"`
+	ResourceCount        int                   `jsonapi:"attr,resource-count"`
+	ApplyDurationAverage time.Duration         `jsonapi:"attr,apply-duration-average"`
+	PlanDurationAverage  time.Duration         `jsonapi:"attr,plan-duration-average"`
+	PolicyCheckFailures  int                   `jsonapi:"attr,policy-check-failures"`
+	RunFailures          int                   `jsonapi:"attr,run-failures"`
+	RunsCount            int                   `jsonapi:"attr,workspace-kpis-runs-count"`
 
 	// Relations
 	AgentPool    *AgentPool    `jsonapi:"relation,agent-pool"`
 	CurrentRun   *Run          `jsonapi:"relation,current-run"`
 	Organization *Organization `jsonapi:"relation,organization"`
 	SSHKey       *SSHKey       `jsonapi:"relation,ssh-key"`
+}
+
+// workspaceWithReadme is the same as a workspace but it has a readme.
+type workspaceWithReadme struct {
+	ID     string           `jsonapi:"primary,workspaces"`
+	Readme *workspaceReadme `jsonapi:"relation,readme"`
+}
+
+// workspaceReadme contains the readme of the workspace.
+type workspaceReadme struct {
+	ID          string `jsonapi:"primary,workspace-readme"`
+	RawMarkdown string `jsonapi:"attr,raw-markdown"`
 }
 
 // VCSRepo contains the configuration of a VCS integration.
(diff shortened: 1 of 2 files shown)


anamahalo (approved):
Pre-emptively ✅ this for GA.

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/hashicorp/go-tfe
 
 require (
-	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/davecgh/go-spew v1.1.1
```
mwhooker, line 4: this is just here for testing. Will pop it out when ready for review

workspace.go
```diff
@@ -314,12 +340,30 @@ func (s *workspaces) Read(ctx context.Context, organization, workspace string) (
...
+		},
+	}
 	err = s.client.do(ctx, req, w)
 	if err != nil {
 		return nil, err
 	}
 
+	// durations come over in ms
+	w.ApplyDurationAverage *= (time.Second / time.Millisecond)
+	w.PlanDurationAverage *= (time.Second / time.Millisecond)
```
chrisarcand, line 365: I don't think I understand this, can you elaborate a bit?

chrisarcand (changes requested):
Heyo! I know this isn't quite ready for review yet, but thought I'd drop some notes because I've done a bit more thinking on this given the current API regarding the README and what I think the interface should be here in the best interests of the client.

I believe the `Readme()` function (which should be in the `Workspaces` interface, by the way) should make a separate call to fetch the README, and should return an `io.Reader`. This is consistent with the API elsewhere, like [fetching plan/apply logs](https://github.com/hashicorp/go-tfe/blob/84e99f88c98cf8da3c9b988ccee3d86ca1a44436/apply.go#L23) or a run's JSON [formatted execution plan](https://www.terraform.io/docs/cloud/api/plans.html#retrieve-the-json-execution-plan).

Note that the current API being non-streamable isn't the point here - it's about the fact that this interface _could_ be streamed with a separate endpoint in the future and no backwards incompatible change in type here in the client. In other words, although not useful at this immediate moment, returning an `io.Reader` here doesn't hurt anything and just future proofs it for an API more appropriate for the client. Today, the API call to fulfill this interface is indeed just a nearly-duplicated `/workspaces/:id?include=readme` call, but it's only out of current circumstance 👍🏻

workspace.go
mwhooker: This was wrong. fixed and I hope it makes more sense now

mwhooker:
I think I got to your feedback @chrisarcand . LMK

mwhooker:
> which should be in the Workspaces interface, by the way

I missed this the first time. Can you elaborate?

chrisarcand:
Sure thing:

This client is organized as a set of interfaces tied to API calls. The plural name `Workspaces` doesn't literally refer to just collection calls (`/workspaces`); it's just the name of the interface in which all Workspace API endpoints are organized. No methods are defined on the struct representing the resource itself, but on the interface to the specific API.

As a simpler example to grok than workspaces, check out [plan.go](https://github.com/hashicorp/go-tfe/blob/1de9e44d1f0eb5f9063d9053580f04e0e1c2be09/plan.go):

https://github.com/hashicorp/go-tfe/blob/1de9e44d1f0eb5f9063d9053580f04e0e1c2be09/plan.go#L19-L25

https://github.com/hashicorp/go-tfe/blob/1de9e44d1f0eb5f9063d9053580f04e0e1c2be09/plan.go#L48-L61

Note that to retrieve the logs for a plan, is not a method on `Plan` but a function in the `Plans` interface. 

In your case, this means you should:

* Take all of the logic of making a call for the readme, unmarshalling it, and returning it in a function `Readme()` included in the `Workspaces` interface and method on `workspaces`.
* The `readme` field in `Workspace` should be removed, along with the `readmeReader` type (you don't really need it)
* The logic for immediately fetching the readme in `Read()` and `ReadByID()` should be removed.

omarismail:
@mwhooker Can we see a sample of what the output json looks like?

workspace.go
```diff
@@ -322,6 +346,10 @@ func (s *workspaces) Read(ctx context.Context, organization, workspace string) (
 		return nil, err
 	}
 
+	// durations come over in ms
+	w.ApplyDurationAverage *= time.Millisecond
+	w.PlanDurationAverage *= time.Millisecond
```
omarismail, line 351: Is the goal to convert this so that it is seconds, even though the data comes over in ms?

omarismail (approved):
lgtm

workspace.go
chrisarcand: Yeah, this got lost in commenting above; I think we should probably leave the duration alone and use what the API provides directly (ms), unless there's some specific reason we should be returning something different from the canonical source (I don't see any?). Any caller can do whatever they want with it. Pity the attribute(s) isn't named `apply-duration-average-ms` for clarity.

chrisarcand:
This requires a bit of fiddling with our CI to be green, and I'll get to that today 👍🏻

workspace.go
mwhooker: The underlying type is a `time.Duration` which expects nanoseconds since epoch, but we send over ms. In order to fit the datatype we need to "tell" the `time.Duration` that the values are milliseconds and not nanoseconds

mwhooker: > Pity the attribute(s) isn't named apply-duration-average-ms for clarity.

agreed

chrisarcand: Ahhhhhh got it, whoops, missed that. 🙇🏻‍♂️

chrisarcand (approved):

PR #193, opened 2021-03-11
Add missing defer cleanups

omarismail:
## Description

There are a number of resources that are left without being cleaned up. This is because some create calls don't include the cleanup function. This isn't added to every test, as some tests explicitly test Deletion.

This PR goes through and adds the missing `defer cleanup()` for various resources.

## Testing plan

1. Run tests.

## Output from tests

```
$ go test -v  ./...  -timeout=30m 
PASS
ok  	github.com/hashicorp/go-tfe	562.416s

...
```


cam-stitt (approved):
Nice job!

PR #195, opened 2021-03-17
Refactored unmarshal into its own function

omarismail:
## Description

This PR extracts some unmarshalling logic into its own function. The reason for this change is to make it easier to test unmarshalling response. It is done in preparation for a further change of using an up-to-date JSONAPI library. 

## Testing plan

Run tests.

## Output from tests

```
$ go test -v  ./...  -failfast -timeout=30m

```

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -664,8 +664,6 @@ func unmarshalResponse(responseBody io.Reader, model interface{}) error {
 	pagination.Set(reflect.ValueOf(p))
 
 	return nil
-
-	return nil
 }
 
 // ListOptions is used to specify pagination options when making API requests.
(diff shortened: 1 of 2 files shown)


tfe.go
```diff
@@ -660,6 +664,8 @@ func (c *Client) do(ctx context.Context, req *retryablehttp.Request, v interface
 	pagination.Set(reflect.ValueOf(p))
 
 	return nil
+
+	return nil
```
lafentres, line 668: looks like an extra `return nil` snuck in here 😁

lafentres (approved):
aside from that extra return, this looks good and makes sense to me 👍🏼

tfe.go
omarismail: Oops 🤦‍♂️

PR #196, opened 2021-03-17
lint fixes

mwhooker:
## Description

Fix some lint issues
## Testing plan

1. unit tests

### DIFF
--- a/notification_configuration.go
+++ b/notification_configuration.go
@@ -96,7 +96,7 @@ type DeliveryResponse struct {
 	Body       string      `json:"body"`
 	Code       int         `json:"code"`
 	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at,iso8601"`
+	SentAt     time.Time   `json:"sent-at"`
 	Successful bool        `json:"successful"`
 	URL        string      `json:"url"`
 }
--- a/notification_configuration.go
+++ b/notification_configuration.go
@@ -96,7 +96,7 @@ type DeliveryResponse struct {
 	Body       string      `json:"body"`
 	Code       int         `json:"code"`
 	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at"`
+	SentAt     time.Time   `json:"sent-at,iso8601"`
 	Successful bool        `json:"successful"`
 	URL        string      `json:"url"`
 }
--- a/admin_run.go
+++ b/admin_run.go
@@ -65,7 +65,7 @@ func (s *adminRuns) List(ctx context.Context, options AdminRunsListOptions) (*Ad
 		return nil, err
 	}
 
-	u := fmt.Sprintf("admin/runs")
+	u := "admin/runs"
 	req, err := s.client.newRequest("GET", u, &options)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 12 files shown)


notification_configuration.go
```diff
@@ -93,12 +93,12 @@ type NotificationConfiguration struct {
 
 // DeliveryResponse represents a notification configuration delivery response.
 type DeliveryResponse struct {
-	Body       string      `json:"body"`
-	Code       int         `json:"code"`
-	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at,iso8601"`
-	Successful bool        `json:"successful"`
-	URL        string      `json:"url"`
+	Body       string      `jsonapi:"attr,body"`
```
omarismail, line 96: This isn't going to work. I'm actually in the process of fixing this. See this slack thread: https://hashicorp.slack.com/archives/C01BR9FR8KZ/p1615941703005300

omarismail (approved):
Once you remove the `jsonapi` change [here](https://github.com/hashicorp/go-tfe/pull/196/files#diff-a80eaffcbd24a1460b26eec1f92686ea8b2e9452cdb6c74bfaa7157e3af200a1R96), its LGTM

✅ 🚢

notification_configuration.go
mwhooker: gotcha. I will remove the `iso8601` annotation though since the json parser doesn't understand it. We'll just have to remember to put it back if jsonapi is ever supported.

notification_configuration.go
```diff
@@ -96,7 +96,7 @@ type DeliveryResponse struct {
 	Body       string      `json:"body"`
 	Code       int         `json:"code"`
 	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at,iso8601"`
+	SentAt     time.Time   `json:"sent-at"`
```
chrisarcand, line 99: ? I don't understand this one; is this a no-op?

mwhooker: it's a no-op, just a lint fix:

```
notification_configuration.go:99:25: SA5008: unknown JSON option "iso8601" (staticcheck)
	SentAt     time.Time   `json:"sent-at,iso8601"`
	                       ^
```

mwhooker: fine dropping the commit, but it's just nonsense the json parser doesn't understand

omarismail: I think this should be fine. I'm not seeing any custom [changes in the forked jsonapi](https://github.com/svanharmelen/jsonapi/blame/master/request.go#L505-L520).

@mwhooker I [just merged](https://github.com/hashicorp/go-tfe/pull/195) a change to extract the unmarshalling bit into a separate function so it makes it easier to validate these parsing changes.

Here is an example test that it uses to [build a response body](https://github.com/hashicorp/go-tfe/blob/master/tfe_test.go#L408-L431) and it calls [unmarshalResponse](https://github.com/hashicorp/go-tfe/blob/master/tfe_test.go#L437).

Is it possible to do that here? Feel free to message me on Slack

mwhooker: Hey @omarismail, Will see if I can take a look at this another day. I probably don't have too much more time to work on this PR. I'll just revert the commit if there's follow-on work you think should be done.

PR #197, opened 2021-03-19
Use up-to-date jsonapi package

omarismail:
## Description

This library has relied on a forked and outdated (~50 commits behind upstream) version of `jsonapi`. This PR uses an up-to-date version of jsonapi, forked on [hashicorp/jsonapi](https://github.com/hashicorp/jsonapi/), and changes all the struct tags to use `jsonapi`.

The changes are primarily to ensure stability of the package. 

There are a few changes that will require users of this package to alter a few things. We will be doing another release once this is merged in, and because these are just type changes, users of this newer version will be able to quickly address the type changes.

1. The `DeliveryResponse` in `NotificationConfiguration` [never worked](https://github.com/hashicorp/go-tfe/pull/197#discussion_r600576577). The `Code` and `Successful` never unmarshalled correctly and had a 0 value. So the types of these two were changed to reflect how they are received.
2. `Policy.Enforce` is no longer a `[]*Enforcement` but rather `[]Enforcement`. The upstream jsonapi package does not handle a slice of pointers. 

## Testing

Run the test suite and everything should pass. There was no behavior change, just type changes.


apply.go
```diff
@@ -57,12 +56,12 @@ type Apply struct {
...
-	FinishedAt      time.Time `json:"finished-at"`
-	ForceCanceledAt time.Time `json:"force-canceled-at"`
-	QueuedAt        time.Time `json:"queued-at"`
-	StartedAt       time.Time `json:"started-at"`
+	CanceledAt      string `jsonapi:"attr,canceled-at"`
+	ErroredAt       string `jsonapi:"attr,errored-at"`
+	FinishedAt      string `jsonapi:"attr,finished-at"`
+	ForceCanceledAt string `jsonapi:"attr,force-canceled-at"`
+	QueuedAt        string `jsonapi:"attr,queued-at"`
+	StartedAt       string `jsonapi:"attr,started-at"`
```
omarismail, line 64: This is waiting on a fix to be merged in the jsonapi package so that it can read other time formats.

configuration_version.go
```diff
@@ -87,9 +86,9 @@ type ConfigurationVersion struct {
 // CVStatusTimestamps holds the timestamps for individual configuration version
 // statuses.
 type CVStatusTimestamps struct {
-	FinishedAt time.Time `json:"finished-at"`
-	QueuedAt   time.Time `json:"queued-at"`
-	StartedAt  time.Time `json:"started-at"`
+	FinishedAt string `jsonapi:"attr,finished-at"`
+	QueuedAt   string `jsonapi:"attr,queued-at"`
+	StartedAt  string `jsonapi:"attr,started-at"`
```
omarismail, line 91: This is waiting on a fix to be merged in the jsonapi package so that it can read other time formats.

cost_estimate.go
```diff
@@ -58,12 +58,12 @@ type CostEstimate struct {
...
-	FinishedAt              time.Time `json:"finished-at"`
-	PendingAt               time.Time `json:"pending-at"`
-	QueuedAt                time.Time `json:"queued-at"`
-	SkippedDueToTargetingAt time.Time `json:"skipped-due-to-targeting-at"`
+	CanceledAt              string `jsonapi:"attr,canceled-at"`
+	ErroredAt               string `jsonapi:"attr,errored-at"`
+	FinishedAt              string `jsonapi:"attr,finished-at"`
+	PendingAt               string `jsonapi:"attr,pending-at"`
+	QueuedAt                string `jsonapi:"attr,queued-at"`
+	SkippedDueToTargetingAt string `jsonapi:"attr,skipped-due-to-targeting-at"`
```
omarismail, line 66: This is waiting on a fix to be merged in the jsonapi package so that it can read other time formats.

notification_configuration.go
```diff
@@ -93,12 +93,19 @@ type NotificationConfiguration struct {
...
-	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at,iso8601"`
-	Successful bool        `json:"successful"`
-	URL        string      `json:"url"`
+	Body string `jsonapi:"attr,body"`
+	Code string `jsonapi:"attr,code"`
+	//Headers    http.Header `jsonapi:"attr,headers"`
+	SentAt     string `jsonapi:"attr,sent-at"`
+	Successful string `jsonapi:"attr,successful"`
+	URL        string `jsonapi:"attr,url"`
```
omarismail, line 101: This is waiting on a fix in the API to send this data in a proper format. Specifically relating to `SentAt` and Code`. `SentAt` is waiting on a time format update in `hashicorp/jsonapi`, and `Code` is supposed to be an integer, which will be updated in the API.

apply.go
omarismail: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

configuration_version.go
omarismail: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

cost_estimate.go
omarismail: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

plan.go
```diff
@@ -62,12 +61,12 @@ type Plan struct {
...
-	FinishedAt      time.Time `json:"finished-at"`
-	ForceCanceledAt time.Time `json:"force-canceled-at"`
-	QueuedAt        time.Time `json:"queued-at"`
-	StartedAt       time.Time `json:"started-at"`
+	CanceledAt      string `jsonapi:"attr,canceled-at"`
+	ErroredAt       string `jsonapi:"attr,errored-at"`
+	FinishedAt      string `jsonapi:"attr,finished-at"`
+	ForceCanceledAt string `jsonapi:"attr,force-canceled-at"`
+	QueuedAt        string `jsonapi:"attr,queued-at"`
+	StartedAt       string `jsonapi:"attr,started-at"`
```
omarismail, line 69: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

plan_export.go
```diff
@@ -58,11 +57,11 @@ const (
...
-	CanceledAt time.Time `json:"canceled-at"`
-	ErroredAt  time.Time `json:"errored-at"`
-	ExpiredAt  time.Time `json:"expired-at"`
-	FinishedAt time.Time `json:"finished-at"`
-	QueuedAt   time.Time `json:"queued-at"`
+	CanceledAt string `jsonapi:"attr,canceled-at"`
+	ErroredAt  string `jsonapi:"attr,errored-at"`
+	ExpiredAt  string `jsonapi:"attr,expired-at"`
+	FinishedAt string `jsonapi:"attr,finished-at"`
+	QueuedAt   string `jsonapi:"attr,queued-at"`
```
omarismail, line 64: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

policy_check.go
```diff
@@ -82,34 +82,33 @@ type PolicyCheck struct {
...
-	ErroredAt    time.Time `json:"errored-at"`
-	HardFailedAt time.Time `json:"hard-failed-at"`
-	PassedAt     time.Time `json:"passed-at"`
-	QueuedAt     time.Time `json:"queued-at"`
-	SoftFailedAt time.Time `json:"soft-failed-at"`
+	ErroredAt    string `jsonapi:"attr,errored-at"`
+	HardFailedAt string `jsonapi:"attr,hard-failed-at"`
+	PassedAt     string `jsonapi:"attr,passed-at"`
+	QueuedAt     string `jsonapi:"attr,queued-at"`
+	SoftFailedAt string `jsonapi:"attr,soft-failed-at"`
```
omarismail, line 111: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

run.go
```diff
@@ -115,33 +115,33 @@ type Run struct {
...
+	ErroredAt            string `jsonapi:"attr,errored-at"`
+	FinishedAt           string `jsonapi:"attr,finished-at"`
+	QueuedAt             string `jsonapi:"attr,queued-at"`
+	StartedAt            string `jsonapi:"attr,started-at"`
+	ApplyingAt           string `jsonapi:"attr,applying-at"`
+	AppliedAt            string `jsonapi:"attr,applied-at"`
+	PlanningAt           string `jsonapi:"attr,planning-at"`
+	PlannedAt            string `jsonapi:"attr,planned-at"`
+	PlannedAndFinishedAt string `jsonapi:"attr,planned-and-finished-at"`
+	PlanQueuabledAt      string `jsonapi:"attr,plan-queueable-at"`
```
omarismail, line 144: This is fixed [here](https://github.com/hashicorp/jsonapi/pull/1), will update this comment soon.

notification_configuration.go
```diff
@@ -93,12 +92,27 @@ type NotificationConfiguration struct {
 
 // DeliveryResponse represents a notification configuration delivery response.
 type DeliveryResponse struct {
-	Body       string      `json:"body"`
-	Code       int         `json:"code"`
```
omarismail, line 97: This never worked as intended. The result would be a `0`. See https://play.golang.org/p/u6bGq49tHa7

notification_configuration.go
```diff
@@ -93,12 +92,27 @@ type NotificationConfiguration struct {
...
-	Headers    http.Header `json:"headers"`
-	SentAt     time.Time   `json:"sent-at,iso8601"`
-	Successful bool        `json:"successful"`
```
omarismail, line 100: This never worked as intended. It would default to `false`. See https://play.golang.org/p/u6bGq49tHa7

notification_configuration.go
```diff
@@ -93,12 +92,27 @@ type NotificationConfiguration struct {
...
-	URL        string      `json:"url"`
+	Body       string    `jsonapi:"attr,body"`
+	Code       string    `jsonapi:"attr,code"`
+	Headers    headers   `jsonapi:"attr,headers"`
```
omarismail, line 97: Changed to use the `headers` struct below instead of `http.Header` because jsonapi does not, currently, handle `map[string][]string`. Also, the headers listed in the struct below match what is listed in the documentation. This is more orderly than an arbitrary list of headers. If a customer wants to add another header, that can very easily be accommodated.

chrisarcand: (I also wondered how the tests pass with this, given that it is a bigger behavior change; [turns out we don't actually test beyond equality](https://github.com/hashicorp/go-tfe/blob/074a0e44b072d8f9174639b0139cf1e9fb2f8e2e/notification_configuration_test.go#L167-L168) 😕)

omarismail: I don't know how this actually ever worked. Even in the outdated fork, along with upstream, if the tag for the struct isn't `jsonapi`, [it gets skipped](https://github.com/svanharmelen/jsonapi/blob/master/request.go#L156-L159).

omarismail: Once [this PR](https://github.com/hashicorp/jsonapi/pull/2) is merged, i'll update this and we'd be good to go!

omarismail: This if fixed!!

chrisarcand (approved):
Nice work!

PR #199, opened 2021-03-25
Audit trail client

chrusty:
## Description

The SDK currently doesn't support the audit trails API for Terraform Cloud (I'm working on a tool which will extract these so they can be stored offline).

On the face of it I can understand why this might not be here (yet):
- The audit trails API doesn't adhere to JSONAPI as the rest of the calls seem to. This means I can't rely on the default response marshaling, but fortunately someone was kind enough to allow the use of an io.Writer so I can bypass this and do it manually.
- The audit trails API requires an org token instead of a regular API token

## Testing plan

- Run a simple test with an org token to see if we can access audit trails
- Make sure that the list is non-zero in length (because even with an api token you _do_ get a response - just no results)

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/audit-trails.html)

## Output from tests

I'm about to add the tests, just wanted to see if this feature is something that you would merge, or would rather not have in this SDK due to the reasons I've highlighted.


chrusty:
I've signed the License

chrisarcand:
Hello! Sorry for the delay, and thanks for your patience.

This is indeed something that we'd accept in this general API client! Thanks for your work on it. There are a few things that I'd like to tweak first, namely that the 'audit trail' is actually a list of audit _events_, so I find the name a bit confusing. I do know that this is exactly what the API documentation uses, though - so I'll go sort through that and see if we can't get that a bit more clear before we go making an API contract in what you're naming the functions and structs here.

chrusty:
> Hello! Sorry for the delay, and thanks for your patience.
> 
> This is indeed something that we'd accept in this general API client! Thanks for your work on it. There are a few things that I'd like to tweak first, namely that the 'audit trail' is actually a list of audit _events_, so I find the name a bit confusing. I do know that this is exactly what the API documentation uses, though - so I'll go sort through that and see if we can't get that a bit more clear before we go making an API contract in what you're naming the functions and structs here.

Hey that's alright, thanks for your response. I've found a bit of an issue since raisi

audit_trail.go
```diff
@@ -0,0 +1,104 @@
...
+// Compile-time proof of interface implementation.
+var _ AuditTrails = (*auditTrails)(nil)
+
+// AuditTrails describes all the audit trail related methods that the
+// Terraform Cloud API supports.
+//
+// TFE API docs: https://www.terraform.io/docs/cloud/api/audit-trails.html
+type AuditTrails interface {
+	// List audit trails visible to the current user.
+	List(ctx context.Context, options AuditTrailListOptions) (*AuditTrailList, error)
```
chrisarcand, line 19:
```suggestion
	// List an organization's audit events. Requires an organization token.
	ListAuditEvents(ctx context.Context, options AuditEventListOptions) (*AuditEventList, error)
```

audit_trail.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// auditTrails implements AuditTrails.
+type auditTrails struct {
+	client *Client
+}
+
+// AuditTrail represents a Terraform Cloud audit trail.
+type AuditTrail struct {
```
chrisarcand, line 28:
```suggestion
type AuditEvent struct {
```

chrisarcand (changes requested):
Ok! The [documentation is clarified a bit](https://github.com/hashicorp/terraform-website/pull/1678), to demonstrate the high level changes I'd make here: Although the API is still called the AuditTrails API (so, `client.AuditTrails` 👌 ) I'd rename anything that refers to the response objects as events, rather than trails themselves (the list function, structs, etc). I've made a few example edits below (not exhaustive).

audit_trail.go
chrusty: Sure thing, I understand. It is best to get the naming right before moving on!

audit_trail.go
chrusty: Hi @chrisarcand, thanks for looking at this. Sorry about the delay on my side this time!

Don't sweat about the renaming. I understand the importance of getting it right! I've changed the name from trail to event throughout the PR (apart from the API endpoint itself, and the link to the API doc).

Since raising this PR I've actually build a tool (using my fork) to store my org's audit events on our own S3 bucket. It works alright, but I noticed that I was only getting 1000 events because pagination wasn't working. I'm guessing it is because this API doesn't use the same JSON-API standard as the rest of them, and the json tags on the Pagination aren't allowing the raw response unmarshal correctly into the pagination attributes.

I ended up butchering the pagination struct a bit to make this work in my fork, but we obviously need a cleaner way to get it done in this repo.

I've included the butchered pagination struct in this PR, but I'm fairly sure it will break the rest of the APIs. I'm trying to think of the least-crufty way to make this work.

tfe.go
```diff
@@ -678,11 +680,11 @@ type ListOptions struct {
...
-	CurrentPage  int `json:"current-page"`
-	PreviousPage int `json:"prev-page"`
-	NextPage     int `json:"next-page"`
-	TotalPages   int `json:"total-pages"`
-	TotalCount   int `json:"total-count"`
+	CurrentPage  int `json:"current_page"`
+	PreviousPage int `json:"prev_page"`
+	NextPage     int `json:"next_page"`
+	TotalPages   int `json:"total_pages"`
+	TotalCount   int `json:"total_count"`
```
chrusty, line 687: This is the JSON tag change. I guess the other unmarshaler is more permissive?

audit_event.go
```diff
@@ -0,0 +1,110 @@
...
+
+// AuditEventList represents a list of audit events.
+type AuditEventList struct {
+	Pagination struct {
+		CurrentPage  int `json:"current_page"`
+		PreviousPage int `json:"prev_page"`
+		NextPage     int `json:"next_page"`
+		TotalPages   int `json:"total_pages"`
+		TotalCount   int `json:"total_count"`
+	} `json:"pagination"`
```
chrusty, line 46: This is the ugly thing that had to be done in order for the pagination data to unmarshal correctly. Note the underscores in the JSON tag names (this reflects the raw response).

PR #201, opened 2021-04-08
Add Workspace Remote State Consumers

omarismail:
## Description

This PR adds the ability to add/remove remote state consumers of a Workspace.

Added a new boolean field on Workspace `GlobalRemoteState`. Added three methods on Workspace
1. `AddRemoteStateConsumers`
2. `RemoveRemoteStateConsumers`
3. `ReadRemoteStateConsumers`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210401204629-95d572587ba2
+	github.com/hashicorp/jsonapi v0.0.0-20210413154710-353a8ffc8171
 	github.com/stretchr/testify v1.3.0
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 )
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210413154710-353a8ffc8171
+	github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1
 	github.com/stretchr/testify v1.3.0
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 )
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1
+	github.com/hashicorp/jsonapi v0.0.0-20210413154710-353a8ffc8171
 	github.com/stretchr/testify v1.3.0
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 )
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module github.com/hashicorp/go-tfe
 
-replace github.com/hashicorp/jsonapi => /Users/omarismail/work/hashi-jsonapi
-
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/google/go-querystring v1.0.0
@@ -9,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1
+	github.com/hashicorp/jsonapi v0.0.0-20210401204629-95d572587ba2
 	github.com/stretchr/testify v1.3.0
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 )
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/hashicorp/go-tfe
 
+replace github.com/hashicorp/jsonapi => /Users/omarismail/work/hashi-jsonapi
+
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/google/go-querystring v1.0.0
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1
+	github.com/hashicorp/jsonapi v0.0.0-20210401204629-95d572587ba2
```
chrisarcand, line 10: Did we want to bump this to latest here, or do it elsewhere?

go.sum
```diff
@@ -14,6 +14,8 @@ github.com/hashicorp/go-uuid v1.0.1 h1:fv1ep09latC32wFoVwnqcnKJGnMSdBanPczbHAYm1
 github.com/hashicorp/go-uuid v1.0.1/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/bN7x4byOro=
 github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1 h1:V95qyPxj4IE39U+fyXdQXUWFCwqXUfTbIl/Gb+lb3zc=
 github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1/go.mod h1:Yog5+CPEM3c99L1CL2CFCYoSzgWm5vTU58idbRUaLik=
```
chrisarcand, line 16: Needs a `go mod tidy`

workspace.go
```diff
@@ -65,6 +67,15 @@ type Workspaces interface {
...
 	UnassignSSHKey(ctx context.Context, workspaceID string) (*Workspace, error)
+
+	// Add remote state consumers to a workspace.
+	AddRemoteStateConsumers(ctx context.Context, workspaceID string, options WorkspaceAddRemoteStateConsumersOptions) error
+
+	// Remove remote state consumers from a workspace.
+	RemoveRemoteStateConsumers(ctx context.Context, workspaceID string, options WorkspaceRemoveRemoteStateConsumersOptions) error
+
+	// ReadRemoteStateConsumers reads the remote state consumers for a workspace.
+	ReadRemoteStateConsumers(ctx context.Context, workspaceID string) (*WorkspaceList, error)
```
chrisarcand, line 78: Looking across the other APIs, I think "fetching the thing" is usually not prefixed with `Read`/`Fetch`/etc; seems like this should be just `RemoteStateConsumers` for consistency.

workspace.go
```diff
@@ -111,10 +123,15 @@ type Workspace struct {
...
 
 	// Relations
-	AgentPool    *AgentPool    `jsonapi:"relation,agent-pool"`
-	CurrentRun   *Run          `jsonapi:"relation,current-run"`
-	Organization *Organization `jsonapi:"relation,organization"`
-	SSHKey       *SSHKey       `jsonapi:"relation,ssh-key"`
+	AgentPool                  *AgentPool                `jsonapi:"relation,agent-pool"`
+	CurrentRun                 *Run                      `jsonapi:"relation,current-run"`
+	Organization               *Organization             `jsonapi:"relation,organization"`
+	RemoteStateConsumerDetails *RemoteStateConsumerLinks `jsonapi:"relation,remote-state-consumers"`
```
chrisarcand, line 129: Hmmm I don't think this needs to be unmarshaled with a response at all; it looks like `jsonapi` provides a `RelationshipLinkable` interface such that I _think_ you can just define a function to build the right URL, like this:

```go
func (w Workspace) JSONAPIRelationshipLinks(relation string) *Links {
	if relation == "remote-state-consumers" {
		return &Links{
			"related": fmt.Sprintf("<the right url>"),
		}
	}
	return nil
}
```

https://github.com/google/jsonapi#links

? Unless I'm misunderstanding

workspace.go
```diff
@@ -65,6 +67,15 @@ type Workspaces interface {
...
+
+	// Add remote state consumers to a workspace.
+	AddRemoteStateConsumers(ctx context.Context, workspaceID string, options WorkspaceAddRemoteStateConsumersOptions) error
+
+	// Remove remote state consumers from a workspace.
+	RemoveRemoteStateConsumers(ctx context.Context, workspaceID string, options WorkspaceRemoveRemoteStateConsumersOptions) error
+
+	// ReadRemoteStateConsumers reads the remote state consumers for a workspace.
+	ReadRemoteStateConsumers(ctx context.Context, workspaceID string) (*WorkspaceList, error)
 }
```
chrisarcand, line 79: You can `PATCH` remote state consumers, too; Might want to add that in, as for users with massive amounts of workspace may find it useful if they're doing their own bespoke API setup (even if our provider just always uses add/remove)

chrisarcand (changes requested):
Nice!

go.mod
omarismail: This was for the links work, but i'll update it to include the recent timestamps addition

workspace.go
omarismail: added

workspace.go
omarismail: Updated to be `RemoteStateConsumers`

workspace.go
omarismail: Implimenting the `Linkable` interface only works for [Marshalling](https://github.com/hashicorp/jsonapi/blob/master/response.go#L468-L478). Which makes sense in that you want to customize what links get sent during a request.

But for unmarshalling, it reads the `links` annotation (whether on a relation or directly on the struct), and [unmarshalls](https://github.com/hashicorp/jsonapi/blob/master/request.go#L311) it [into](https://github.com/hashicorp/jsonapi/blob/master/request.go#L469) an arbitrary [map[string]interface](https://github.com/hashicorp/jsonapi/blob/master/request.go#L541)

chrisarcand: Ahhh I see the disconnect here.

I'm looking at it like the actual raw relationships link doesn't need to be fetchable, unmarshalled data. With the API you've built to interact with those remote state consumers, I wonder if we necessarily need to include every bit of link metadata at all, especially if the client itself ignores it (at marshalling, like you said).

There _are_ cases where we probably will need to actually deserialize the raw link and utilize the awesome work you did in the jsonapi package fork, like [the blob storage upload link for policy set versions](https://www.terraform.io/docs/cloud/api/policy-sets.html#create-a-policy-set-version), where someone may want to take the raw url and feed it to some external pipeline to upload to. 

For here, it's just a reference to domain objects still within the domain of the client, so the link is fairly moot. 

I should add that my biggest hesitation here is stamping `XDetails` as the convention for exposing any relationships with links in the future; something feels like it should be generalized more in to some metadata abstraction. At the very least, I think we should remove this small bit (just because it's not really necessary) and continue discussion to add links in general, maybe add it later?

omarismail: All those points make sense, will remove and if we see a need later on, we can add it back and have more context as to why we'd use it.

chrisarcand (approved):
Looks great! Just needs a few corrections to appease the linter and maybe another `go mod tidy` 👍🏻

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.5.2
 	github.com/hashicorp/go-slug v0.4.1
 	github.com/hashicorp/go-uuid v1.0.1
-	github.com/hashicorp/jsonapi v0.0.0-20210326175023-166ae20eb0d1
+	github.com/hashicorp/jsonapi v0.0.0-20210413154710-353a8ffc8171
```
omarismail, line 10: https://github.com/hashicorp/jsonapi/commit/353a8ffc8171165ee211cb4cbacb2b34e0f0bcaf

chrisarcand (approved):

Issue #202, opened 2021-04-09
Unable to create workspace (wrong VCS settings marshal)


SkYNewZ:
I'm unable to create a workspace with VCS settings: `missing required vcs creation params`
Simple code snippet:

```go
package main

import (
	"context"
	"github.com/hashicorp/go-tfe"
	"log"
)

func main() {
	client, err := tfe.NewClient(&tfe.Config{Token: "[redacted]"})
	if err != nil {
		log.Fatalln(err)
	}

	_, err = client.Workspaces.Create(context.Background(), "[redacted]", tfe.WorkspaceCreateOptions{
		AllowDestroyPlan:     tfe.Bool(true),
		AutoApply:            tfe.Bool(true),
		Description:          tfe.String("testing"),
		ExecutionMode:        tfe.String("remote"),
		MigrationEnvironment: nil,
		Name:                 tfe.String("quentin-testing-go-tfe"),
		TerraformVersion:     tfe.String("0.14.7"),
		VCSRepo:              &tfe.VCSRepoOptions{
			Branch:            tfe.String("master"),
			Identifier:        tfe.String("[redacted]"),
			OAuthTokenID:      tfe.String("[redacted]"),
			IngressSubmodules:  tfe.Bool(false),
		},
	})

	log.Fatalln(err)
}
```

Digging into the code, I think the error comes from [`jsonapi`](https://github.com/google/jsonapi/) which does not marshal nested objects (and here the `VCSRepo` object).

I have reproduced https://github.com/hashicorp/go-tfe/blob/master/tfe.go#L551

```go
package main

import (
	"fmt"
	"github.com/hashicorp/go-tfe"
	"github.com/hashicorp/jsonapi"
	"os"
)

func main() {
	onlyVCSSettings := &tfe.VCSRepoOptions{
		Branch:            tfe.String("master"),
		Identifier:        tfe.String("[redacted]"),
		OAuthTokenID:      tfe.String("[redacted]"),
		IngressSubmodules:  tfe.Bool(false),
	}

	entireObject := &tfe.WorkspaceCreateOptions{
		AllowDestroyPlan:     tfe.Bool(true),
		AutoApply:            tfe.Bool(true),
		Description:          tfe.String("testing"),
		ExecutionMode:        tfe.String("remote"),
		MigrationEnvironment: nil,
		Name:                 tfe.String("quentin-testing-go-tfe"),
		TerraformVersion:     tfe.String("0.14.7"),
[...]

PR #204, opened 2021-04-12
fix marshalling nested structs in requests

radditude:
## Description

The jsonapi library doesn't marshal model attributes that are structs properly if they use the jsonapi attr struct tag, but only when marshalling requests; when unmarshalling the response, `attr` tags work just fine. In other words, [`WorkspacePermissions`](https://github.com/hashicorp/go-tfe/blob/4f141e7aaf7fa5d9ad8353125e2eab5f6ddd1876/workspace.go#L149-L160) are working fine since they're included in the response, but aren't something that's set in the request; [`VCSRepoOptions`](https://github.com/hashicorp/go-tfe/blob/4f141e7aaf7fa5d9ad8353125e2eab5f6ddd1876/workspace.go#L273-L278) are options for the request itself, so they're not working.

## Testing plan

1. Using a nested struct in create or update options marshalls to real JSON.

## External links

#202

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -241,9 +241,9 @@ func (o RegistryModuleCreateWithVCSConnectionOptions) valid() error {
 }
 
 type RegistryModuleVCSRepoOptions struct {
-	Identifier        *string `jsonapi:"attr,identifier"`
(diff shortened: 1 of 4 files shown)


omarismail (approved):
🚢

Issue #206, opened 2021-04-14
[BUG] The workspaceID is not url encoded when listing the variables


jonathanheaden:
The `list` function of [variable.go](https://github.com/hashicorp/go-tfe/blob/v0.13.1/variable.go#L80) does not use `url.QueryEscape(workspaceID)` so is unable to find the workspace.

passing the workspaceID pre-escaped results in `invalid value for workspace ID`

PR #207, opened 2021-04-14
Url Encode the workspace ID when listing variables

jonathanheaden:
Fix for #206
Use the same method for url encoded as the other functions

## Description

This is to fix the List function for variables, which fails to find workspaces unless they have been url encoded

## Testing plan

1.  _Describe how to replicate_
1.  _the conditions_
1.  _under which your code performs its purpose,_
1.  _including example code to run where necessary._

## External links

## Output from tests

```
$ go test

...
```


chrisarcand (approved):
Thanks!

PR #212, opened 2021-04-23
Expose NewRequest() and Do()

omarismail:
## Description

This addition exposes `NewRequest()` and `Do()` so that users of this client may make a request using the underlying configuration of the client, which includes everything from parsing the path, adding headers, serializing the request, etc. This is helpful when working with newer endpoints and avoids the need to reimplement some of the functionality within `newRequest()`, `do()`

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -434,13 +434,17 @@ func (c *Client) configureLimiter(rawLimit string) {
 	c.limiter = rate.NewLimiter(limit, burst)
 }
 
-// newRequest creates an API request. A relative URL path can be provided in
+// NewRequest creates an API request. A relative URL path can be provided in
 // path, in which case it is resolved relative to the apiVersionPath of the
 // Client. Relative URL paths should always be specified without a preceding
 // slash.
 // If v is supplied, the value will be JSONAPI encoded and included as the
 // request body. If the method is GET, the value will be parsed and added as
 // query parameters.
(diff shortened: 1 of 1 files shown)


ryanuber (changes requested):
Hey @omarismail, thanks for the PR! I understand that you want to customize requests sent via this library, but I don't think that exposing the raw request details here is the right approach. go-tfe is a first-class, HashiCorp supported API consumer of the Terraform Cloud API. As such, all supported API interactions should have first-class Go API's within this library. Adding an "escape hatch" like this is a strong indicator that a needed bit of functionality is missing.

Additionally, I think that this opens up a lot of opportunity for misuse or unintended use cases. `NewRequest()` and `Do()` would be a pretty open-ended API that gives the user a ton of flexibility to use the API in unsupported ways.

I am very much in favor of exposing a great Go API that feels natural to use and from the user's point of view, eschews fussing with the actual API semantics. If there are missing resources, actions, options, or anything else, I'd err on the side of implementing those directly such that they can be leveraged in a supported way without requiring deep understanding of the actual HTTP API.

omarismail:
Those are all very valid points @ryanuber. The intention behind this was to use those routes for functionality that is missing but not yet ready, so to speak. But, one could just fork, add the functionality, and then when its ready, open a PR to upstream. Will close this out.

PR #214, opened 2021-04-25
Add skip for tests that require paid features

omarismail:
## Description

If someone tries to run the [tests](https://github.com/hashicorp/go-tfe/blob/master/TESTS.md#2-set-up-environment-variables) against Terraform Cloud (`TFE_ADDRESS=https://app.terraform.io`), some of the tests will fail because they depend on paid-only features. 

This PR adds a flag `SKIP_PAID` that skips the tests that require paid features. 

So now, anyone can get a passing suite by using this new flag:
```
TFE_TOKEN=<token> TFE_ADDRESS=https://app.terraform.io SKIP_PAID=1 go test ./...
```

These are a few (but not all) of the resources that are paid only features:
* [Cost Estimate](https://www.terraform.io/docs/cloud/api/team-members.html)
* [Policy Checks](https://www.terraform.io/docs/cloud/api/policy-checks.html)
* [Team Member](https://www.terraform.io/docs/cloud/api/team-members.html)

## Testing plan

Run the test suite with the new `SKIP_PAID=1` flag.

## Output from tests

```
$  TFE_TOKEN=<redacted> TFE_ADDRESS=https://app.terraform.io SKIP_PAID=1 gotestsum --format testname -- -v -timeout=30m ./...

DONE 581 tests, 110 skipped in 0.501s
```


TESTS.md
```diff
@@ -26,6 +26,7 @@ Tests are run against an actual backend so they require a valid backend address
 1. `GITHUB_POLICY_SET_IDENTIFIER` - GitHub policy set repository identifier in the format `username/repository`. Required for running policy set tests.
 1. `GITHUB_REGISTRY_MODULE_IDENTIFIER` - GitHub registry module repository identifier in the format `username/repository`. Required for running registry module tests.
 1. `ENABLE_TFE` - Some tests are only applicable to Terraform Enterprise or Terraform Cloud. By setting `ENABLE_TFE=1` you will enable enterprise only tests and disable cloud only tests. In CI `ENABLE_TFE` is not set so if you are writing enterprise only features you should manually test with `ENABLE_TFE=1` against a Terraform Enterprise instance.
+1. `SKIP_PAID` - Some tests depend on paid only featuers. By setting `SKIP_PAID=1`, you will skip those tests that are depend on paid only features.
```
radditude, line 29: A couple small typos!

```suggestion
1. `SKIP_PAID` - Some tests depend on paid only features. By setting `SKIP_PAID=1`, you will skip tests that access paid features.
```

team_member_test.go
```diff
@@ -9,6 +9,8 @@ import (
 )
 
 func TestTeamMembersList(t *testing.T) {
+	skipIfFreeOnly(t)
```
radditude, line 12: Hmmm, so technically free orgs _are_ able to add/remove/list team members and create team tokens, it's just that they only have access to one team (the owners team) and can't create new ones - that is, this test fails because it tries to create a new team, not because the team members list API endpoint is actually unavailable to free orgs!

Eventually it might be nice to have tests that exercise the owners team too, but in the meantime, do you think it's worth adding a note about the owners team thing to explain why these tests are `skipIfFreeOnly`?

radditude (approved):
Nice addition! This will be really helpful for contributors.

team_member_test.go
omarismail: Yup, will do!

radditude (approved):

PR #217, opened 2021-05-05
fix flaky run-status-related tests

radditude:
## Description

Currently, we're seeing pretty frequent test flakes with errors about [mismatched policy check output
](https://app.circleci.com/pipelines/github/hashicorp/go-tfe/498/workflows/7d6dc2c8-85d5-4cf1-9894-2b8f5c05d3a4/jobs/641/tests#failed-test-1) or [`transition not allowed`](https://app.circleci.com/pipelines/github/hashicorp/go-tfe/497/workflows/1fed2a81-978e-4912-bcca-1a1792f49c6f/jobs/634/tests#failed-test-0).

It seems likely that both of these errors have the same root cause: we've been overloading the `createPlannedRun` helper so that it looks for runs with planned, cost-estimated, **or** policy-checked statuses, instead of just the planned status.

This leaves us open to race conditions when, say, the `createPlannedRun` helper returns an actual planned run, but the test is really expecting a cost-estimated or policy-checked run, and so it fails. In the case of the `transition not allowed` errors, the error comes from attempting to discard a run that's isn't discardable (yet).

This change adds a `createRunWithStatus` helper to make it easy to be very specific about which statuses to accept when creating test runs.

## Testing plan

1. The tests pass, and do so consistently.

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -503,7 +503,7 @@ func createRun(t *testing.T, client *Client, w *Workspace) (*Run, func()) {
 	}
 }
 
-func createRunWithStatus(t *testing.T, client *Client, w *Workspace, timeout int, statuses ...RunStatus) (*Run, func()) {
+func createRunWithStatus(t *testing.T, client *Client, w *Workspace, timeout int, desiredStatuses ...RunStatus) (*Run, func()) {
 	r, rCleanup := createRun(t, client, w)
 
 	var err error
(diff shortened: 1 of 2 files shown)


omarismail (approved):
yeeeesssssssssss 🚢 :shipit:  ✅

helper_test.go
```diff
@@ -514,73 +514,45 @@ func createPlannedRun(t *testing.T, client *Client, w *Workspace) (*Run, func())
 			t.Fatal(err)
 		}
 
-		switch r.Status {
-		case RunPlanned, RunCostEstimated, RunPolicyChecked, RunPolicyOverride:
-			return r, rCleanup
+		for _, status := range statuses {
+			if status == RunApplied && (r.Status == RunPlanned || r.Status == RunCostEstimated) {
```
radditude, line 518: Note to self: add PolicyChecked statuses in here as well.

chrisarcand (approved):
Nice! This is way better, great work!

PR #218, opened 2021-05-12
add refresh and replace attributes for runs

radditude:
## Description

Add support for the following run options:

- [`Run.Refresh`](https://www.terraform.io/docs/cli/commands/plan.html#refresh-false): allow specifying whether or not state should be refreshed prior to a plan
- [`Run.ReplaceAddrs`](https://www.terraform.io/docs/cli/commands/plan.html#replace-address): allow specifying resource addresses to be forcibly replaced during the apply
- `Run.RefreshOnly`: allow specifying whether the run should ignore configuration changes and only refresh the state

### DIFF
--- a/run_test.go
+++ b/run_test.go
@@ -112,17 +112,6 @@ func TestRunsCreate(t *testing.T) {
 		assert.Equal(t, cvTest.ID, r.ConfigurationVersion.ID)
 	})
 
-	t.Run("with refresh disabled", func(t *testing.T) {
-		options := RunCreateOptions{
-			Workspace: wTest,
-			Refresh:   Bool(false),
-		}
-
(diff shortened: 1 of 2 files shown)


run_test.go
```diff
@@ -112,6 +112,27 @@ func TestRunsCreate(t *testing.T) {
 		assert.Equal(t, cvTest.ID, r.ConfigurationVersion.ID)
 	})
 
+	t.Run("with refresh not set", func(t *testing.T) {
```
omarismail, line 115: Excuse the nit, but could we change this to something like `"with refresh not set as a create option, it defaults to true"`

radditude: Good callout for our future selves trying to remember why this got tested separately 😆

chrisarcand (approved):

PR #219, opened 2021-05-17
Fix IPRanges client baseURL munging

chrisarcand:
## Description

The IP Ranges API is not at /api/v2/. This function was changing the client singleton such that any time this function is called, all other calls will break as the baseURL is now incorrect. An absolute path can be provided here instead.

This bug was found downstream, here: https://github.com/hashicorp/terraform-provider-tfe/issues/315


radditude (approved):

nfagerlund (approved):
Also, good to know that go's net/url works the way I would have asked for it to work. 👍🏼

PR #221, opened 2021-05-18
Unmarshal link members; Add Upload() for registry module version slugs

### LINKED ISSUE #159: I am unable to get the upload link when creating a TF Module version
Hello! I am hoping to use this library to create tooling to publish modules to a private registry, but do not see an upload link attribute in [RegistryModuleVersion](https://github.com/hashicorp/go-tfe/blob/master/registry_module.go#L87).

Are there plans to add this attribute? I see a related conversation in PR #150 (`Add policy set version`)

chrisarcand:
## Description

With https://github.com/hashicorp/jsonapi/pull/10, we're now able to do something often-requested for this client: Fetch upload URLs for policy sets and registry modules (which differ from configuration versions in that they provide upload links in [the `links` member of a response](https://jsonapi.org/format/#document-links), not a standard attribute).

Here I've included registry module uploads; policy set versions can be handled separately.

Closes #159 

## Output from tests

```
» go test -v ./... -timeout=30m -run TestRegistryModulesUpload
=== RUN   TestRegistryModulesUpload
=== RUN   TestRegistryModulesUpload/with_valid_upload_URL
=== RUN   TestRegistryModulesUpload/with_missing_upload_URL
--- PASS: TestRegistryModulesUpload (1.84s)
    --- PASS: TestRegistryModulesUpload/with_valid_upload_URL (0.11s)
    --- PASS: TestRegistryModulesUpload/with_missing_upload_URL (0.00s)
PASS
ok      github.com/hashicorp/go-tfe     2.634s
...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -1,10 +1,14 @@
 package tfe
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"net/url"
+	"os"
+
+	slug "github.com/hashicorp/go-slug"
 )
 
 // Compile-time proof of interface implementation.
@@ -35,6 +39,11 @@ type RegistryModules interface {
 
 	// Delete a specific registry module version
 	DeleteVersion(ctx context.Context, organization string, name string, provider string, version string) error
+
+	// Upload Terraform configuration files for the provided registry module version. It
+	// requires a path to the configuration files on disk, which will be packaged by
+	// hashicorp/go-slug before being uploaded.
+	Upload(ctx context.Context, rmv RegistryModuleVersion, path string) error
 }
 
 // registryModules implements RegistryModules.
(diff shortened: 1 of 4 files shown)


registry_module.go
```diff
@@ -94,6 +103,41 @@ type RegistryModuleVersion struct {
...
+	if err != nil {
+		return err
+	}
+	if !file.Mode().IsDir() {
+		return ErrMissingDirectory
+	}
+
+	body := bytes.NewBuffer(nil)
+
+	_, err = slug.Pack(path, body, true)
```
radditude, line 130: I think `slug.Pack` here will respect the settings in a `.terraformignore` file if the user has one, which seems like it could possibly lead to this having confusing behavior. Maybe that's actually the more intuitive behavior, though! Either way, I wanted to call it out in case it's something you hadn't considered.

radditude (approved):
Looks good! I'm curious if we want to leave the default `.terraformignore` behavior of go-slug enabled, though.

registry_module.go
chrisarcand: Ahh good call out! No, I hadn't thought of that at all.

I'm not sure if it should respect the ignore file or not (my gut reaction is 'it should' but honestly I dunno), but regardless in the name of consistency I'll leave it as is - this is identical to [how configuration versions currently work](https://github.com/hashicorp/go-tfe/blob/db513c343c1f3e84c44895fb27a88df11aab490a/configuration_version.go#L196-L199).

radditude: That makes sense!

PR #222, opened 2021-05-18
eliminate some more test flakiness

radditude:
## Description

https://github.com/hashicorp/go-tfe/pull/217 fixed a bunch of race conditions in our tests around planned runs where we were waiting for multiple run statuses and getting inconsistent results. However, it also introduced a new race condition around _applied_ runs where we sometimes get [`transition not allowed` errors](https://app.circleci.com/pipelines/github/hashicorp/go-tfe/515/workflows/46a1da32-1880-46bb-a2dd-44de4f0e8449/jobs/680/tests#failed-test-0) from trying to, say, apply a run that's finished planning, but hasn't finished its cost estimation check yet.

This change makes that applied run check even moar deterministic so we're only ever looking for one status to see if it's ok to apply a run.

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -516,8 +516,8 @@ func createRunWithStatus(t *testing.T, client *Client, w *Workspace, timeout int
 
 		for _, desiredStatus := range desiredStatuses {
 			// if we're creating an applied run, we need to manually confirm the apply once the plan finishes
-			hasApplyableStatus := r.Status == RunPlanned || r.Status == RunCostEstimated || r.Status == RunPolicyChecked
-			if desiredStatus == RunApplied && hasApplyableStatus {
+			isApplyable := hasApplyableStatus(r)
+			if desiredStatus == RunApplied && isApplyable {
 				err := client.Runs.Apply(ctx, r.ID, RunApplyOptions{})
 				if err != nil {
 					t.Fatal(err)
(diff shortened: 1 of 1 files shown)


helper_test.go
```diff
@@ -557,6 +553,22 @@ func createAppliedRun(t *testing.T, client *Client, w *Workspace) (*Run, func())
 	return createRunWithStatus(t, client, w, 90, RunApplied)
 }
 
+func plannedRunStatus() RunStatus {
+	if paidFeaturesDisabled() {
```
chrisarcand, line 557: Since you technically could have paid features without cost estimation enabled, and because you're checking the state of the run specifically in `hasApplyableStatus`, I'd consider changing this to be `if r.CostEstimate != nil` or similar, but I guess that would require some reworking of `createRunWithStatus` as you'll need some sort of argument to say "Planned, whatever 'planned' means in the context of the run". 

It almost seems like the API here should allow for 'statuses' to adverbs not directly tied to specific run states (e.g. "applyable"), necessarily - because the state changes depending on the nature of the workspace/run (what features are enabled, etc).

chrisarcand (approved):
Got some thoughts but ultimately we can revisit this and change it at any time so if this fixes flakiness for now and you don't want to spend any more time on it, sounds good to me!

omarismail (approved):
nice

helper_test.go
radditude: Very good point! The `hasApplyableStatus` helper should certainly be able to make the distinction between paid features enabled in general and cost estimation enabled specifically, since we have access to the run there.

It seems more reasonable for the `createPlannedRun` helper to continue to expect different statuses depending on whether paid features are enabled - if in the future someone wants to test with cost estimation disabled, they can use `createRunWithStatus` by itself.

PR #225, opened 2021-05-26
Organization attribute cleanup

omarismail:
## Description

`EnterprisePlan` is an attribute in Organization that is no longer in the API. This PR removes that.

## External links
- [API documentation](https://www.terraform.io/docs/cloud/api/organizations.html#sample-response-1)


radditude (approved):
Thanks for cleaning that up!

chrisarcand (approved):

PR #227, opened 2021-05-27
Correct RunStatusTimestamps

### LINKED ISSUE #226: RunStatusTimestamps are wrong/out of date
Hi, we are investigating using this API for a Terraform Enterprise installation and noticed some anomalies that we hope can be easily explained/solved.

The TFE API returns a status timestamp object as shown below, and the go-tfe API has different attributes, thus some are not populated (expected).
"status-timestamps": {
                "applied-at": "2021-05-27T17:51:56+00:00",
                "planned-at": "2021-05-27T17:45:06+00:00",
                "applying-at": "2021-05-27T17:50:54+00:00",
                "planning-at": "2021-05-27T17:44:18+00:00",
                "confirmed-at": "2021-05-27T17:50:50+00:00",
                "plan-queued-at": "2021-05-27T17:44:14+00:00",
                "apply-queued-at": "2021-05-27T17:50:51+00:00",
                "plan-queueable-at": "2021-05-27T17:44:12+00:00",
                "policy-checked-at": "2021-05-27T17:45:16+00:00"
            },

go-tfe returns timestamps such as QueuedAt, FinishedAt, ErroredAt. Our TFE version 202104-1 does not have these properties. Is this API coded against a specific version of Terraform Cloud?

Thanks

Don

chrisarcand:
## Description

These timestamp fields are wrong, and were wrong when originally introduced pre-Terraform Cloud. As they've never been populated correctly before, no client could be relying on them and the ones that don't work are removed as a 'not-so-breaking' change, with all the modern timestamps on runs added.

Closes #226

### DIFF
--- a/admin_run_test.go
+++ b/admin_run_test.go
@@ -269,7 +269,7 @@ func TestAdminRun_Unmarshal(t *testing.T) {
 				"has-changes": true,
 				"status":      RunApplied,
 				"status-timestamps": map[string]string{
-					"queued-at": "2020-03-16T23:15:59+00:00",
+					"plan-queued-at": "2020-03-16T23:15:59+00:00",
 				},
 			},
 		},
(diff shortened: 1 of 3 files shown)


thrashr888 (approved):
wow

edit: tests need updated but 👍🏻

PR #232, opened 2021-06-09
Add IngressAttributes of ConfigurationVersion

### LINKED ISSUE #79: Ingress Attribute endpoint missing
It looks like there is an endpoint that is missing from the client, namely the `ingress-attribute`s endpoint. 

I can't seem to find direct documentation to it, although it is mentioned on the [configuration versions](https://www.terraform.io/docs/enterprise/api/configuration-versions.html#available-related-resources) API page in passing. However, it would be super useful to have this as it would allow us to work out when a run is associated to a PR or not

chrisarcand:
## Description

Ingress attributes are resources which include commit information associated with configuration versions sourced from VCS.

They're odd ducks in that they're separate resources but really specific to configuration versions; a long-time vestiage of the Terraform Enterprise API. They also have several ways to retrieve them, both with a standalone ID, a nested URL off of configuration version, and a JSON:API includeable resource.

Here, I've intentionally added the includeable resource bit and no more. Practically speaking, it isn't useful to fetch IA by ID only, and any time you're fetching them via a configuration version you usually want the data within the CV itself as well. So, accessing them as just an extra bit of information on a CV - only - seems most appropriate.

To test this functionality, I ironically needed to add a similar WithOptions function to workspaces, in order to fetch the configuration version of the current run with ingress attributes associated with it.

Closes #79 
Closes #137

### DIFF
--- a/configuration_version_test.go
+++ b/configuration_version_test.go
@@ -151,7 +151,7 @@ func TestConfigurationVersionsReadWithOptions(t *testing.T) {
 	orgTest, orgTestCleanup := createOrganization(t, client)
 	defer orgTestCleanup()
 
-	wTest, wTestCleanup := createWorkspaceWithVCS(t, client, orgTest, true)
+	wTest, wTestCleanup := createWorkspaceWithVCS(t, client, orgTest, WorkspaceCreateOptions{QueueAllRuns: Bool(true)})
 	defer wTestCleanup()
 
 	// Hack: Wait for TFC to ingress the configuration and queue a run
--- a/configuration_version_test.go
+++ b/configuration_version_test.go
@@ -144,6 +144,46 @@ func TestConfigurationVersionsRead(t *testing.T) {
 	})
 }
 
+func TestConfigurationVersionsReadWithOptions(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, wTestCleanup := createWorkspaceWithVCS(t, client, orgTest, true)
+	defer wTestCleanup()
+
+	// Hack: Wait for TFC to ingress the configuration and queue a run
+	time.Sleep(3 * time.Second)
+
+	w, err := client.Workspaces.ReadByIDWithOptions(ctx, wTest.ID, &WorkspaceReadOptions{
+		Include: "current-run.configuration-version",
+	})
+
+	if err != nil {
+		require.NoError(t, err)
+	}
+
+	if w.CurrentRun == nil {
+		t.Fatal("A run was expected to be found on this workspace as a test pre-condition")
+	}
+
+	cv := w.CurrentRun.ConfigurationVersion
+
+	t.Run("when the configuration version exists", func(t *testing.T) {
+		options := &ConfigurationVersionReadOptions{
+			Include: "ingress-attributes",
+		}
+
+		cv, err := client.ConfigurationVersions.ReadWithOptions(ctx, cv.ID, options)
+		require.NoError(t, err)
+
+		assert.NotZero(t, cv.IngressAttributes)
+		assert.NotZero(t, cv.IngressAttributes.CommitURL)
+	})
+}
+
 func TestConfigurationVersionsUpload(t *testing.T) {
 	client := testClient(t)
 	ctx := context.Background()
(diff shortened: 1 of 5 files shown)


configuration_version_test.go
```diff
@@ -144,6 +144,46 @@ func TestConfigurationVersionsRead(t *testing.T) {
...
+
+	if w.CurrentRun == nil {
+		t.Fatal("A run was expected to be found on this workspace as a test pre-condition")
+	}
+
+	cv := w.CurrentRun.ConfigurationVersion
+
+	t.Run("when the configuration version exists", func(t *testing.T) {
+		options := &ConfigurationVersionReadOptions{
+			Include: "ingress-attributes",
```
omarismail, line 176: Could we use a constant here instead of text? Even though one can determine the [relationships via looking at the API Docs](https://www.terraform.io/docs/cloud/api/configuration-versions.html#show-a-configuration-version-39-s-commit-information), I think it'd be helpful for users of this library to be able to see the list of relationships here, and when they use it, to reference a constant instead of strings.

configuration_version_test.go
```diff
@@ -144,6 +144,46 @@ func TestConfigurationVersionsRead(t *testing.T) {
...
+	defer orgTestCleanup()
+
+	wTest, wTestCleanup := createWorkspaceWithVCS(t, client, orgTest, true)
+	defer wTestCleanup()
+
+	// Hack: Wait for TFC to ingress the configuration and queue a run
+	time.Sleep(3 * time.Second)
+
+	w, err := client.Workspaces.ReadByIDWithOptions(ctx, wTest.ID, &WorkspaceReadOptions{
+		Include: "current-run.configuration-version",
```
omarismail, line 161: Same as the [comment below](https://github.com/hashicorp/go-tfe/pull/232/files#r649213396) regarding using a constant.

configuration_version_test.go
omarismail: Another point that is not relevant at all to this PR, but its worth mentioning just for awareness. I think having a typed string for all of the `Include` in options would be very helpful for users of this library. What you can include is only a limited set of strings, not all strings, so being explicit about this is better imo.

helper_test.go
```diff
@@ -916,7 +916,11 @@ func createWorkspace(t *testing.T, client *Client, org *Organization) (*Workspac
 	}
 }
 
-func createWorkspaceWithVCS(t *testing.T, client *Client, org *Organization) (*Workspace, func()) {
+// queueAllRuns: Whether runs should be queued immediately after workspace creation. When set to
+// false, runs triggered by a VCS change will not be queued until at least one run is manually
+// queued. If set to true, a run will be automatically started after the configuration is ingressed
+// from VCS.
```
omarismail, line 922: Could we add a note here saying that 

> if more `WorkspaceCreateOptions` variables are added, consider refactoring this method and passing in `options WorkspaceCreateOptions` directly to this instead of adding more parameters.

omarismail (approved):
A couple comments, otherwise LGTM

configuration_version_test.go
chrisarcand: I generally agree, an enumerated list of types would be nice here, but the fact that you can nest values makes it a little wonky and probably not as ergonomic as we'd hope? What would you do with something like the one below with `current-run.configuration-version`? `CurrentRunType + "." + "ConfigurationVersionType`? Starts to feel more overbearing than helpful, at first glance.

helper_test.go
chrisarcand: Better idea, let's do the thing!

I didn't think of this at all and it's a wonderful idea, I pushed a change to do this instead.

chrisarcand:
@omarismail I'm thinking we merge this as-is without the constants idea; maybe that's a good direction to go or not, but for consistency's sake I'd have the discussion separately and make the change across the board rather than this API being different from the others. Whatcha think?

configuration_version_test.go
omarismail: I can see that. Although, the list of relations for every model is bounded and only has a handful of relations at most. Runs seems like it's on the higher end with 7 relations. `CurrentRunConfigurationVersion` seems good to me?

omarismail:
> @omarismail I'm thinking we merge this as-is without the constants idea; maybe that's a good direction to go or not, but for consistency's sake I'd have the discussion separately and make the change across the board rather than this API being different from the others. Whatcha think?

Agreed in terms of consistency and making the change across the board.

🚢 !

hoshsadiq:
@chrisarcand @omarismail this missing the `commit-sha` attribute.

PR #234, opened 2021-06-15
Add PolicySetVersions relation to PolicySet

omarismail:
## Description

The [API for PolicySets includes two relations](https://www.terraform.io/docs/cloud/api/policy-sets.html#relationships) that we currently do not implement:
```
- newest-version: The most recently created policy set version, regardless of status. Note that this relationship may include an errored and unusable version, and is intended to allow checking for VCS errors.
- current-version: The most recent successful policy set version.
```

This PR updates the PolicySet to include these two relations, and adds a test to assert how/when the relations are set. It also adds a method on PolicySet to `ReadWithOptions` so that one can `Include` other relationships.

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/policy-sets.html#relationships)

## Output from tests
```
go-tfe % TFE_TOKEN=<token> TFE_ADDRESS=https://tfe-zone-4c292d5b.ngrok.io/ go test -v -run TestPolicySetsRead
=== RUN   TestPolicySetsRead
=== RUN   TestPolicySetsRead/with_a_valid_ID
=== RUN   TestPolicySetsRead/without_a_valid_ID
=== RUN   TestPolicySetsRead/with_policy_set_version
--- PASS: TestPolicySetsRead (2.12s)
    --- PASS: TestPolicySetsRead/with_a_valid_ID (0.18s)
    --- PASS: TestPolicySetsRead/without_a_valid_ID (0.00s)
    --- PASS: TestPolicySetsRead/with_policy_set_version (0.80s)
PASS
ok      github.com/hashicorp/go-tfe     2.718s
```

### DIFF
--- a/policy_set.go
+++ b/policy_set.go
@@ -79,7 +79,7 @@ type PolicySet struct {
 	Policies []*Policy `jsonapi:"relation,policies"`
 	// The most recently created policy set version, regardless of status.
 	// Note that this relationship may include an errored and unusable version,
-	// and is intended to allow checking for VCS errors.
+	// and is intended to allow checking for errors.
 	PolicySetVersionNewest *PolicySetVersion `jsonapi:"relation,newest-version"`
 	// The most recent successful policy set version.
 	PolicySetVersionCurrent *PolicySetVersion `jsonapi:"relation,current-version"`
--- a/policy_set.go
+++ b/policy_set.go
@@ -25,6 +25,9 @@ type PolicySets interface {
 	// Read a policy set by its ID.
 	Read(ctx context.Context, policySetID string) (*PolicySet, error)
 
+	// ReadWithOptions reads a policy set by its ID using the options supplied.
+	ReadWithOptions(ctx context.Context, policySetID string, options *PolicySetReadOptions) (*PolicySet, error)
+
 	// Update an existing policy set.
 	Update(ctx context.Context, policySetID string, options PolicySetUpdateOptions) (*PolicySet, error)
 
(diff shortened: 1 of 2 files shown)


policy_set.go
```diff
@@ -71,9 +71,18 @@ type PolicySet struct {
...
+	// The workspaces to which the policy set applies.
+	Workspaces []*Workspace `jsonapi:"relation,workspaces"`
+	// Individually managed policies which are associated with the policy set.
+	Policies []*Policy `jsonapi:"relation,policies"`
+	// The most recently created policy set version, regardless of status.
+	// Note that this relationship may include an errored and unusable version,
+	// and is intended to allow checking for errors.
+	PolicySetVersionNewest *PolicySetVersion `jsonapi:"relation,newest-version"`
+	// The most recent successful policy set version.
+	PolicySetVersionCurrent *PolicySetVersion `jsonapi:"relation,current-version"`
```
chrisarcand, line 85: How about `NewestVersion` and `CurrentVersion`?

The context of it being a version of a policy set is present in the owning type (PolicySet); that's why the attributes themselves are called `newest-version` and `current-version`; this would also match `Workspace.CurrentRun` (rather than `RunCurrent`!)

chrisarcand (approved):
Looks good, but I'd suggest name changes to the fields before merging:

omarismail:
@chrisarcand I've added a new method `ReadWithOptions`, thats why I requested a re-review.

policy_set.go
omarismail: Added

policy_set.go
```diff
@@ -71,9 +74,18 @@ type PolicySet struct {
...
-	Policies     []*Policy     `jsonapi:"relation,policies"`
-	Workspaces   []*Workspace  `jsonapi:"relation,workspaces"`
+	// The workspaces to which the policy set applies.
+	Workspaces []*Workspace `jsonapi:"relation,workspaces"`
+	// Individually managed policies which are associated with the policy set.
+	Policies []*Policy `jsonapi:"relation,policies"`
+	// The most recently created policy set version, regardless of status.
+	// Note that this relationship may include an errored and unusable version,
+	// and is intended to allow checking for errors.
+	PolicySetNewestVersion *PolicySetVersion `jsonapi:"relation,newest-version"`
```
dylanegan, line 86: It looks like this went through as `PolicySetNewestVersion` rather than the suggested `NewestVersion` since the context is within the type already, i.e., PolicySets. It looks odd to type the type.

omarismail: :doh: I read Chris's [comment](https://github.com/hashicorp/go-tfe/pull/234#discussion_r651923824) as to swap the `VersionNewest` with `NewestVersion`. I didn't read it as replace `PolicySetVersionNewest` with _just_ `NewestVersion`.

I'll open up a PR for this.

omarismail: PR: https://github.com/hashicorp/go-tfe/pull/235

PR #235, opened 2021-06-15
Policy Set field name change

omarismail:
## Description

Two new fields were added to PolicySet in a #234. One [suggestion regarding](https://github.com/hashicorp/go-tfe/pull/234#discussion_r652160904) naming was overlooked.

This PR changes the naming to remain consistent with how naming is done in this client.


chrisarcand (approved):
Whoops! Looks good

PR #243, opened 2021-08-11
Add Tags to Workspaces

rexredinger:
## Description

Add support for adding/removing tags from individual workspaces.

## External links

## Output from tests

so far...
```
$ go test

...
=== RUN   TestWorkspacesCreate/with_valid_options
--- PASS: TestWorkspacesCreate/with_valid_options (0.45s)	
...
=== RUN   TestWorkspaces_/successfully_adds_tags
    require.go:794:
        	Error Trace:	workspace_test.go:1059
        	Error:      	Received unexpected error:
        	            	internal server error
        	Test:       	TestWorkspaces_/successfully_adds_tags
...
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -44,6 +44,9 @@ var (
 	// ErrWorkspaceMinLimit is returned when the length of Workspaces is 0.
 	ErrWorkspaceMinLimit = errors.New("must provide at least one workspace")
 
+	// ErrMissingTagIdentifier is returned if a Tag argmunet lacks an Id or Name to match on
+	ErrMissingTagIdentifier = errors.New("Invalid tag relation. Please provide a non-empty tag name or tag ID")
+
 	// Run/Apply errors
 
 	// ErrInvalidRunID is returned when the run ID is invalid.
--- a/errors.go
+++ b/errors.go
@@ -45,7 +45,7 @@ var (
 	ErrWorkspaceMinLimit = errors.New("must provide at least one workspace")
 
 	// ErrMissingTagIdentifier is returned if a Tag argmunet lacks an Id or Name to match on
-	ErrMissingTagIdentifier = errors.New("Invalid tag relation. Please provide a non-empty tag name or tag ID")
+	ErrMissingTagIdentifier = errors.New("bad request\n\nMust specify at least one tag")
 
 	// Run/Apply errors
 
--- a/workspace.go
+++ b/workspace.go
@@ -86,10 +86,10 @@ type Workspaces interface {
 	UpdateRemoteStateConsumers(ctx context.Context, workspaceID string, options WorkspaceUpdateRemoteStateConsumersOptions) error
 
 	// AddTags appends tags to a workspace
-	AddTags(ctx context.Context, workspaceID string, tags []*Tag) error
+	AddTags(ctx context.Context, workspaceID string, options WorkspaceAddTagsOptions) error
 
 	// RemoveTags removes tags from a workspace
-	RemoveTags(ctx context.Context, workspaceID string, tags []*Tag) error
+	RemoveTags(ctx context.Context, workspaceID string, options WorkspaceRemoveTagsOptions) error
 }
 
 // workspaces implements Workspaces.
@@ -1001,23 +1001,11 @@ func (s *workspaces) UpdateRemoteStateConsumers(ctx context.Context, workspaceID
 	return s.client.do(ctx, req, nil)
 }
 
-type workspaceTags struct {
-	Tags []*workspaceTag `json:"data,omitempty"`
-}
-type workspaceTag struct {
-	ID         string            `json:"id,omitempty"`
-	Attributes workspaceTagAttrs `json:"attributes,omitempty"`
-	Type       string            `json:"type,omitempty"`
-}
-type workspaceTagAttrs struct {
-	Name string `json:"name,omitempty"`
-}
-
-type WorkspaceTagsOptions struct {
+type WorkspaceAddTagsOptions struct {
 	Tags []*Tag
 }
 
-func (o WorkspaceTagsOptions) valid() error {
+func (o WorkspaceAddTagsOptions) valid() error {
 	for _, s := range o.Tags {
 		if s.Name == "" && s.ID == "" {
 			return ErrMissingTagIdentifier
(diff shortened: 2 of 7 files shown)


workspace.go
```diff
@@ -309,6 +316,16 @@ type WorkspaceCreateOptions struct {
...
+
+	// A list of tags to attach to the workspace. If the tag does not already
+	// exist, it is created and added to the workspace.
+	Tags []TagOptions `jsonapi:"relation,tags,data,omitempty`
+}
+
+type TagOptions struct {
+	Id   *string `json:"id,omitempty"`
+	Name *string `json:"attributes,name,omitempty"`
+	Type *string `json:"type,omitempty"`
```
thrashr888, line 328: does this need to be dynamic? can we just set it to `tags` so this works?
```go
TagOptions{
				{
					Name: String("tag1"),
				},
				{
					Id:   String("tag-1234abcd568EFGH"),
				},
			},
```

workspace.go
```diff
@@ -969,3 +986,47 @@ func (s *workspaces) UpdateRemoteStateConsumers(ctx context.Context, workspaceID
...
+	req, err := s.client.newRequest("POST", u, &options)
+
+	if err != nil {
+		return err
+	}
+
+	return s.client.do(ctx, req, nil)
+}
+
+func (s *workspaces) DeleteTags(ctx context.Context, workspaceID string, options WorkspaceTagsOptions) error {
```
thrashr888, line 1019: nit: `RemoveTags` might be more appropriate because we're not deleting the tag resource. we're just removing it from this workspace.

workspace_test.go
```diff
@@ -119,6 +119,14 @@ func TestWorkspacesCreate(t *testing.T) {
 			TerraformVersion:           String("0.11.0"),
 			TriggerPrefixes:            []string{"/modules", "/shared"},
 			WorkingDirectory:           String("bar/"),
+			Tags: []*Tag{
+				{
+					Name: "tag1",
+				},
+				{
+					Name: "tag2",
+				},
```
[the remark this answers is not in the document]
omarismail: I definitely do *not* have full context on all the recent Tagging work. Just doing some digging. So in the [OrganizationTags](https://www.terraform.io/docs/cloud/api/organization-tags.html), it says 

> This API returns the list of tags used with all resources across the organization. 

[But we only support](https://github.com/hashicorp/atlas/blob/master/config/routes/v2.rb#L118-L119) `GET` and `DELETE`. And then you can create tags in a [Workspace via the id/name presence](https://github.com/hashicorp/atlas/blob/master/app/controllers/api/v2/workspaces/tags_controller.rb#L32-L38).

I'm sure I'm missing something, but why don't we just add a `POST` route to `organizations/tags#create` to create tags, and then reference the tag created here?

workspace_test.go
```diff
@@ -1066,6 +1100,130 @@ func TestWorkspaces_UpdateRemoteStateConsumers(t *testing.T) {
...
+		require.Error(t, err)
+		assert.EqualError(t, err, ErrMissingTagIdentifier.Error())
+	})
+
+	t.Run("without a valid workspace ID", func(t *testing.T) {
+		err := client.Workspaces.AddTags(ctx, badIdentifier, WorkspaceAddTagsOptions{})
+		require.Error(t, err)
+		assert.EqualError(t, err, ErrInvalidWorkspaceID.Error())
+	})
+}
```
omarismail, line 1164: Can we add one more `t.Run` test here for handling duplicates?

```
	t.Run("successfully handles duplicate tags to be added", func(t *testing.T) {
		err := client.Workspaces.AddTags(ctx, wTest.ID, options)
		require.NoError(t, err)

		w, err := client.Workspaces.Read(ctx, orgTest.Name, wTest.Name)
		require.NoError(t, err)
		assert.Equal(t, 3, len(w.TagNames))
		assert.Equal(t, w.TagNames, []string{"tag1", "tag2", "tag3"})

		err = client.Workspaces.AddTags(ctx, wTest.ID, WorkspaceAddTagsOptions{
			Tags: []*Tag{
				{
					Name: "tag3",
				},
			},
		})
		require.NoError(t, err)

		w, err = client.Workspaces.Read(ctx, orgTest.Name, wTest.Name)
		require.NoError(t, err)
		assert.Equal(t, 3, len(w.TagNames))
		assert.Equal(t, w.TagNames, []string{"tag1", "tag2", "tag3"})
	}
```

workspace_test.go
```diff
@@ -1066,6 +1100,130 @@ func TestWorkspaces_UpdateRemoteStateConsumers(t *testing.T) {
...
+		assert.Equal(t, w.TagNames, []string{"tag3"})
+	})
+
+	t.Run("with invalid options", func(t *testing.T) {
+		err := client.Workspaces.RemoveTags(ctx, wTest.ID, WorkspaceRemoveTagsOptions{
+			Tags: []*Tag{},
+		})
+		require.Error(t, err)
+		assert.EqualError(t, err, ErrMissingTagIdentifier.Error())
+	})
```
omarismail, line 1218: One more test here 🙏 . Could we try removing a tag that doesn't exist?

```
	t.Run("attempts to remove a tag that doesn't exist", func(t *testing.T) {
		err := client.Workspaces.RemoveTags(ctx, wTest.ID, WorkspaceRemoveTagsOptions{
			Tags: []*Tag{
			   {
			      Name: "NonExistentTag",
		            },
			},
		})
		require.Error(t, err)
	})
```

workspace_test.go
thrashr888: wouldn't that be more a function of the API server rather than this client? or is there some mechanism in go-tfe we would test with this?

workspace_test.go
thrashr888: it's a silent failure in the server so i'll test `NoError` instead 👍🏻

workspace_test.go
omarismail: > wouldn't that be more a function of the API server rather than this client
Fair point 👍

omarismail (approved):
LGTM!

errors.go
```diff
@@ -44,6 +44,9 @@ var (
 	// ErrWorkspaceMinLimit is returned when the length of Workspaces is 0.
 	ErrWorkspaceMinLimit = errors.New("must provide at least one workspace")
 
+	// ErrMissingTagIdentifier is returned if a Tag argmunet lacks an Id or Name to match on
+	ErrMissingTagIdentifier = errors.New("bad request\n\nMust specify at least one tag")
```
chrisarcand, line 48: Nit: no other errors have a header like this
```suggestion
	ErrMissingTagIdentifier = errors.New("must specify at least one tag")
```

thrashr888: i thought that was strange too but that's what came back from the API in the tests.

rexredinger:
Big thanks to @thrashr888 for getting this shipped, and @omarismail and @chrisarcand for all the helpful feedback!

errors.go
chrisarcand: You don't need to match that. The format you see is [the client itself interpreting a response from the platform, using the response code as a header.](https://github.com/hashicorp/go-tfe/blob/0743349bef2e7efb1ce3666321c718e6fa9dbd28/tfe.go#L744) This is a client-side validation. 

No biggie, we'll amend it before releasing.

PR #245, opened 2021-08-18
Update state-version outputs for complex types

elsesiy:
Replace string with json.RawMessage to account for non-string state outputs

## Description
TF state outputs can be complex types, but the current implementation always assumes strings.

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/state-version-outputs.html#show-a-state-version-output)

## Output from tests

I have to rely on the CI to execute tests as I don't have access to a live instance right now, sorry.

### DIFF
--- a/state_version_output.go
+++ b/state_version_output.go
@@ -2,6 +2,7 @@ package tfe
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 )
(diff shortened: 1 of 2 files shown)


omarismail:
Hey @elsesiy, I appreciate you taking the time to do this!

The `string` type of `StateVersionOutput.Value` has been a long standing issue, and this [PR now converts](https://github.com/hashicorp/go-tfe/pull/246) it to a `interface{}` type. So you can now type cast it into anything you'd like

https://github.com/hashicorp/go-tfe/pull/246

PR #248, opened 2021-08-22
Workspace ReadWithOptions

omarismail:
## Description

According to the [Workspace API](https://www.terraform.io/docs/cloud/api/workspaces.html#available-related-resources), there are multiple resources to be included with reading a workspace. But in this client, we do not have the option to do so.

This adds a new method to `ReadWithOptions` that reads a workspace with additional options. 

_Note: there exists another method to `ReadByIDWithOptions`, but that assumes you have the workspace ID. This method takes an organization name and workspace name with the options._

## Testing plan

1. Create a workspace
2. Create state version with outputs
3. Query the workspace and ensure the outputs are present.

This setup is done in the added test.

## External links

- [API documentation](https://www.terraform.io/docs/cloud/api/workspaces.html#available-related-resources)

## Output from tests

```
$ TFE_TOKEN=<token> TFE_ADDRESS=https://tfe-zone-4c292d5b.ngrok.io/ go test -v -run TestWorkspacesReadWithOptions
=== RUN   TestWorkspacesReadWithOptions
=== RUN   TestWorkspacesReadWithOptions/when_options_to_include_resource
--- PASS: TestWorkspacesReadWithOptions (3.33s)
    --- PASS: TestWorkspacesReadWithOptions/when_options_to_include_resource (0.46s)
PASS
ok      github.com/hashicorp/go-tfe     3.459s
```

### DIFF
--- a/workspace.go
+++ b/workspace.go
@@ -24,8 +24,11 @@ type Workspaces interface {
 	// Create is used to create a new workspace.
 	Create(ctx context.Context, organization string, options WorkspaceCreateOptions) (*Workspace, error)
 
-	// Read a workspace by its name.
+	// Read a workspace by its name and organization name.
 	Read(ctx context.Context, organization string, workspace string) (*Workspace, error)
+
+	// ReadWithOptions reads a workspace by name and organization name with given options.
+	ReadWithOptions(ctx context.Context, organization string, workspace string, options *WorkspaceReadOptions) (*Workspace, error)
 
 	// Readme gets the readme of a workspace by its ID.
 	Readme(ctx context.Context, workspaceID string) (io.Reader, error)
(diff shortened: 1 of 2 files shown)


workspace.go
```diff
@@ -131,10 +134,19 @@ type Workspace struct {
...
-	SSHKey       *SSHKey       `jsonapi:"relation,ssh-key"`
+	AgentPool    *AgentPool          `jsonapi:"relation,agent-pool"`
+	CurrentRun   *Run                `jsonapi:"relation,current-run"`
+	Organization *Organization       `jsonapi:"relation,organization"`
+	SSHKey       *SSHKey             `jsonapi:"relation,ssh-key"`
+	Outputs      []*WorkspaceOutputs `jsonapi:"relation,outputs"`
+}
+
+type WorkspaceOutputs struct {
+	ID        string      `jsonapi:"primary,workspace-outputs"`
```
omarismail, line 145: This is required as opposed to using the `StateVersionOutputs` because the primary ID here is `workspace-outputs` while the other is `state-version-outputs`

chrisarcand: * What do you think about using an embedded struct, so it really is more of an alias of the same thing?
* I'm noticing that it looks like workspace outputs attribute names differ ever so slightly from SVO names in the API? 😭

Hastily glancing I think it'd be:

```golang
type WorkspaceOutput struct {
	ID string                `jsonapi:"primary,workspace-outputs"`
	Type      string      `jsonapi:"attr,output-type"`
	StateVersionOutput
}
```

chrisarcand (commented):
Looks good, one quick suggestion...

workspace.go
omarismail: Yeah already I tried to embed the struct, but something about how it unmarshals doesn't work. 

Here is an example: https://play.golang.org/p/63kdiT043q0

If you run it, you will see that name/type don't get unmarshaled. But if you comment `Bar` and uncomment the two other fields, it unmarshals correctly. I didn't dig into the exact reason, but I imagine it probably has to do with how jsonapi is looking at the field tags. I think its worth an independent look at that.

chrisarcand: Ah bummer, alrighty. That'd be neat to fix.

chrisarcand (approved):

PR #254, opened 2021-09-01
Remove redundant run test cleanups

chrisarcand:
## Description

I'm not positive, but believe there is a race condition sometimes happening where a cleanup function - which causes a cascading delete of child resources in TFC - causes a later cleanup function to fail at random.

At worst, I'm wrong and these extra cleanup calls just aren't necessary.  Any time an org or workspace is created in anticipation of being used to create other resource below it, the cleanup for that parent resource will destroy those runs, etc. They don't need to be manually managed and may cause problems like the above.

## Testing plan

The tests, do they pass?

### DIFF
--- a/run_test.go
+++ b/run_test.go
@@ -17,13 +17,10 @@ func TestRunsList(t *testing.T) {
 
 	orgTest, orgTestCleanup := createOrganization(t, client)
 	defer orgTestCleanup()
-	wTest, wTestCleanup := createWorkspace(t, client, orgTest)
-	defer wTestCleanup()
 
-	rTest1, rTestCleanup1 := createRun(t, client, wTest)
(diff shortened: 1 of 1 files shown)


brandonc (approved):
if orgTestCleanup cleans up everything then this is a nice simplification

omarismail (approved):
I think this is a nice cleanup, but I don't think this defer stuff would do it. `defer`'s [execute in LIFO ordering](https://play.golang.org/p/rFkLRhqqYNN), and how they are currently ordered, I don't think it will effect resource deletion in the manner that would cause an error.

Said differently, deleting a run before a workspace before an organization means that when the organization.Delete is called, it is the last one called and all the resources below it should be already deleted.

PR #259, opened 2021-09-14
Add integration tag to tests

barrettclark:
## Description

Add `integration` tag to all tests. This rounds out the changes started in https://github.com/hashicorp/go-tfe/pull/258

## Testing plan

To run tests you will now need to include `tags=integration`. I've updated TESTS.md to reflect the changes needed to run tests now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           command: |
               export GO111MODULE=on
               gotestsum --format short-verbose --junitfile \
-              $TEST_RESULTS_DIR/tests.xml -- -timeout=40m
+              $TEST_RESULTS_DIR/tests.xml -- -timeout=40m -tags=integration
           no_output_timeout: 1800
 
       - store_test_results:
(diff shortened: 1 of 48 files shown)


.circleci/config.yml
```diff
@@ -39,3 +39,6 @@ workflows:
     jobs:
       - run-tests:
           context: core-team-access
+          filters:
+            tags:
+              only: integration
```
omarismail, line 44: I am reading the [circleci tags documentation](https://circleci.com/docs/2.0/configuration-reference/#tags), but I'm not understanding how this relates here. 

We don't have to change anything in the command on line 26 of this file?

barrettclark: I think I was conflating the notion of CircleCI testing tags with the golang tag. No tests were run with this setting, so I force pushed an update that tacks the `--tags=integration` onto the end of the run command that starts on line 26.

barrettclark: CircleCI is running all the tests again ✔️

omarismail (approved):
:shipit:

PR #263, opened 2021-10-04
first pass on running script to generate mocks inside of a github act…

uturunku1:
…ions job

## Description

_Describe why you're making this change._

## Testing plan

1.  _Describe how to replicate_
1.  _the conditions_
1.  _under which your code performs its purpose,_
1.  _including example code to run where necessary._

## External links

## Output from tests

```
$ go test

...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+name: test
+on:
+  pull_request:
+    branches: [ main ]
+jobs:
+  codegen:
+    name: Codegen
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Restore go build cache #This action allows caching dependencies 
+        #and build outputs to improve workflow execution time https://github.com/marketplace/actions/cache
+        - uses: actions/cache@v2
+          with:
+            path: ~/.cache/go-build
+            key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+      - name: Setup Go Environment #https://github.com/marketplace/actions/setup-go-environment
+        - uses: actions/setup-go@v2
+          with:
+            go-version: '1.17.1' # The Go version to download (if necessary) and use.
+        - run: go version
+      - name: Get dependencies
+          run: go mod download
+      - name: Generate mocks
+        run:  ./generate_mocks.sh
+      - name: Ensure nothing changed
+        run: git diff --exit-code
+
+
+ 


.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v2
+      - name: Restore go build cache #This action allows caching dependencies and build outputs to improve workflow execution time 
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+      - name: Setup Go Environment #https://github.com/marketplace/actions/setup-go-environment
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.1' # The Go version to download (if necessary) and use
```
omarismail, line 21: This is great. I think we may forget to update this version whenever we update the [.go-version](https://github.com/hashicorp/go-tfe/blob/main/.go-version). 

what do you think about adding a extra step in here that outputs the content of [.go-version](https://github.com/hashicorp/go-tfe/blob/main/.go-version) and ensures it is `1.17.1`. That way, if someone changes .go-version, this step will fail and we will know to update the Go version here as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo /usr/local/ >> $GITHUB_PATH
+      - name: Get dependencies
+        run: go mod download
+      - name: Generate mocks
+        run:  ./generate_mocks.sh
+      - name: Ensure nothing changed
+        run: git diff --exit-code
+
+
+
```
omarismail, line 38: Nit: could we remove the extra lines?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+          sudo tar xvzf $MOCKGEN_ZIP --strip-components 1 -C /usr/local
+          sudo chmod +x /usr/local/mockgen
+          rm -f $MOCKGEN_ZIP
+          echo /usr/local/ >> $GITHUB_PATH
+      - name: Get dependencies
+        run: go mod download
+      - name: Generate mocks
+        run:  ./generate_mocks.sh
+      - name: Ensure nothing changed
+        run: git diff --exit-code
```
omarismail, line 35: Is this all that is required? I think we need to actually check that the exit code is 0. 

```
go-tfe % git diff --exit-code
go-tfe % echo $?
0
```

omarismail (changes requested):
A few minor changes, but otherwise this is great! Ping me when its ready to review and we could 🚢 this!

.github/workflows/ci.yml
uturunku1: NIce! I love the suggestion

.github/workflows/ci.yml
uturunku1: Can do. I will also check gif olanci-lint can warn us when there are extra lines

PR #275, opened 2021-10-25
WIP: convert workspace include string to typed string constants

uturunku1:
Many of our APIs offer an "include" capability to include other resources. For example, if we take a look at the Workspaces API, one can include many other resources: https://www.terraform.io/cloud-docs/api-docs/workspaces#available-related-resources for the GET endpoints.

In the go-tfe client, we just set this as a string for the `Include` Param in ReadOptions and ListOptions. 

```
type WorkspaceReadOptions struct {
	Include string `url:"include"`
}

// WorkspaceListOptions represents the options for listing workspaces.
type WorkspaceListOptions struct {
	ListOptions

	// A search string (partial workspace name) used to filter the results.
	Search *string `url:"search[name],omitempty"`

	// A search string (comma-separated tag names) used to filter the results.
	Tags *string `url:"search[tags],omitempty"`

	// A list of relations to include. See available resources https://www.terraform.io/docs/cloud/api/workspaces.html#available-related-resources
	Include *string `url:"include,omitempty"`
}
```

This is alright and it works, but its is not entirely useful as it requires someone using this client to then reference the API to see what are the strings available. Occasionally, they made even have a typo in there and not notice until a little late in their process, when they get the Atlas error for passing an incorrect include param.

It would make for a better client experience to know from the Go-tfe code what resources are possible.

This is doable via the use of a typed string in Golang.

## Implementation Details

Each resource that offers Include params will have a typed string that represents that string param that can be pass as part includes query in the GET request. Example:

```
type ConfigurationVersionReadOptions struct {
	Include string `url:"include"`
}
```

becomes:

```
type ConfigurationVersionIncludeOps string

const ConfigurationVerIngressAttributes ConfigurationVersionIncludeOps = "ingress_attributes"

type ConfigurationVersionReadOptions struct {
	Include []ConfigurationVersionIncludeOps `url:"include"`
}
```

Notice that we are defining `Include` struct field a slice of type ConfigurationVersionIncludeOps rather than just a type of ConfigurationVersionIncludeOps.

Most APIs allow for passing multiple Include query params, for which the user would write a string of comma separated values, like `Include : PolicySetReadOptions{"current-version,newest-version"}`. To keep that same behavior as we implement typed strings for those "current-version" and "newest-version" params, we can allow for slice of those typed strings. And now the user would have to do:
`Include: PolicySetReadOptions{[]PolicySetIncludeOps{PolicySetCurrentVersion, PolicySetNewestVersion}`

A trade off here, as we convert those strings to typed strings is that, as we add new Include options for a resource, we'll have come to go-tfe every time to add a new typed string. This is an effort, but a small one, considering the value this change provides to the user experience.

Besides converting types, an additional relevant change I made in tfe.go is adding `func EncodeIncludeParams(v url.Values) string`. With this change. I am overwriting the previous work done by encode() method from the net/url standard go library. The `encode` method was working well, except it won't process a slice of include params, the same way it process a string of comma-separated values for include params.
that Atlas still expects a single string for multiple include params. Example, this how the include looks in the request URL: `include=current-version%2Cnewest-version`. The url.Values in go, which is a map with an include key and a slice of strings which only has one string "current-version,newest-version". After the type change, encode method would make a include key with a slice of strings with two strings [current-version newest-version" which in the request URL looks like `include=current_version&include=newest_version`. The server will only read the last include.


uturunku1:
I made a big mess as I was merging main into this branch and squashing commits, so I am opening a new clean PR to replace this one: https://github.com/hashicorp/go-tfe/pull/294

Issue #276, opened 2021-10-26
admin setting missing global module sharing flag


nhomble:
https://www.terraform.io/docs/cloud/api/admin/organizations.html#update-an-organization

Looking at `AdminOrganization`, it is missing a flag for global module sharing.

Likewise, the request struct `AdminOrganizationUpdateOptions` needs the same update.

PR #278, opened 2021-10-26
add support for GlobalModuleSharing flag on AdminOrganization

nhomble:
## Description

Proposing an enhancement for https://github.com/hashicorp/go-tfe/issues/276

## Testing plan

Updated integration test file with the same approach as `IsDisabled` flag


brandonc:
@nhomble Hey, Nicolas. I'm sorry that this fell off my radar. I think it's still a valuable addition. Is it OK to reopen and get it tested?

nhomble:
@brandonc it would be awesome to get this through, I don't think I'll have bandwidth to run through the integration testing in the short term though :/

brandonc:
@nhomble I'll take care of it from here, thanks! We just [added support](https://github.com/hashicorp/go-tfe/pull/288) for module-consumers so this is an obvious companion.

PR #291, opened 2022-01-19
listOptions in list functions should be pointers

uturunku1:
This PR contains breaking changes and it will get merge against `releases-1-0.x` branch. This is branch dedicated to all changes related to Go-TFE 1.0.

## Description

All listing functions should have pointer types, since they are optional.
There are a many locations where list options are not pointers, which is inconsistent with the
rest of the API.

## Expectations:

1. We want to pass all ListOptions for all API endpoints (like WorkspaceListOptions, WorkspaceTagListOptions, RunListOptions, PolicyCheckListOptions, etc) as pointers. 

2. To give a quick example, https://github.com/hashicorp/go-tfe/blob/26689edbfddfbf08037bcaea9a1a060533148992/workspace.go#L1017 has `WorkspaceTagListOptions` not as a  pointer type.  
So we would convert 
`Tags(ctx context.Context, workspaceID string, options WorkspaceTagListOptions) (*TagList, error)` 
into 
`Tags(ctx context.Context, workspaceID string, options *WorkspaceTagListOptions) (*TagList, error)`  

3. Most, if not all, readOptions params in reading functions, like this one https://github.com/hashicorp/go-tfe/blob/26689edbfddfbf08037bcaea9a1a060533148992/workspace.go#L31 are pointer type . So changing listOptions to be pointers as well, would create parity among all options params for GET endpoints (listing functions and reading functions) because they would match in type.

## Implementation Details

The amount of file changes in this PR is overwhelming, so let me give you a summary here of what you can expect to see for each API endpoint that got modified. 

I am going to use the List() function in workspace.go as an example, but then you can assume I repeated that same change pattern for other listing functions:

1. Refer to WorkspaceListOptions as a pointer type for the List method declared in Workspaces interface:

 https://github.com/hashicorp/go-tfe/blob/0ee09b9783faa294d51cc9b706ad41e794ff494d/workspace.go#L22

2. Refer to WorkspaceListOptions as a pointer in the definition of the function List():

https://github.com/hashicorp/go-tfe/blob/0ee09b9783faa294d51cc9b706ad41e794ff494d/workspace.go#L225

3. In the integration test that is testing this list() function, modified the WorkspaceListOptions accordingly:
   A) In the case where it was declare an empty struct (WorkspaceListOptions{}), we change it to `nil`:
      https://github.com/hashicorp/go-tfe/blob/0ee09b9783faa294d51cc9b706ad41e794ff494d/workspace_integration_test.go#L35

   B) In the case where it was assigning values to fields, we add the `&` reference symbol:
       https://github.com/hashicorp/go-tfe/blob/0ee09b9783faa294d51cc9b706ad41e794ff494d/workspace_integration_test.go#L47

4. Finally, we run `./generate_mocks.sh` so the interface for Workspaces can be updated with the new type pointer in its corresponding mock file.

## Additional-Exceptional changes to certain ListOptions:

1. Notice that in run_trigger.go file, I removed `func (o RunTriggerListOptions) valid() error`  as well as this check `if err := options.valid(); err != nil {` from within List() function.

I did this because now that RunTriggerListOptions can be `nil`, we would get a panic error for pointer dereference if we try to call `valid()` on `options`. We could make adjustments to not get this panic and still run that validation. However, the whole purpose of the `valid()` function was to verified that the fields for runTriggerListOptions do not go undefined or have invalid values. 
We no longer need this `valid()`  function. 
If we pass nil for options to GET request,  the `run_triggers_controller` would do the job of verify the validity of listOptions, and will return the error `bad request\n\nFilter parameter run-trigger type is required and must be either 'inbound' or 'outbound'` if the fields for runTriggerListOptions are not valid, yay.

I made a similar change in a couple of other listing functions.

2. In this code block:

https://github.com/hashicorp/go-tfe/blob/0ee09b9783faa294d51cc9b706ad41e794ff494d/state_version.go#L91-L97

I have added that check for options as nil.
Previously the behavior was that if a user passed `stateVersionListOptions{}` , then we would return the error "organization is required" or "workspace is required" coming from line 95 when those fields are undefined. I cannot removed the `valid()` function in this case. The controller for this get request will not an specific, well explained error (only a 500 internal error) if `stateVersionListOptions` is nil. So, we need to keep that validation check in go-tfe, and on top of that, add a check when options is nil.

I have done something similar for the listing function in team_access.go

### DIFF
--- a/admin_organization.go
+++ b/admin_organization.go
@@ -15,7 +15,7 @@ var _ AdminOrganizations = (*adminOrganizations)(nil)
 // TFE API docs: https://www.terraform.io/docs/cloud/api/admin/organizations.html
 type AdminOrganizations interface {
 	// List all the organizations visible to the current user.
-	List(ctx context.Context, options AdminOrganizationListOptions) (*AdminOrganizationList, error)
+	List(ctx context.Context, options *AdminOrganizationListOptions) (*AdminOrganizationList, error)
 
 	// Read attributes of an existing organization via admin API.
 	Read(ctx context.Context, organization string) (*AdminOrganization, error)
@@ -101,9 +101,9 @@ type AdminOrganizationID struct {
 }
 
 // List all the organizations visible to the current user.
-func (s *adminOrganizations) List(ctx context.Context, options AdminOrganizationListOptions) (*AdminOrganizationList, error) {
+func (s *adminOrganizations) List(ctx context.Context, options *AdminOrganizationListOptions) (*AdminOrganizationList, error) {
 	url := "admin/organizations"
-	req, err := s.client.newRequest("GET", url, &options)
+	req, err := s.client.newRequest("GET", url, options)
 	if err != nil {
 		return nil, err
 	}
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ if err != nil {
 	log.Fatal(err)
 }
 
-orgs, err := client.Organizations.List(context.Background(), tfe.OrganizationListOptions{})
+orgs, err := client.Organizations.List(context.Background(), nil)
 if err != nil {
 	log.Fatal(err)
 }
--- a/admin_organization_integration_test.go
+++ b/admin_organization_integration_test.go
@@ -22,7 +22,7 @@ func TestAdminOrganizations_List(t *testing.T) {
 	defer orgTestCleanup()
 
 	t.Run("with no list options", func(t *testing.T) {
-		adminOrgList, err := client.Admin.Organizations.List(ctx, AdminOrganizationListOptions{})
+		adminOrgList, err := client.Admin.Organizations.List(ctx, nil)
 		require.NoError(t, err)
 
 		// Given that org creation occurs on every test, the ordering is not
@@ -36,7 +36,7 @@ func TestAdminOrganizations_List(t *testing.T) {
 		_, orgTestCleanup := createOrganization(t, client)
 		defer orgTestCleanup()
 
-		adminOrgList, err := client.Admin.Organizations.List(ctx, AdminOrganizationListOptions{
+		adminOrgList, err := client.Admin.Organizations.List(ctx, &AdminOrganizationListOptions{
 			Query: &org.Name,
 		})
 		assert.NoError(t, err)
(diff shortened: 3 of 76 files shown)


admin_organization.go
```diff
@@ -101,7 +101,7 @@ type AdminOrganizationID struct {
 }
 
 // List all the organizations visible to the current user.
-func (s *adminOrganizations) List(ctx context.Context, options AdminOrganizationListOptions) (*AdminOrganizationList, error) {
+func (s *adminOrganizations) List(ctx context.Context, options *AdminOrganizationListOptions) (*AdminOrganizationList, error) {
 	url := "admin/organizations"
 	req, err := s.client.newRequest("GET", url, &options)
```
nfagerlund, line 106: Right, so this is the only thing I think is standing out to me as an issue: 

```suggestion
	req, err := s.client.newRequest("GET", url, options)
```

In the GET functions that were already receiving `options *SomethingOptions` before this PR, it looks like we were just passing the pointer onwards to the client... but in the updated functions, we're now referencing the pointer again and passing a pointer to a pointer. 

I ran some of the updated functions in a test project (passing both `nil` and a proper pointer to an Options), and they work just fine... but for the sake of clarity and consistency, it feels like we should update the spots where we pass the options onward to use single pointers instead of nested ones.

nfagerlund (changes requested):
This mostly looks good to me, and the code seems to work as hoped-for! I do have one request for changes, and that's to update the calls to client methods that consume these options pointers.

admin_organization.go
uturunku1: Indeed! When we call client.newRequest, there is no need to pass the argument options with a reference to a pointer because client.newRequest ends up receiving the address itself, rather than the value. I was trying to think if there would be a good reason to pass a pointer by address rather than passing a pointer by value but I couldn't think of anything. At least on the context of client.newRequest.
To clarify why, when you were testing, both implementation worked just fine:   

https://github.com/hashicorp/go-tfe/blob/3ea3b96bc9080a229116ba94a9f60e2bb7c10918/tfe.go#L475

the .Values function uses the Go Reflect library to unwrap the content of the interface we are passing (in this case &options). Once the Reflect library finds that it's dealing with pointers, it will dig into them until it determines if the values passed are really nil. So, even with our awful nested pointer, the .Values function did a good job in finding the value behind that pointer variable. 

Thanks for pointing this out! I have made the necessary changes. @nfagerlund

nfagerlund (approved):
Excellent, thank you for fixing that! And **extra** thanks for following that back to the Thing That Worked Anyway — I _knew_ you were gonna tell me something new about Go on this one. 😄

PR #294, opened 2022-01-25
Include field string to an slice of typed strings

uturunku1:
I leave this as a draft because this PR includes breaking changes. Please DO NOT MERGE.  The PR is ready for review though!

## Description

Many of our APIs offer an "include" capability to include other resources. For example, if we take a look at the Workspaces API, one can include many other resources: https://www.terraform.io/cloud-docs/api-docs/workspaces#available-related-resources for the GET endpoints.

In the go-tfe client, we just set this as a string for the `Include` Param in ReadOptions and ListOptions. 

```
type WorkspaceReadOptions struct {
	Include string `url:"include"`
}

// WorkspaceListOptions represents the options for listing workspaces.
type WorkspaceListOptions struct {
	ListOptions

	// A search string (partial workspace name) used to filter the results.
	Search *string `url:"search[name],omitempty"`

	// A search string (comma-separated tag names) used to filter the results.
	Tags *string `url:"search[tags],omitempty"`

	// A list of relations to include. See available resources https://www.terraform.io/docs/cloud/api/workspaces.html#available-related-resources
	Include *string `url:"include,omitempty"`
}
```

This is alright and it works, but its is not entirely useful as it requires someone using this client to then reference the API to see what are the strings available. Occasionally, they made even have a typo in there and not notice until a little late in their process, when they get the Atlas error for passing an incorrect include param.

It would make for a better client experience to know from the Go-tfe code what resources are possible.

This is doable via the use of a typed string in Golang.

## Implementation Details

Each resource that offers Include params will have a typed string that represents that string param that can be pass as part includes query in the GET request. Example:

```
type ConfigurationVersionReadOptions struct {
	Include string `url:"include"`
}
```

becomes:

```
type ConfigurationVersionIncludeOps string

const ConfigurationVerIngressAttributes ConfigurationVersionIncludeOps = "ingress_attributes"

type ConfigurationVersionReadOptions struct {
	Include []ConfigurationVersionIncludeOps `url:"include"`
}
```

Notice that we are defining `Include` struct field a slice of type ConfigurationVersionIncludeOps rather than just a type of ConfigurationVersionIncludeOps.

Most APIs allow for passing multiple Include query params, for which the user would write a string of comma separated values, like `Include : PolicySetReadOptions{"current-version,newest-version"}`. To keep that same behavior as we implement typed strings for those "current-version" and "newest-version" params, we can allow for slice of those typed strings. And now the user would have to do:
`Include: PolicySetReadOptions{[]PolicySetIncludeOps{PolicySetCurrentVersion, PolicySetNewestVersion}`

A trade off here, as we add new Include options for a resource, we'll have come to go-tfe every time to add a new typed string that represents that new Include option. This is an effort, but a small one, considering the value this change provides to the user experience.

Besides converting types, an additional relevant change I made, can be found in tfe.go. 

I added `func EncodeIncludeParams(v url.Values) string`. 

I am using this function to replace the previous work done by encode() method from the net/url standard go library. The `encode` method was working well, except  that now, with the changes, it will return a string that looks like this:

`include=current_version&include=newest_version` to represent the include query in the request URL. If we were to send that request URL, the server will only parse the last include param. With the function I added , we are able to get well-format string for the include query: `include=current-version%2Cnewest-version`

### DIFF
--- a/admin_organization.go
+++ b/admin_organization.go
@@ -70,6 +70,14 @@ type AdminOrganizationList struct {
 	Items []*AdminOrganization
 }
 
+// A list of relations to include. See available resources
+// https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
+type AdminOrgIncludeOps string
+
+const (
+	AdminOrgOwners AdminOrgIncludeOps = "owners"
+)
+
 // AdminOrganizationListOptions represents the options for listing organizations via Admin API.
 type AdminOrganizationListOptions struct {
 	ListOptions
@@ -78,9 +86,7 @@ type AdminOrganizationListOptions struct {
 	// Any organizations with a name or notification email partially matching this value will be returned.
 	Query *string `url:"q,omitempty"`
 
-	// A list of relations to include. See available resources
-	// https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
-	Include *string `url:"include"`
+	Include *[]AdminOrgIncludeOps `url:"include,omitempty"`
 }
 
 // AdminOrganizationListModuleConsumersOptions represents the options for listing organization module consumers through the Admin API
--- a/admin_run_integration_test.go
+++ b/admin_run_integration_test.go
@@ -69,7 +69,7 @@ func TestAdminRuns_List(t *testing.T) {
 
 	t.Run("with workspace included", func(t *testing.T) {
 		rl, err := client.Admin.Runs.List(ctx, AdminRunsListOptions{
-			Include: String("workspace"),
+			Include: &([]AdminRunIncludeOps{AdminRunWorkspace}),
 		})
 
 		assert.NoError(t, err)
@@ -81,7 +81,7 @@ func TestAdminRuns_List(t *testing.T) {
 
 	t.Run("with workspace.organization included", func(t *testing.T) {
 		rl, err := client.Admin.Runs.List(ctx, AdminRunsListOptions{
-			Include: String("workspace.organization"),
+			Include: &([]AdminRunIncludeOps{AdminRunWorkspaceOrg}),
 		})
 
 		assert.NoError(t, err)
--- a/admin_user.go
+++ b/admin_user.go
@@ -64,6 +64,14 @@ type AdminUserList struct {
 	Items []*AdminUser
 }
 
+// A list of relations to include. See available resources
+// https://www.terraform.io/docs/cloud/api/admin/users.html#available-related-resources
+type AdminUserIncludeOps string
+
+const (
+	AdminUserOrgs AdminUserIncludeOps = "organizations"
+)
+
 // AdminUserListOptions represents the options for listing users.
 // https://www.terraform.io/docs/cloud/api/admin/users.html#query-parameters
 type AdminUserListOptions struct {
@@ -78,9 +86,7 @@ type AdminUserListOptions struct {
 	// Can be "true" or "false" to show only suspended users or users who are not suspended.
 	SuspendedUsers *string `url:"filter[suspended]"`
 
-	// A list of relations to include. See available resources
-	// https://www.terraform.io/docs/cloud/api/admin/users.html#available-related-resources
-	Include *string `url:"include"`
+	Include *[]AdminUserIncludeOps `url:"include,omitempty"`
 }
 
 // List all user accounts in the Terraform Enterprise installation
--- a/configuration_version.go
+++ b/configuration_version.go
@@ -98,9 +98,18 @@ type CVStatusTimestamps struct {
 	StartedAt  time.Time `jsonapi:"attr,started-at,rfc3339"`
 }
 
+// A list of relations to include. See available resources:
+// https://www.terraform.io/docs/cloud/api/configuration-versions.html#available-related-resources
+type ConfigurationVersionIncludeOps string
+
+const (
+	ConfigurationVerIngressAttributes ConfigurationVersionIncludeOps = "ingress_attributes"
+	ConfigurationRun                  ConfigurationVersionIncludeOps = "run"
+)
+
 // ConfigurationVersionReadOptions represents the options for reading a configuration version.
 type ConfigurationVersionReadOptions struct {
-	Include string `url:"include"`
+	Include []ConfigurationVersionIncludeOps `url:"include,omitempty"`
 }
 
 // ConfigurationVersionListOptions represents the options for listing
@@ -108,9 +117,7 @@ type ConfigurationVersionReadOptions struct {
 type ConfigurationVersionListOptions struct {
 	ListOptions
 
-	// A list of relations to include. See available resources:
-	// https://www.terraform.io/docs/cloud/api/configuration-versions.html#available-related-resources
-	Include *string `url:"include"`
+	Include *[]ConfigurationVersionIncludeOps `url:"include,omitempty"`
 }
 
 // IngressAttributes include commit information associated with configuration versions sourced from VCS.
--- a/policy_set_integration_test.go
+++ b/policy_set_integration_test.go
@@ -284,20 +284,20 @@ func TestPolicySetsRead(t *testing.T) {
 		require.NoError(t, err)
 
 		opts := &PolicySetReadOptions{
-			Include: "current-version,newest-version",
+			Include: []PolicySetIncludeOps{PolicySetCurrentVersion, PolicySetNewestVersion},
 		}
-		ps, err = client.PolicySets.ReadWithOptions(ctx, psTest.ID, opts)
+		psWithOptions, err := client.PolicySets.ReadWithOptions(ctx, psTest.ID, opts)
 		require.NoError(t, err)
 
 		// The newest policy set version is changed to the most recent one
 		// that was created.
-		assert.Equal(t, ps.NewestVersion.ID, psvNew.ID)
-		assert.Equal(t, ps.NewestVersion.Status, PolicySetVersionPending)
+		assert.Equal(t, psWithOptions.NewestVersion.ID, psvNew.ID)
+		assert.Equal(t, psWithOptions.NewestVersion.Status, PolicySetVersionPending)
 		// The current one is now set because policies were uploaded to the
 		// policy set version. Notice how it is set to the one that was uplaoded,
 		// not the newest policy set version.
-		assert.Equal(t, ps.CurrentVersion.ID, psv.ID)
-		assert.Equal(t, ps.CurrentVersion.Status, PolicySetVersionReady)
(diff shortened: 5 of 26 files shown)


sebasslash (changes requested):
Just a couple of notes:

- Optional list option parameters should be a pointer and have `omitempty` in the JSON encoding
- String enums should be constants. Considering they are global (module-wide) variables, we want to be less error prone and ensure they can't be reassigned. 
- Are we missing file changes? I thought there would be more endpoints that make use of the `include` param.

admin_organization.go
```diff
@@ -78,9 +86,7 @@ type AdminOrganizationListOptions struct {
 	// Any organizations with a name or notification email partially matching this value will be returned.
 	Query *string `url:"q,omitempty"`
 
-	// A list of relations to include. See available resources
-	// https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
-	Include *string `url:"include"`
+	Include *[]AdminOrgIncludeOps `url:"include"`
```
brandonc, line 89: In go, a slice is already passed by reference (really, a pointer and 2 ints) so I don't see a lot of value in making this a pointer type and mostly just not idomatic

tfe.go
```diff
@@ -509,6 +510,38 @@ func (c *Client) newRequest(method, path string, v interface{}) (*retryablehttp.
...
+func EncodeIncludeParams(v url.Values) string {
+	if v == nil {
+		return ""
+	}
+	var buf strings.Builder
+	keys := make([]string, 0, len(v))
+	for k := range v {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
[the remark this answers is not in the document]
brandonc: I like it because its predictable. May be for simplified test assertions?

configuration_version.go
```diff
@@ -98,19 +98,25 @@ type CVStatusTimestamps struct {
...
+type ConfigurationVersionIncludeOps string
+
+const (
+	ConfigurationVerIngressAttributes ConfigurationVersionIncludeOps = "ingress_attributes"
+)
+
 // ConfigurationVersionReadOptions represents the options for reading a configuration version.
 type ConfigurationVersionReadOptions struct {
-	Include string `url:"include"`
+	Include []ConfigurationVersionIncludeOps `url:"include"`
```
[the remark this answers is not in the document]
brandonc: I'm in favor of just a slice, not a pointer to a slice here. See previous comment

admin_organization.go
sebasslash: Slices are actually passed by value (but yeah they hold a pointer to the underlying data we care about). Found this [interesting medium article that discusses using a pointer to a slice](https://medium.com/swlh/golang-tips-why-pointers-to-slices-are-useful-and-how-ignoring-them-can-lead-to-tricky-bugs-cac90f72e77b)

uturunku1: OMG! I totally had this conversation about the weirdness of pointer-to-slice practice among some Go libraries with one of my colleagues back at New Relic. But to Brandon's point, I was considering removing this pointer from certain Include fields (though in a different PR) where I am adjusting for instead having ListOptions be a  pointer when passed to a listing functions. So, kind of TBD (if we need pointer for an Include field), depending of which PR gets reviewed first, and then I'll have to make adjustments to the other PR.

sebasslash: From the way the I've come to understand the code, list opts represent any optional URL param otherwise it would be an explicit parameter in the method. So ListOptions should always be a pointer, and I think there are some functions that deviate from this consistency. It's rather annoying to have to pass an empty struct when `nil` would suffice nicely. 

As for the point to not point a slice discussion, I would err on using a pointer *but* for our purposes I don't foresee something breaking if we choose the latter. As long as it's consistent, I'm A-OK 👍

brandonc: A different PR #291 addresses the pointer inconsistency on Options. But this is a field within. As long as the empty slice doesnt not translate to an empty URL param (I don't think it does) I think empty slice might be more user friendly

admin_run_integration_test.go
```diff
@@ -81,7 +81,7 @@ func TestAdminRuns_List(t *testing.T) {
 
 	t.Run("with workspace.organization included", func(t *testing.T) {
 		rl, err := client.Admin.Runs.List(ctx, AdminRunsListOptions{
-			Include: String("workspace.organization"),
+			Include: []AdminRunIncludeOps{AdminRunWorkspaceOrg},
```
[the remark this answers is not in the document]
uturunku1: It is a pointer: https://github.com/hashicorp/go-tfe/blob/daa23cbc0792ac12f26c42f4170820c821e1ae49/admin_run_integration_test.go#L84
Are we looking at different commits?

admin_user.go
```diff
@@ -64,6 +64,14 @@ type AdminUserList struct {
 	Items []*AdminUser
 }
 
+// A list of relations to include. See available resources
+// https://www.terraform.io/docs/cloud/api/admin/users.html#available-related-resources
+type AdminUserIncludeOps string
+
+var (
```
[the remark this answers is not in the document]
uturunku1: Yes, I can make that change. As you may have noticed , I have intentionally had some typed string declared as `var` and some as `const`. This is left-over from my previous approach when I was making the include field a typed string rather than a slice of typed strings. Because a `constant` is not "addressable" (&) for those instance where include was `*string`. Now it's no longer needed, yay.

configuration_version.go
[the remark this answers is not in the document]
uturunku1: I have not converted to a pointer because this Include field was never defined as pointer to string, unlike other include fields within go-tfe where they were intentionally (though inconsistently) defined as pointers. So I can address your suggestion in a different PR but for now, in this PR, I rather stick to one breaking change at a time 😊

tfe.go
```diff
@@ -509,6 +510,38 @@ func (c *Client) newRequest(method, path string, v interface{}) (*retryablehttp.
...
 
+// Encode encodes the values into ``URL encoded'' form
+// ("bar=baz&foo=quux") sorted by key.
+func EncodeIncludeParams(v url.Values) string {
+	if v == nil {
+		return ""
+	}
+	var buf strings.Builder
+	keys := make([]string, 0, len(v))
+	for k := range v {
```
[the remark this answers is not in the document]
uturunku1: Here it is intended to append the Index, not the value.

tfe.go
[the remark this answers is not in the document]
uturunku1: That encoding function is a little confusing to read but to clarify what is going on here:

We are taking input `v` which is map of request url values (for example, map[include:[current_version newest_version]])

We are creating a sorted slice of the keys or indexes in the map. In this case the only key is `include` but if the map was `map[string][]string{"somekey": []string{"somestring"}, "include": []string{"current-version" , "newest-version"} }` , then the sorted slice would be `[]string{"include", "somekey"}`.

Then getting the value for each key **that is present** in the map `v` , and using that strings.builder library to concatenate that key with its values into a single string, a string that has the expected format for request url params, like `include=current-version%2Cnewest-version&somekey=somestring`

policy_set_integration_test.go
```diff
@@ -284,20 +284,20 @@ func TestPolicySetsRead(t *testing.T) {
 		require.NoError(t, err)
 
 		opts := &PolicySetReadOptions{
-			Include: "current-version,newest-version",
+			Include: []PolicySetIncludeOps{PolicySetCurrentVersion, PolicySetNewestVersion},
 		}
-		ps, err = client.PolicySets.ReadWithOptions(ctx, psTest.ID, opts)
```
uturunku1, line 289: I changed this because there was a race condition here, making this integration test fail occasionally.

sebasslash (approved):
Everything looks awesome, great work! I have a question: is `omitempty` in the serialization implied if the field is a pointer and set to `nil`? Or would the resulting URL look like: `&include=nil`

tfe.go
```diff
@@ -509,6 +510,38 @@ func (c *Client) newRequest(method, path string, v interface{}) (*retryablehttp.
 	return req, nil
 }
 
+// Encode encodes the values into ``URL encoded'' form
+// ("bar=baz&foo=quux") sorted by key.
+func EncodeIncludeParams(v url.Values) string {
```
brandonc, line 515: What do you think about adding unit tests for this function?

tfe.go
```diff
@@ -509,6 +510,38 @@ func (c *Client) newRequest(method, path string, v interface{}) (*retryablehttp.
...
+	var buf strings.Builder
+	keys := make([]string, 0, len(v))
+	for k := range v {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		vs := v[k]
+		if len(vs) > 1 {
+			val := strings.Join(vs, ",")
```
brandonc, line 528: I think there might be an issue here. By my reading, this changes the default behavior of url Encode to encoding multiple values as repeated keys (include=A&include=B) to _add_ a third key, (include=A&include=B&include=A,B)

I don't think this is what we want for include parameters, and probably not what we want universally, so maybe you should actually see if k == "include" before encoding as comma separated value. Comma-separated values aren't parsed by rails as an array automatically so it feels a little aggressive to always do this, even though I'm unaware of any other array-style get parameters in go-tfe

brandonc (changes requested):

tfe.go
uturunku1: Can do!

tfe.go
uturunku1: geez :face_palm: , I left behind an import line between 528 and 529: `vs = [:0]`
This was working nonetheless because the right concatenated string (include=A,B) was being appended to the end of the slice, and rails is looking for a param named "include" and matching only that one occurrence. 
I will adjust.

uturunku1:
> Everything looks awesome, great work! I have a question: is `omitempty` in the serialization implied if the field is a pointer and set to `nil`? Or would the resulting URL look like: `&include=null`

Good question!

If the field is a pointer and set to `nil`, for example like this `client.Runs.List(ctx, ID, RunListOptions{ Include: nil,})` then the resulting request URL for the query would be `include=` because `q, err := query.Values(v)` in go-tfe will return a map with the key "include" that has a value of empty slice: `map[include:[]]`

This would be the same resulting URL and outcome even if include was not set to `nil` and listOptions was an empty struct, like so,`client.Runs.List(ctx, ID, RunListOptions{})`

In both cases,  Atlas will properly evaluate this `include=` as if the request had no include params. This is line that ends up catching for any include with empty slice as value: `options[:include] = include_values if include_values.any?`, phew. 
@sebasslash

brandonc (approved):
@uturunku1 let's capture an omitempty audit as a task -- it seems like a lot of include params don't specify it

PR #246, opened 2021-08-19
StateVersionOutputs

omarismail:
## Description

The `Value` field in `StateVersionOutput` can be any number of types, and is currently restricted to being only a string. This issue has been brought up before (#231 #158). 

A recent endpoint was added to [StateVersions to retrieve all the outputs for a given StateVersion](https://www.terraform.io/docs/cloud/api/state-versions.html#list-state-version-outputs). This is different than adding `outputs` to an [include query](https://www.terraform.io/docs/cloud/api/state-versions.html#available-related-resources) due to how we handle permissions. This new endpoint follows _Read state outputs_ in the [Workspace Permissions](https://www.terraform.io/docs/cloud/users-teams-organizations/permissions.html#workspace-permissions).

The value is now an `interface{}` and can be type cast to any other type. See the `TestStateVersionOutputs` for some examples.

## Testing plan

1. Create a workspace and a run that produces outputs
2. Use this new handler and pass in the state version ID.
3. Ensure the outputs are all present and the Value for each one is also present. You can do `fmt.Printf("%v", output.Value)` to see it.

## External links

- [StateVersionOutputs API documentation](https://www.terraform.io/docs/cloud/api/state-versions.html#list-state-version-outputs)
- [jsonapi change](https://github.com/hashicorp/jsonapi/pull/11)
- Issue #231 
- Issue #158 

## Output from tests

```
go-tfe % TFE_TOKEN=<token> TFE_ADDRESS=https://tfe-zone-4c292d5b.ngrok.io/ go test -v -run TestState
=== RUN   TestStateVersionOutputsRead
    --- PASS: TestStateVersionsList/without_list_options (0.27s)                                                                                                                                                                                                                                                                        [168/5345]
    --- PASS: TestStateVersionsList/with_list_options (0.36s)
    --- PASS: TestStateVersionsList/without_an_organization (0.00s)
    --- PASS: TestStateVersionsList/without_a_workspace (0.00s)
=== RUN   TestStateVersionsCreate
=== RUN   TestStateVersionsCreate/with_valid_options
=== RUN   TestStateVersionsCreate/with_the_force_flag_set
=== RUN   TestStateVersionsCreate/with_a_run_to_associate_with
    state_version_test.go:196: This can only be tested with the run specific token
=== RUN   TestStateVersionsCreate/without_md5_hash
=== RUN   TestStateVersionsCreate/withous_serial
=== RUN   TestStateVersionsCreate/without_state
[...]

PR #295, opened 2022-01-26
Allow for pagination of StateVersionOutputs

elsesiy:
## Description

The current implementation does not return the pagination information needed to traverse all output pages. Support for the StateVersion Outputs API has been introduced in https://github.com/hashicorp/go-tfe/pull/246 but it seems like pagination was omitted.

### DIFF
--- a/state_version.go
+++ b/state_version.go
@@ -40,7 +40,7 @@ type StateVersions interface {
 	Download(ctx context.Context, url string) ([]byte, error)
 
 	// Outputs retrieves all the outputs of a state version by its ID.
-	Outputs(ctx context.Context, svID string, options StateVersionOutputsListOptions) ([]*StateVersionOutput, error)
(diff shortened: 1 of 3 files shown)


elsesiy:
@omarismail Can you take a look please & trigger the test workflow? Thanks!

brandonc:
@elsesiy Hello, and thank you for fixing this-- it does look like an oversight to me right now. CI found that there is one additional caller that needs to be changed to the new method signature before this can be retested: `go-tfe/workspace_integration_test.go`

Reviewers: DNM until we deciding how to release this

elsesiy:
@brandonc No worries & thanks for the quick reply. I pushed a fix for the missing integration test, could you trigger the workflow again and merge when convenient please?

brandonc:
@elsesiy I reopened this within the repo as #299 in order to make some mock/test fixes. At a minimum, this needs to be released in the next minor version bump, and we are evaluating what other kinds of breaking changes we have outstanding to include. Thanks again!

PR #297, opened 2022-01-26
fix: API does not allow ssh key text value updating

brandonc:
## Description

From the very beginning it's not been possible to update the value of an SSH key text. Indeed, the tests didn't even test that the value was updated!

The reason for this is the non-repudiation of security auditing. For example, if an attacker uses the ssh key to access a resource, the security audit of the ssh authentication can be trusted to mean key X was always key X and never updated.

## External links

- [API documentation Related PR](https://github.com/hashicorp/terraform-website/pull/2105)

This should be merged into a major release branch!

### DIFF
--- a/ssh_key.go
+++ b/ssh_key.go
@@ -153,9 +153,6 @@ type SSHKeyUpdateOptions struct {
 
 	// A new name to identify the SSH key.
 	Name *string `jsonapi:"attr,name,omitempty"`
-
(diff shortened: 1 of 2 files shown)


sebasslash (approved):
This is all fairly interesting as the test you removed never really validated that the "value" of the ssh key changed.

brandonc:
I don't this rises to the level of "breaking change" because it never worked!

PR #300, opened 2022-02-03
point workflows to release branch

uturunku1:
## Description

Our github actions workflows are running for PRs that want to merge against main branch. We have a new branch "releases-1.0.x" for the purposes of releasing Go-TFE 1.0 and we should have the same github actions applied to this branch.
This is a temporary change. Once we are done with "releases-1.0.x" branch and before it gets merge into main, we should change back the github actions workflows to point back to main branch.

## Suggestion for Testing a workflow

Add any kind of change that would make golangci-lint fail, push that change in this branch. Golangci-lint workflow should  test for `releases-1.0.x` branch and fail with the proper error.
Example: I pushed an unneeded extra line after a function and I got the error `admin_run.go:90: unnecessary trailing newline (whitespace)`.
https://github.com/hashicorp/go-tfe/pull/300/commits/fe6988fe44164ef6b24e65ddb021ff7fef740907

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: test
 on:
   pull_request:
-    branches: [ releases-1.0.x ]
+    branches: [ main, releases-1.0.x ]
 jobs:
   codegen:
     name: Codegen
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: test
 on:
   pull_request:
-    branches: [ main ]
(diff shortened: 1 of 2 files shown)


uturunku1:
> This will eventually get merged into main- should we just append the releases-1.0.x branch to this list?

I was considering that but then it seems unnecessary to have the workflow test for a branch we are not comparing the changes to.

brandonc:
@uturunku1 but shouldn't that release branch be linted/tested as we backport to it?

uturunku1:
I will close this PR and we can take a look at https://github.com/hashicorp/go-tfe/pull/301

PR #301, opened 2022-02-03
add  releases-1.0.x  to the list of branches in GH actions

uturunku1:
## Description

Our github actions workflows are running for PRs that want to merge against main branch. We have a new branch "releases-1.0.x" for the purposes of releasing Go-TFE 1.0 and we should have the same github actions applied to this branch.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,8 +1,7 @@
 name: golangci-lint
 on:
   push:
-    branches:
-      - main
+    branches: [ main, releases-1.0.x ]
   pull_request:
 jobs:
   golangci:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: test
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, releases-1.0.x ]
 jobs:
   codegen:
     name: Codegen


uturunku1:
Am I smelling some flaky test?

.github/workflows/golangci-lint.yml
```diff
@@ -1,8 +1,7 @@
 name: golangci-lint
 on:
   push:
-    branches:
-      - main
+    branches: [ main, releases-1.0.x ]
   pull_request:
```
sebasslash, line 5: So I'm actually curious about why the lint action fires on push and not when a PR is created? Should we move this to when a pull request is created to reduce the # of lint jobs fired?

sebasslash (commented):
Everything else looks g2m

.github/workflows/golangci-lint.yml
brandonc: It probably needs to run on each commit?

uturunku1: That's right. Push event is every commit that gets pushed. Pull request event is when you open a PR, when you synchronize a PR, when you close a PR, and a bunch of other activities that can happen for a PR. Though only some of those activities are enabled by default.

nfagerlund (approved):
Seems fine!

PR #304, opened 2022-02-08
State outputs paged

uturunku1:

```

...
```

### DIFF
--- a/state_version_integration_test.go
+++ b/state_version_integration_test.go
@@ -438,10 +438,10 @@ func TestStateVersionOutputs(t *testing.T) {
 		outputs, err := client.StateVersions.Outputs(ctx, sv.ID, StateVersionOutputsListOptions{})
 		require.NoError(t, err)
 
-		assert.NotEmpty(t, outputs)
+		assert.NotEmpty(t, outputs.Items)
 
 		values := map[string]interface{}{}
-		for _, op := range outputs {
+		for _, op := range outputs.Items {
 			values[op.Name] = op.Value
 		}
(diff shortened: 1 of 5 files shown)


state_version_integration_test.go
```diff
@@ -465,7 +465,7 @@ func TestStateVersionOutputs(t *testing.T) {
 		}
 		outputs, err := client.StateVersions.Outputs(ctx, sv.ID, options)
 		require.NoError(t, err)
-		assert.Empty(t, outputs)
+		assert.Empty(t, outputs.Items)
```
sebasslash, line 468: One thing I would add is asserting the total count and the current page as an extra sanity check, here is an example:
```go
rl, err := client.Runs.List(ctx, wTest.ID, RunListOptions{
    ListOptions: ListOptions{
	PageNumber: 999,
	PageSize:   100,
    },
})
require.NoError(t, err)
assert.Empty(t, rl.Items)
assert.Equal(t, 999, rl.CurrentPage)
assert.Equal(t, 2, rl.TotalCount)
```
To clarify, two runs were created for that workspace so that's why `rl.TotalCount == 2` is asserted

sebasslash (changes requested):
All in all looks good. I didn't run the tests locally however.

state_version_integration_test.go
brandonc: Did you intend to request changes on this? The context is I reverted these commits out of main and re-opened it against releases-1.0.x so all this was already code reviewed

PR #310, opened 2022-02-15
move all static error messages to errors.go

uturunku1:
## Description

This is PR is grouping all instances of errors.New() into a single place.
go-tfe has a file errors.go where we have nicely organized and classified error messages. There are many other places in the repo where we are instantiating errors outside of this file. For the sake of consistency let's keep those errors that are static in errors.go along with other errors that live there.

Besides improving on consistency, another two little issues we are solving with this refactoring are:

1.  There were a bunch of errors.New() outside of errors.go that were creating the same error message unnecessarily.
2. There were two errors created in errors.go that were going unused. They were probably added by the contributor intentionally, but then contributor forgot to use their own error. It happens.

```

...
```

### DIFF
--- a/admin_run.go
+++ b/admin_run.go
@@ -2,7 +2,6 @@ package tfe
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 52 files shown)


sebasslash (approved):
This initial effort is awesome Luces! We'll know if we missed any when we introduce the `goconst` linter into our pipeline.

PR #311, opened 2022-02-15
convert run trigger filter options to a typed constant

uturunku1:
## Description

Convert RunTriggerType string field to a typed constant as part of the go-tfe 1.0 efforts.

### DIFF
--- a/run_trigger.go
+++ b/run_trigger.go
@@ -54,11 +54,19 @@ type RunTrigger struct {
 	Workspace  *Workspace `jsonapi:"relation,workspace"`
 }
 
+// https://www.terraform.io/cloud-docs/api-docs/run-triggers#query-parameters
+type RunTriggerFilterOps string
+
+const (
+	RunTriggerOutbound RunTriggerFilterOps = "outbound"
+	RunTriggerInbound  RunTriggerFilterOps = "inbound"
+)
+
 // RunTriggerListOptions represents the options for listing
 // run triggers.
 type RunTriggerListOptions struct {
 	ListOptions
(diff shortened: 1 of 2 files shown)


run_trigger_integration_test.go
```diff
@@ -89,18 +89,6 @@ func TestRunTriggerList(t *testing.T) {
 		assert.Nil(t, rtl)
 		assert.EqualError(t, err, "bad request\n\nFilter parameter run-trigger type is required and must be either 'inbound' or 'outbound'")
 	})
-
-	t.Run("with invalid run-trigger type", func(t *testing.T) {
```
sebasslash, line 93: On top of deleting a test that covers an invalid run trigger type (string enums prevent this yay), we should test against a valid filter parameter.

sebasslash (changes requested):

run_trigger.go
```diff
@@ -30,6 +30,14 @@ type RunTriggers interface {
 	Delete(ctx context.Context, RunTriggerID string) error
 }
 
+// https://www.terraform.io/cloud-docs/api-docs/run-triggers#query-parameters
+type RunTriggerFilterOps string
+
+const (
+	RunTriggerOutbound = "outbound"
```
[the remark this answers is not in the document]
uturunku1: geez, I can't believe I didn't notice I was missing writing the type right before the `=`.  
Adding it right now!

run_trigger_integration_test.go
uturunku1: indeed!

sebasslash (approved):

PR #313, opened 2022-02-16
Kirch/provider apis

joekarl:

```

...
```

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -711,10 +711,10 @@ func createPrivateRegistryProvider(t *testing.T, client *Client, org *Organizati
 
 	ctx := context.Background()
 
-	privateName := Private
+	privateName := PrivateRegistry
 
 	options := RegistryProviderCreateOptions{
-		Name:         String("name"),
+		Name:         String("tst-name-" + randomString(t)),
 		Namespace:    &org.Name,
 		RegistryName: &privateName,
 	}
@@ -745,11 +745,11 @@ func createPublicRegistryProvider(t *testing.T, client *Client, org *Organizatio
 
 	ctx := context.Background()
 
-	publicName := Public
+	publicName := PublicRegistry
 
 	options := RegistryProviderCreateOptions{
-		Name:         String("name"),
-		Namespace:    String("namespace"),
+		Name:         String("tst-name-" + randomString(t)),
+		Namespace:    String("tst-namespace-" + randomString(t)),
 		RegistryName: &publicName,
 	}
 	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
--- a/helper_test.go
+++ b/helper_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"testing"
 	"time"
@@ -724,7 +725,13 @@ func createPrivateRegistryProvider(t *testing.T, client *Client, org *Organizati
 	}
 
 	return prv, func() {
-		if err := client.RegistryProviders.Delete(ctx, org.Name, prv.RegistryName, prv.Namespace, prv.Name); err != nil {
+		id := RegistryProviderID{
+			OrganizationName: org.Name,
+			RegistryName:     prv.RegistryName,
+			Namespace:        prv.Namespace,
+			Name:             prv.Name,
+		}
+		if err := client.RegistryProviders.Delete(ctx, id); err != nil {
 			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
 				"may exist! The full error is shown below.\n\n"+
 				"Registry Provider: %s/%s\nError: %s", prv.Namespace, prv.Name, err)
@@ -758,7 +765,13 @@ func createPublicRegistryProvider(t *testing.T, client *Client, org *Organizatio
 	}
 
 	return prv, func() {
-		if err := client.RegistryProviders.Delete(ctx, org.Name, prv.RegistryName, prv.Namespace, prv.Name); err != nil {
+		id := RegistryProviderID{
+			OrganizationName: org.Name,
+			RegistryName:     prv.RegistryName,
+			Namespace:        prv.Namespace,
+			Name:             prv.Name,
+		}
+		if err := client.RegistryProviders.Delete(ctx, id); err != nil {
 			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
 				"may exist! The full error is shown below.\n\n"+
 				"Registry Provider: %s/%s\nError: %s", prv.Namespace, prv.Name, err)
@@ -766,6 +779,47 @@ func createPublicRegistryProvider(t *testing.T, client *Client, org *Organizatio
 
 		if orgCleanup != nil {
 			orgCleanup()
+		}
+	}
+}
+
+func createRegistryProviderVersion(t *testing.T, client *Client, provider *RegistryProvider) (*RegistryProviderVersion, func()) {
+	var providerCleanup func()
+
+	if provider == nil {
+		provider, providerCleanup = createPrivateRegistryProvider(t, client, nil)
+	}
+
+	providerId := RegistryProviderID{
+		OrganizationName: provider.Organization.Name,
+		RegistryName:     provider.RegistryName,
+		Namespace:        provider.Namespace,
+		Name:             provider.Name,
+	}
+
+	ctx := context.Background()
+
+	options := RegistryProviderVersionCreateOptions{
+		Version: randomSemver(t),
+	}
+	prvv, err := client.RegistryProviderVersions.Create(ctx, providerId, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return prvv, func() {
+		id := RegistryProviderVersionID{
+			Version:            options.Version,
+			RegistryProviderID: providerId,
+		}
+		if err := client.RegistryProviderVersions.Delete(ctx, id); err != nil {
+			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"Registry Provider Version: %s/%s/%s\nError: %s", prvv.RegistryProvider.Namespace, prvv.RegistryProvider.Name, prvv.Version, err)
+		}
+
+		if providerCleanup != nil {
+			providerCleanup()
 		}
 	}
 }
@@ -1073,6 +1127,10 @@ func randomString(t *testing.T) string {
 		t.Fatal(err)
 	}
 	return v
+}
+
+func randomSemver(t *testing.T) string {
+	return fmt.Sprintf("%d.%d.%d", rand.Intn(99)+3, rand.Intn(99)+1, rand.Intn(99)+1)
 }
 
 // skips a test if the environment is for Terraform Cloud.
--- a/helper_test.go
+++ b/helper_test.go
@@ -702,6 +702,74 @@ func createRegistryModuleWithVersion(t *testing.T, client *Client, org *Organiza
 	}
 }
 
+func createPrivateRegistryProvider(t *testing.T, client *Client, org *Organization) (*RegistryProvider, func()) {
+	var orgCleanup func()
+
+	if org == nil {
+		org, orgCleanup = createOrganization(t, client)
+	}
+
+	ctx := context.Background()
+
+	privateName := Private
+
+	options := RegistryProviderCreateOptions{
+		Name:         String("name"),
+		Namespace:    &org.Name,
+		RegistryName: &privateName,
+	}
+	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return prv, func() {
+		if err := client.RegistryProviders.Delete(ctx, org.Name, prv.RegistryName, prv.Namespace, prv.Name); err != nil {
+			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"Registry Provider: %s/%s\nError: %s", prv.Namespace, prv.Name, err)
+		}
+
+		if orgCleanup != nil {
+			orgCleanup()
+		}
+	}
+}
+
+func createPublicRegistryProvider(t *testing.T, client *Client, org *Organization) (*RegistryProvider, func()) {
+	var orgCleanup func()
+
+	if org == nil {
+		org, orgCleanup = createOrganization(t, client)
+	}
+
+	ctx := context.Background()
+
+	publicName := Public
+
+	options := RegistryProviderCreateOptions{
+		Name:         String("name"),
+		Namespace:    String("namespace"),
+		RegistryName: &publicName,
+	}
+	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return prv, func() {
+		if err := client.RegistryProviders.Delete(ctx, org.Name, prv.RegistryName, prv.Namespace, prv.Name); err != nil {
+			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"Registry Provider: %s/%s\nError: %s", prv.Namespace, prv.Name, err)
+		}
+
+		if orgCleanup != nil {
+			orgCleanup()
+		}
+	}
+}
+
 func createSSHKey(t *testing.T, client *Client, org *Organization) (*SSHKey, func()) {
 	var orgCleanup func()
 
--- a/registry_provider.go
+++ b/registry_provider.go
@@ -22,10 +22,10 @@ type RegistryProviders interface {
 	Create(ctx context.Context, organization string, options RegistryProviderCreateOptions) (*RegistryProvider, error)
 
 	// Read a registry provider
-	Read(ctx context.Context, organization string, registryName RegistryName, namespace string, name string, options *RegistryProviderReadOptions) (*RegistryProvider, error)
+	Read(ctx context.Context, providerId RegistryProviderID, options *RegistryProviderReadOptions) (*RegistryProvider, error)
 
 	// Delete a registry provider
-	Delete(ctx context.Context, organization string, registryName RegistryName, namespace string, name string) error
+	Delete(ctx context.Context, providerId RegistryProviderID) error
 }
 
 // registryProviders implements RegistryProviders.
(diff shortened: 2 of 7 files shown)


sebasslash (commented):
All in all, everything looks awesome so far 🔥 ! Admittedly, I haven't given it a thorough review but a few 1.0 related things stuck out to me looking over your changes.

One thing I'll add is if you add any `include` params to a read or list option to make sure the valid relations that can be included live as string enums #294

uturunku1 (commented):
Hey Karl!
I left two comments for you. 
I am happy to see that your PR is following our go-tfe best practices so far.
Once your PR is no longer a draft and you are ready for a final review, I'd like to smoke test your changes. If you can leave in the PR description detailed instructions on how to smoke test for this feature, I'd greatly appreciate it!

registry_provider_platform.go
```diff
@@ -0,0 +1,16 @@
+package tfe
+
+// RegistryProviderPlatform represents a registry provider platform
+type RegistryProviderPlatform struct {
```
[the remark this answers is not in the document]
joekarl: Yes, this will be filled in, just haven't gotten to it yet...

registry_provider_version.go
```diff
@@ -0,0 +1,173 @@
...
+	return nil
+}
+
+// Create a registry provider
+func (r *registryProviderVersions) Create(ctx context.Context, providerId RegistryProviderID, options RegistryProviderVersionCreateOptions) (*RegistryProviderVersion, error) {
+	if err := providerId.valid(); err != nil {
+		return nil, err
+	}
+	if providerId.RegistryName != PrivateRegistry {
+		return nil, errors.New("only private registry is allowed")
```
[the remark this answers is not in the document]
joekarl: Ah missed that one, thanks!

daniel-laszlo:
Hi @joekarl , @sebasslash , @uturunku1 

We are very much interested in publishing providers to private registry. Do I understand right that this MR is the missing piece for it?

sebasslash:
Hi @daniel-laszlo ! Yes if your means is to use go-tfe, however in the meantime you can do so directly calling [the private registry/provider API](https://www.terraform.io/cloud-docs/api-docs/private-registry/providers).

Once the go-tfe changes are released, we'll subsequently add support to the `tfe` provider.

annawinkler:
I've been thinking about breaking this PR up into 3 separate PRs, and I think I'll do that.

registry_provider_integration_test.go
```diff
@@ -0,0 +1,575 @@
...
+				ListOptions: ListOptions{
+					PageNumber: 0,
+					PageSize:   providerN,
+				},
+			})
+			require.NoError(t, err)
+			assert.NotEmpty(t, returnedProviders.Items)
+			assert.Equal(t, providerN, returnedProviders.TotalCount)
+			assert.Equal(t, 1, returnedProviders.TotalPages)
+			for _, rp := range returnedProviders.Items {
```
annawinkler, line 50: @sebasslash per this [comment](https://github.com/hashicorp/go-tfe/pull/404#discussion_r876429056), are you suggesting that this check be removed from this test because it is unnecessary?

registry_provider_integration_test.go
```diff
@@ -0,0 +1,575 @@
...
+					if rp.ID == p.ID {
+						foundProvider = true
+						break
+					}
+				}
+				assert.True(t, foundProvider, "Expected to find provider %s but did not:\nexpected:\n%v\nreturned\n%v", rp.ID, providers, returnedProviders)
+			}
+		})
+
+		t.Run("returns pages", func(t *testing.T) {
```
annawinkler, line 62: @sebasslash Per this [comment](https://github.com/hashicorp/go-tfe/pull/404#discussion_r876430985), if there isn't enough data to support 999 pages, is that still a valid test? I would guess that we would return pages for data that we have and stop there. 🤔

helper_test.go
```diff
@@ -751,6 +753,167 @@ func createRunTask(t *testing.T, client *Client, org *Organization) (*RunTask, f
...
+	options := RegistryProviderCreateOptions{
+		Name:         "test-registry-provider-" + randomString(t),
+		Namespace:    namespaceName,
+		RegistryName: registryName,
+	}
+
+	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
+
+	if err != nil {
+		t.Fatal(err)
```
annawinkler, line 783: Following up @brandonc re this [comment](https://github.com/hashicorp/go-tfe/pull/405/files#r876368479), I think we're OK using Fatal per my [comment](https://github.com/hashicorp/go-tfe/pull/405/files#r879789207).

registry_provider.go
```diff
@@ -0,0 +1,259 @@
...
+	PrivateRegistry RegistryName = "private"
+	PublicRegistry  RegistryName = "public"
+)
+
+// RegistryProviderIncludeOps represents which jsonapi include can be used with registry providers
+type RegistryProviderIncludeOps string
+
+// List of available includes
+const (
+	RegistryProviderVersionsInclude RegistryProviderIncludeOps = "registry-provider-versions"
```
annawinkler, line 49: @brandonc commented that this is undocumented, and I believe that to be correct. I tested the API locally with this request:

```
curl --location --request GET 'https://LOCALURL/api/v2/organizations/amazing-org/registry-providers/private/amazing-org/zebra-provider?include=registry-provider-versions' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Authorization: Bearer TOKEN' \
--header 'Cookie: __profilin=p%3Dt' \
--data-raw '{
  "data": {
    "type": "registry-providers",
    "attributes": {
      "name": "test-registry-provider-5de6cb61-b5ea-e72e-13a2-c9aab0b5c3ac",
      "namespace": "amazing-org",
      "registry-name": "private"
    }
  }
}'
```

And the response does include the registry provider versions (see the attached json). 

I think we'll need to update the docs but it makes sense to me to include this functionality in go-tfe. What are your thoughts?

[response.txt](https://github.com/hashicorp/go-tfe/files/8758191/response.txt)

registry_provider.go
```diff
@@ -0,0 +1,259 @@
...
+	Organization             *Organization              `jsonapi:"relation,organization"`
+	RegistryProviderVersions []*RegistryProviderVersion `jsonapi:"relation,registry-provider-versions"`
+
+	// Links
+	Links map[string]interface{} `jsonapi:"links,omitempty"`
+}
+
+type RegistryProviderPermissions struct {
+	CanDelete bool `jsonapi:"attr,can-delete,omitempty"`
+	CanUploadAsset bool `jsonapi:"attr,can-upload-asset,omitempty"`
```
annawinkler, line 72: @brandonc `CanDelete` is a permission only for the registry provider, and `CanUploadAsset` is only for a registry provider version. Initially I liked having both permissions in one structure, but I could see splitting this up into 2 separate structs, one for RegistryProvider and one for RegistryProviderVersion. What are your thoughts?

annawinkler: Also see this [comment](https://github.com/hashicorp/go-tfe/pull/405/files#r876424213) & this [comment](https://github.com/hashicorp/go-tfe/pull/405/files#r876425389)

registry_provider_version.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+type RegistryProviderVersionListOptions struct {
+	ListOptions
+}
+
+type RegistryProviderVersionCreateOptions struct {
+	Version   string   `jsonapi:"attr,version"`
+	KeyID     string   `jsonapi:"attr,key-id"`
+	Protocols []string `jsonapi:"attr,protocols"`
```
annawinkler, line 73: Suggestion to make this strongly typed (see this [comment](https://github.com/hashicorp/go-tfe/pull/405/files#r876436790)).

registry_provider.go
brandonc: I found it in the controller! It appears that it just need to be documented.

registry_provider.go
brandonc: Yeah, that is my preference

registry_provider.go
```diff
@@ -0,0 +1,256 @@
...
+	Type string `jsonapi:"primary,registry-providers"`
+
+	Name         string       `jsonapi:"attr,name"`
+	Namespace    string       `jsonapi:"attr,namespace"`
+	RegistryName RegistryName `jsonapi:"attr,registry-name"`
+}
+
+type RegistryProviderReadOptions struct {
+	// Optional: Include related jsonapi relationships
+	Include *[]RegistryProviderIncludeOps `url:"include,omitempty"`
```
brandonc, line 119: The empty value of a slice is nil, which makes pointers to slices a little symbolically redundant and harder to use. I suggest removing the pointer from the typedef

registry_provider.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (r *registryProviders) List(ctx context.Context, organization string, options *RegistryProviderListOptions) (*RegistryProviderList, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+	if options != nil {
```
brandonc, line 126: 💅 if your `valid` function used a pointer receiver, you could test for nil there. This could reduce the conditions in this function.

registry_provider_platform.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Links
+	Links map[string]interface{} `jsonapi:"links,omitempty"`
+}
+
+// RegistryProviderPlatformID is the multi key ID for identifying a provider platform
+type RegistryProviderPlatformID struct {
+	RegistryProviderVersionID
+	OS   string `jsonapi:"attr,os"`
+	Arch string `jsonapi:"attr,arch"`
```
brandonc, line 53: I don't think these fields need encoding metadata because they will not be encoded in a jsonapi request.

brandonc (commented):
Submitting a partial review. I still need to review the tests. Looks great!

registry_provider.go
annawinkler: I'll add a ticket to update the documentation.

registry_provider_version.go
annawinkler: Per an internal discussion, I'm going to hold off on making the protocols strongly typed because there is a discrepancy between internal documents regarding what the values should be. I've opened a ticket to look at this issue more.

registry_provider.go
annawinkler: I was wondering what our overall approach is in the repo, and I searched for `CreateOptions) valid`, and noticed that our template example does take this approach, but given 26 other uses, only 3 use a pointer 🤔 Maybe this is an asana issue for more 💅 cleanup!

registry_provider_version.go
brandonc: Once we release this, we can't really change it without incrementing the major version since its part of the public interface. I think we should decide now if this is the API we want and then never change it.

sebasslash: My 2 🪙  if the `Protocols` field is a limited set of known values, they should definitely be enumerated. This is a pattern we've enforced with v1.0. 
```golang
type ProviderProtocol string
const (
   // values here
)
```

registry_provider_version.go
```diff
@@ -34,15 +34,15 @@ type registryProviderVersions struct {
...
-	Permissions        RegistryProviderPermissions `jsonapi:"attr,permissions"`
-	ShasumsUploaded    bool                        `jsonapi:"attr,shasums-uploaded"`
-	ShasumsSigUploaded bool                        `jsonapi:"attr,sasums-sig-uploaded"`
+	ID                 string                             `jsonapi:"primary,registry-provider-versions"`
+	Version            string                             `jsonapi:"attr,version"`
+	CreatedAt          string                             `jsonapi:"attr,created-at,iso8601"`
+	UpdatedAt          string                             `jsonapi:"attr,updated-at,iso8601"`
+	KeyID              string                             `jsonapi:"attr,key-id"`
+	Protocols          []string                           `jsonapi:"attr,protocols,omitempty"`
+	Permissions        RegistryProviderVersionPermissions `jsonapi:"attr,permissions"`
```
sebasslash, line 43: Should be `*RegistryProviderVersionPermissions`

registry_provider_version.go
```diff
@@ -34,15 +34,15 @@ type registryProviderVersions struct {
...
-	Protocols          []string                    `jsonapi:"attr,protocols,omitempty"`
-	Permissions        RegistryProviderPermissions `jsonapi:"attr,permissions"`
-	ShasumsUploaded    bool                        `jsonapi:"attr,shasums-uploaded"`
-	ShasumsSigUploaded bool                        `jsonapi:"attr,sasums-sig-uploaded"`
+	ID                 string                             `jsonapi:"primary,registry-provider-versions"`
+	Version            string                             `jsonapi:"attr,version"`
+	CreatedAt          string                             `jsonapi:"attr,created-at,iso8601"`
+	UpdatedAt          string                             `jsonapi:"attr,updated-at,iso8601"`
+	KeyID              string                             `jsonapi:"attr,key-id"`
+	Protocols          []string                           `jsonapi:"attr,protocols,omitempty"`
```
sebasslash, line 42: omitempty is not necessary here as the `Protocols` field will be initialized to `[]` if the corresponding json field is empty.

sebasslash (changes requested):
Looking great! *Almost* there across the finish line 🐎

registry_provider_integration_test.go
sebasslash: I think it's a simpler check to use `999` for the page number because obviously there aren't 999 pages, so we should expect zero items returned. Which inevitably means the list query params are working as expected.

registry_provider_version.go
annawinkler: @brandonc and I chatted about this field today. and it seems like the intent is to provide major version compatibility. Given that this is a library, it seems like our best bet is to keep this field a string, and let the API do the checking (which it does currently).

registry_provider_version.go
annawinkler: Hm... this was changed per a comment from @brandonc because the permissions are required. Maybe I'm missing something? 🤔

registry_provider_version.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// RegistryProviderVersionID is the multi key ID for addressing a version provider
+type RegistryProviderVersionID struct {
+	RegistryProviderID
+	Version string
+}
+
+type RegistryProviderVersionPermissions struct {
+	CanUploadAsset bool `jsonapi:"attr,can-upload-asset,omitempty"`
```
brandonc, line 62: This can't be omitted because bool can't be empty. Or, perhaps false values would be considered empty? Either way, it's undesired

brandonc: It looks like can_delete is also a field in this [response](https://github.com/hashicorp/atlas/blob/main/app/serializers/api/v2/registry_provider_version_serializer.rb#L66-L69)?

registry_provider_version.go
brandonc: Right... This struct is never empty in the response so I didn't think a pointer is necessary for two bools

registry_provider_version.go
sebasslash: Correct, the only way you can omit a bool is to make the value a pointer.

annawinkler: That's a good point. And in reviewing the [docs](https://app.asana.com/0/1202314304798782/1202383446952123/f) for registry provider version, there are two permissions that are serialized. I'll update this.

registry_provider_platform.go
```diff
@@ -0,0 +1,233 @@
...
+	ListOptions
+}
+
+// Create a new registry provider platform
+func (r *registryProviderPlatforms) Create(ctx context.Context, versionID RegistryProviderVersionID, options RegistryProviderPlatformCreateOptions) (*RegistryProviderPlatform, error) {
+	if err := versionID.valid(); err != nil {
+		return nil, err
+	}
+
+	if versionID.RegistryName != PrivateRegistry {
```
uturunku1, line 79: I see that the block above `versionID.valid()` is validating the RegistryName can only be "private", why do we validate registryName again on this line?

registry_provider_platform.go
```diff
@@ -0,0 +1,233 @@
...
+// RegistryProviderPlatformID is the multi key ID for identifying a provider platform
+type RegistryProviderPlatformID struct {
+	RegistryProviderVersionID
+	OS   string
+	Arch string
+}
+
+// RegistryProviderPlatformCreateOptions represents the set of options for creating a registry provider platform
+type RegistryProviderPlatformCreateOptions struct {
+	OS       string `jsonapi:"attr,os"`
```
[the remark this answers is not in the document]
annawinkler: It looks like there are some inconsistencies regarding how we indicate if a field is required. I'll open an internal issue for us to look at this, take a decision, update our contributing docs, and perhaps make the code consistent.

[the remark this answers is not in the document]
annawinkler: Re-reviewing our contributing [docs](https://github.com/hashicorp/go-tfe/blob/main/docs/CONTRIBUTING.md#best-practices-for-adding-a-new-endpoint), I'll follow that example.

annawinkler:
A question was raised about [these](https://github.com/hashicorp/go-tfe/blob/4a9b53b8b5297986638b44a112b1169ff6c280b8/registry_provider_version.go#L196-L238) methods - should they be exported? I don't think they should be.

annawinkler:
> A question was raised about [these](https://github.com/hashicorp/go-tfe/blob/4a9b53b8b5297986638b44a112b1169ff6c280b8/registry_provider_version.go#L196-L238) methods - should they be exported? I don't think they should be.

Actually, we do want to export them. The idea is the workflow would be:

- Create a registry provider version
- Get the upload URLs from the newly-created registry provider version and use them to upload the shasums and shasums sig. Alternatively, could check the uploaded boolean values to see if the items have been uploaded, and if not upload them.
- Then later could use the download URLs to download the shasums and shasums sig.

registry_provider_platform_integration_test.go
```diff
@@ -0,0 +1,398 @@
...
+			Namespace:        provider.Namespace,
+			Name:             provider.Name,
+		},
+		Version: version.Version,
+	}
+
+	t.Run("with valid options", func(t *testing.T) {
+		options := RegistryProviderPlatformCreateOptions{
+			OS:       "foo",
+			Arch:     "scrimbles",
```
brandonc, line 38: ah, yes, scrimbles64

brandonc (dismissed):
Very nice

registry_provider_platform_integration_test.go
uturunku1: Lol, I couldn't find the meaning of "scrimbles" in the dictionary, but this sounds funny 😆

annawinkler: It's a made up word from a former coworker from long ago that I still use 🥳

sebasslash (approved):
Woohooo @annawinkler and @joekarl !!! :celebration_dance

brandonc (approved):

PR #314, opened 2022-02-16
add omitempty to json tag for struct fields that are optional

uturunku1:
## Description

I audited the repo looking for options struct fields to update for omitempty json tag: 

The fields that are not required should have the tag because if their values are null then the fields will get omitted from the encoding.
The fields that are required should not have the tag.

This PR is bringing those two types of changes.

I have added omitempty to
Two url filter tags that are optional: https://www.terraform.io/cloud-docs/api-docs/admin/users#query-parameters
One shh key attribute said to be optional: https://www.terraform.io/cloud-docs/api-docs/oauth-tokens#request-body

And I have removed omitempty from three fields that are required. For those fields, I didn't need to look in the documentation to find out if they are optional or not. Right there, in the same go file, I can see that we are running a validation check (options.valid()) to verify that those fields are not empty values. 

There will be a follow up PR where I'll be adding a few more options.valid() for ListOptions that have required fields so we can match the other ones that already have this validation.

```

...
```

### DIFF
--- a/admin_user.go
+++ b/admin_user.go
@@ -81,10 +81,10 @@ type AdminUserListOptions struct {
 	Query string `url:"q,omitempty"`
 
 	// Can be "true" or "false" to show only administrators or non-administrators.
-	Administrators string `url:"filter[admin]"`
+	Administrators string `url:"filter[admin],omitempty"`
 
 	// Can be "true" or "false" to show only suspended users or users who are not suspended.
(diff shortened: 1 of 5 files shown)


brandonc (commented):
I think the Include fields on run_task and task_stages can also be omitempty. I missed them in my review of that endpoint

PR #315, opened 2022-02-16
fix: vscode go settings for test files

brandonc:
This should fix the issue of the go language server not being able to give editor feedback in *_test.go files.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "go.buildFlags": [
+      "-tags=integration"
+  ],
+  "go.testTags": "integration",
+}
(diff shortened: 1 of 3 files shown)


sebasslash (commented):
Pulled these changes and it definitely resolves the issue on vscode. However I think we should offer more general guidance for configuring `gopls` in `TESTS.md`. The configuration (at least for gopls) is virtually the same, barring the configuration file you have to edit.

For Neo/Vim users using `coc.nvim` the file can be opened using `:CocConfig` and adding this entry to the `coc-settings.json` file:
```json
"go.goplsOptions" : {
  "buildFlags": [ "-tags=integration" ],
  "env": {
    "GOFLAGS" : "-tags=integration"
  }
}
```

sebasslash (approved):
Love this. Great starting point; I imagine as the 1.0 effort continues we'll add more to `CONTRIBUTING.md`

sebasslash (approved):

PR #316, opened 2022-02-18
Add scaffold code example to CONTRIBUTING.md

brandonc:
## Description

Adds a complete example resource to CONTRIBUTING.md

## Testing plan

Obviously this isn't working code but I don't think it contains any compilation errors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,10 +2,269 @@
 
 If you find an issue with this package, please create an issue in GitHub. If you'd like, we welcome any contributions. Fork this repository and submit a pull request.
 
-### Writing Tests
+## Writing Tests
 
 The test suite contains many acceptance tests that are run against the latest version of Terraform Enterprise. You can read more about running the tests against your own Terraform Enterprise environment in [TESTS.md](TESTS.md). Our CI system (Circle) will not test your fork unless you are an authorized employee, so a HashiCorp maintainer will initiate the tests or you and report any missing tests or simple problems. In order to speed up this process, it's not uncommon for your commits to be incorportated into another PR that we can commit test changes to.
 
-### Editor Settings
+## Editor Settings
 
 We've included VSCode settings to assist with configuring the go extension. For other editors that integrate with the [Go Language Server](https://github.com/golang/tools/tree/master/gopls), the main thing to do is to add the `integration` build tags so that the test files are found by the language server. See `.vscode/settings.json` for more details.
+
+## Adding a new Endpoint
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,10 +2,269 @@
...
+}
+
+// List all the examples for an organization
+func (s *example) List(ctx context.Context, organization string, options *ExampleListOptions) (*ExampleList, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+
+	u := fmt.Sprintf("organizations/%s/examples", url.QueryEscape(organization))
+	req, err := s.client.newRequest("GET", u, &options)
```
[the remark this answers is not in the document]
uturunku1: To expand, we do need it for all POST request but not for GET requests

CONTRIBUTING.md
```diff
@@ -2,10 +2,269 @@
...
+	return s.ReadWithOptions(ctx, exampleID, nil)
+}
+
+// ExampleReadOptions represents the set of options for reading an example
+type ExampleReadOptions struct {
+	Include []RunTaskIncludeOps `url:"include,omitempty"`
+}
+
+// Read is used to read an organization's example by ID with options
+func (s *example) ReadWithOptions(ctx context.Context, exampleID string, options *ExampleReadOptions) (*Example, error) {
```
uturunku1, line 181: Maybe we can make a reference here of the new rule we were discussing with Sebastian on our proj-go-tfe-1-0 channel: If the ID required is going to be composed of more than two individual strings, say `ReadWithOptions(ctx context.Context, exampleID, example2ID, example3ID string, options *ExampleReadOptions)`, then we prefer this to be passed as a single struct that contains the ID parts as fields for that struct.  like this example: https://github.com/hashicorp/go-tfe/blob/546d40891d90d05d42550682c8735d7dc9428bca/registry_provider.go#L173-L178

uturunku1 (changes requested):

CONTRIBUTING.md
brandonc: Does it need to be decorated as jsonapi in your example, or wouldn't we expect a plain ol struct?

CONTRIBUTING.md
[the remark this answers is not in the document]
brandonc: I fixed this in run_task.go as well in this PR

CONTRIBUTING.md
brandonc: I don't really want to complicate the example to accommodate multiple IDs

uturunku1 (approved):

PR #319, opened 2022-02-18
prefix constants to prevent export

laurenolivia:
## Description

This PR is intended to prevent constants in the `tfe.go` file from being exported. 

Top-level variables and constants have a package scope, meaning that all the constants declared on top of go-tfe files, even when they start with a lowercase letter, can be used in other go-tfe files. So all those generic names could accidentally get overwritten or use the wrong value in a different file.

## Testing plan

1. Find a constant in the `tfe.go` file that is prefixed with an underscore.
2. Verify that you are unable to reference that constant in a different file.

## External links

- [Asana](https://app.asana.com/0/1200894272253435/1201804478448342/f)

```

...
```

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -28,10 +28,10 @@ import (
 )
 
 const (
-	userAgent        = "go-tfe"
-	headerRateLimit  = "X-RateLimit-Limit"
-	headerRateReset  = "X-RateLimit-Reset"
-	headerAPIVersion = "TFP-API-Version"
+	_userAgent        = "go-tfe"
+	_headerRateLimit  = "X-RateLimit-Limit"
+	_headerRateReset  = "X-RateLimit-Reset"
+	_headerAPIVersion = "TFP-API-Version"
 
 	// DefaultAddress of Terraform Enterprise.
 	DefaultAddress = "https://app.terraform.io"
@@ -81,7 +81,7 @@ func DefaultConfig() *Config {
 	}
 
 	// Set the default user agent.
-	config.Headers.Set("User-Agent", userAgent)
+	config.Headers.Set("User-Agent", _userAgent)
 
 	return config
 }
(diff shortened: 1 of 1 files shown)


brandonc (commented):
I don't have a strong opinion one way or the other, but I don't see a lot of value in this change. And actually, more in favor of Capitalizing these particular variables and exporting them because of their importance.

@uturunku1 What exactly are we trying to avoid by appending a prefix? Accidental shadowing within the tfe package? The captialization may help with that as well.

brandonc:
I'd like to add that prefixing _ doesn't change any behavior, but I could see using it for implementation-specific variables and not these in particular.

uturunku1:
@brandonc I requested this change because un-exported variable names (without capitalized first letter) have a package scope, meaning they can be used in other Go files within `package tfe`. As we know this project is one single large package. Is a potential risk in the future anyone could accidentally use the wrong value in a different file for a generic variable name. 

We can test this locally with `userAgent` variable from tfe.go file:

Go to another file in the package like, workspace.go and write `userAgent := "luces"`
Now print the value of the variable with fmt.Println("userAgent", userAgent) to see what happens.

Underscoring the variable should help:

Start writing `userAgent =` somewhere in workspace.go

If you hover over the variable name, you should see a message like `const userAgent untyped string = "go-tfe"` because it's value its visible within the package.

Now write underscore to the front of the variable userAgent in tfe.go file.

Then go back workspace.go and write `userAgent := "luces"`, this time if you hover over userAgent you should not have visibility into the value of _userAgent.  

You are very right that the underscore does not make things invisible, but it does force the contributor to be more intentional about reusing generic names. 

For most Go project composed by multiple packages, I would not bother about underscoring because lowercased variables are not available to other packages, but as go-tfe continues to bring more and more files into our one single package, I would not be surprised in we run into this little accident at some point. Of course the best solution to this is restructured the project's directory/package, but when is that going to happen.

brandonc (approved):
Seems fine

PR #320, opened 2022-02-18
use %w to wrap upstream errors within fmt.Errorf

brandonc:
## Description

Finally get consistent with returning errors that contain other errors

also improves the error related to unmarshaling Items that are not defined as a slice

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -135,3 +135,15 @@ var (
 	// ErrInvalidTerraformVersionType is returned when the type is not valid.
 	ErrInvalidTerraformVersionType = errors.New("invalid type for terraform version. Please use 'terraform-version'")
 )
+
+// Library errors that usually indicate a bug in the implementation of go-tfe
+var (
+	// ErrBugItemsSlice is returned when an API response attribute called Items is not a slice
+	ErrBugItemsSlice = errors.New(`model field "Items" must be a slice`)
+
+	// ErrBugRequestBodyType is returned when a request body for DELETE/PATCH/POST is not a reference type
+	ErrBugRequestBodyType = errors.New("go-tfe bug: DELETE/PATCH/POST body must be nil, ptr, or ptr slice")
+
+	// ErrBugStructFieldTags is returned when a mix of json and jsonapi tagged fields are used in the same struct
+	ErrBugStructFieldTags = errors.New("go-tfe bug: struct can't use both json and jsonapi attributes")
+)
--- a/tfe.go
+++ b/tfe.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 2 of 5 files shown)


tfe.go
```diff
@@ -639,22 +639,22 @@ func unmarshalResponse(responseBody io.Reader, model interface{}) error {
...
 	// Items and Pagination struct fields.
 	if !items.IsValid() || !pagination.IsValid() {
 		return jsonapi.UnmarshalPayload(responseBody, model)
 	}
 
-	// Return an error if v.Items is not a slice.
+	// Return an error if model.Items is not a slice.
 	if items.Type().Kind() != reflect.Slice {
-		return fmt.Errorf("v.Items must be a slice")
+		return fmt.Errorf("model field \"Items\" must be a slice")
```
uturunku1, line 650: Can you change this to create a static error with errors.New. No need to use fmt.Errorf because we are not formatting anything. Weird that the golangci-lint didn't catch this, but maybe we have not enable one of those lint libraries that looks after unneeded formatting.
Let's have that errors.New as part of the errors.go file

tfe_integration_test.go
```diff
@@ -471,7 +471,7 @@ func Test_unmarshalResponse(t *testing.T) {
 		}
 		err := unmarshalResponse(responseBody, &malformattedItemStruct)
 		require.Error(t, err)
-		assert.EqualError(t, err, "v.Items must be a slice")
+		assert.EqualError(t, err, "model field \"Items\" must be a slice")
```
uturunku1, line 474: Heads up that this file does not longer contain this test in releases branch. We moved this test to a new file called tfe_test.go. In this new file we are testing for helper functions.

uturunku1 (changes requested):
Nice catch noticing that v argument got renamed to model.
I left two comments.

tfe.go
uturunku1: Oh, also, no need to escape quotes. Go uses `` back-ticks to allow for quotes inside of a string. So this should be errors.New(`Model field "Items" must be a slice`)

uturunku1: Ironically, github is formatting my errors.New message inside of the parenthesis because I put backticks in there 😆

tfe_integration_test.go
brandonc: Whoever rebases releases-1.0.x will need to worry about this

errors.go
```diff
@@ -135,3 +135,15 @@ var (
...
 	ErrInvalidTerraformVersionType = errors.New("invalid type for terraform version. Please use 'terraform-version'")
 )
+
+// Library errors that usually indicate a bug in the implementation of go-tfe
+var (
+	// ErrBugItemsSlice is returned when an API response attribute called Items is not a slice
+	ErrBugItemsSlice = errors.New(`model field "Items" must be a slice`)
+
+	// ErrBugRequestBodyType is returned when a request body for DELETE/PATCH/POST is not a reference type
+	ErrBugRequestBodyType = errors.New("go-tfe bug: DELETE/PATCH/POST body must be nil, ptr, or ptr slice")
```
uturunku1, line 145: Heads up, ErrBugRequestBodyType and ErrBugStructFieldTags already were added to this file in releases branch under a slightly different name. Though I do not mind rebasing releases branch to have your names instead if this is the intent.

uturunku1 (changes requested):

brandonc:
I'll just make this a releases-1.0 PR since I created so many conflicts

brandonc:
see https://github.com/hashicorp/go-tfe/pull/324

PR #322, opened 2022-02-22
use options.valid helper function for required fields

uturunku1:

```

...
```


uturunku1:
I changed my mind about including some reorg of functions as I was making changes for options.valid(). It made the readability of the PR very confusing. I am opening a new PR exclusively with changes related to options.valid: https://github.com/hashicorp/go-tfe/pull/325

PR #324, opened 2022-02-22
use %w to wrap upstream errors within fmt.Errorf

brandonc:
## Description

Get more consistent with returning errors that wrap other errors

also improves the error related to unmarshaling Items that are not defined as a slice


uturunku1 (approved):
🏁🐎

PR #325, opened 2022-02-22
use options.valid for option fields that are required

uturunku1:
## Description

Create consistency around the usage of options.valid()

This PR implements three changes for this purpose:

1. That all types of validations for the values of the `options` param happens within the valid function.  Previously there were a couple of validation checks happening outside of the valid function, like this one:

```
if options == nil {
		return nil, ErrRequireTeamAccessListOps
}

		return nil, err
}
```

`if options == nil` should happen inside of valid().

2.  That when we call valid(), we follow the same syntax, that should be:

```
		return nil, err
}
```

This syntax deviates:

```
if !options.valid() {
		return nil, ErrInvalidSMTPAuth
}
```

3. That all required fields for an options param validate that its fields have values assigned. These were the options params that were missing validation for their fields:

AdminTwilioSettingsVerifyOptions
AdminTerraformVersionCreateOptions
RunTriggerListOptions

4. Finally, add a map[RunTriggerFilterOps]struct{}{} to validate that the filter field for the options param is only allow to use a value present in that map. I followed the example set in Admin Settings SMTP.

```
var validSMTPAuthType = map[SMTPAuthType]struct{}{
	SMTPAuthNone:  {},
	SMTPAuthPlain: {},
	SMTPAuthLogin: {},
}
```

```

...
```

### DIFF
--- a/admin_run.go
+++ b/admin_run.go
@@ -69,10 +69,8 @@ type AdminRunsListOptions struct {
 // List all the runs of the terraform enterprise installation.
 // https://www.terraform.io/docs/cloud/api/admin/runs.html#list-all-runs
 func (s *adminRuns) List(ctx context.Context, options *AdminRunsListOptions) (*AdminRunsList, error) {
-	if options != nil {
-		if err := options.valid(); err != nil {
-			return nil, err
-		}
+	if err := options.valid(); err != nil {
+		return nil, err
 	}
 
 	u := "admin/runs"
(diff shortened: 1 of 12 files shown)


admin_run.go
```diff
@@ -114,7 +112,10 @@ func (s *adminRuns) ForceCancel(ctx context.Context, runID string, options Admin
 }
 
 // Check that the field RunStatus has a valid string value
-func (o AdminRunsListOptions) valid() error {
+func (o *AdminRunsListOptions) valid() error {
+	if o == nil { //no need to validate fields
```
brandonc, line 116:
```suggestion
	if o == nil { // no need to validate fields
```

would our new linters do it like this?

brandonc (approved):
Nice! Since valid() now returns error instead of bool, I might rename them all validate() but either works

admin_run.go
uturunku1: Yes, not sure which lint library, but one of them forces you to add a space in between the `//` and the comment

PR #327, opened 2022-02-22
Use consistent naming across resource methods

sebasslash:
## Description

This PR introduces a refactor for several methods that poorly express an action on a relationship between two resources. By using verbs, we can be more specific as to what action we're performing when calling these methods. An example would be
`ModuleConsumers()` -- from this method signature we can't necessarily tell what action this method is performing without
referring to the docs. Renaming it to `ListModuleConsumers()` we can now assess what the function does more conveniently.

### DIFF
--- a/organization.go
+++ b/organization.go
@@ -31,14 +31,14 @@ type Organizations interface {
 	// Delete an organization by its name.
 	Delete(ctx context.Context, organization string) error
 
-	// Capacity shows the current run capacity of an organization.
-	Capacity(ctx context.Context, organization string) (*Capacity, error)
+	// ReadCapacity shows the current run capacity of an organization.
+	ReadCapacity(ctx context.Context, organization string) (*Capacity, error)
 
-	// Entitlements shows the entitlements of an organization.
-	Entitlements(ctx context.Context, organization string) (*Entitlements, error)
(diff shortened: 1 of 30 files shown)


organization.go
```diff
@@ -31,14 +31,14 @@ type Organizations interface {
...
 
-	// Capacity shows the current run capacity of an organization.
-	Capacity(ctx context.Context, organization string) (*Capacity, error)
+	// ReadCapacity shows the current run capacity of an organization.
+	ReadCapacity(ctx context.Context, organization string) (*Capacity, error)
 
-	// Entitlements shows the entitlements of an organization.
-	Entitlements(ctx context.Context, organization string) (*Entitlements, error)
+	// ListEntitlements shows the entitlements of an organization.
+	ListEntitlements(ctx context.Context, organization string) (*Entitlements, error)
```
uturunku1, line 38: Should we name this `ReadEntitlements` instead? Though the name of the resource is plural, is returning a single struct. Also, if we take a look at the controllers file in Atlas for this resource, it's calling the `def show` method which in go-tfe we usually translate as ReadSomeResource method.

brandonc (commented):
We need to make a note about this impactful change in order to highlight it in the release notes

PR #329, opened 2022-02-23
Sync run task changes with 1.0 effort

sebasslash:
Updates the run task related changes to use some of the conventions established by the 1.0 release effort.

### DIFF
--- a/task_stages_integration_test.go
+++ b/task_stages_integration_test.go
@@ -32,12 +32,12 @@ func TestTaskStagesRead(t *testing.T) {
 	rTest, rTestCleanup := createRun(t, client, wkspaceTest)
 	defer rTestCleanup()
 
-	t.Run("without include param", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


task_stages_integration_test.go
```diff
@@ -34,7 +34,7 @@ func TestTaskStagesRead(t *testing.T) {
 
 	t.Run("without include param", func(t *testing.T) {
```
brandonc, line 35: This says the test does not have the include param

sebasslash: Yeah I probably should rename; it uses an include param to fetch the task stages of a run so that I can then call the endpoint to fetch a task stage by ID (where the include param is not included).

uturunku1:
Aside from Brandon's request for change, this PR looks good to me.

uturunku1 (approved):
🏁🐎

PR #331, opened 2022-02-24
create consistency around godoc parsing

uturunku1:
## Description

The goal for this PR is to get the godocs ready for 1.0

There's a bit of inconsistency with the docs, where certain patterns are used for some resources and others deviate from the pattern. To mention a few:

1. Linking the API docs for the resource when documenting the resource interface
2. Denoting which fields are required/optional in create, list, update option structs. 
3. Some option structs include docs for each field, others do not
4. A couple of interfaces have the wrong link to API docs
5. GoDocs is randomly parsing some of our variables and constants: https://pkg.go.dev/github.com/hashicorp/go-tfe#pkg-constants As you see is only showing the ones defined in organization_membership.go and notification_configuration.go because they just happen to have a comment on top which is what godoc uses to parse and display.

Implementation and rules:

1. All interfaces must have a comment with a link to the API doc related to it.
2. All methods for this interface must have a explanation on top of the definition of the method
3. Helper functions must not have explanation on top of its definition
4. Variables and functions for internal usage (like the ones on tfe.go) do not need to be documented in GoDoc. This can be avoided by:

- Having the variables/constants lowercase or prefix with underscore. GoDoc only displays the ones that are uppercase.
- Adding a space in between the declaration and the comment. GoDoc does not parse comments that are not adjacent to a top-level declaration and will be omitted from godoc’s output

More rules around GoDoc can be found here: https://go.dev/blog/godoc
And their source code can be found here: https://github.com/golang/tools/tree/master/godoc
You can also see how their own source code gets parsed here https://pkg.go.dev/golang.org/x/tools@v0.1.9/godoc

6. Constants that are typed such as 

```
const (
	Advisory  TaskEnforcementLevel = "advisory"
	Mandatory TaskEnforcementLevel = "mandatory"
)
```

must be sitting next to the declaration of the type like this:

```
//TaskEnforcementLevel is an enum that describes the enforcement levels for a run task
type TaskEnforcementLevel string

const (
	Advisory  TaskEnforcementLevel = "advisory"
	Mandatory TaskEnforcementLevel = "mandatory"
)
```

That way GoDoc can document the type and the constants available to this type as a single group.  If they are not sitting next to each other, for example in this case:

```
//TaskEnforcementLevel is an enum that describes the enforcement levels for a run task
type TaskEnforcementLevel string

const (
	TaskPassed      TaskResultStatus = "passed"
	TaskFailed      TaskResultStatus = "failed"

	Advisory  TaskEnforcementLevel = "advisory"
	Mandatory TaskEnforcementLevel = "mandatory"
)
```
Then GoDoc will not make the connection between those two and will documented them separately.

## Testing plan

These changes were tested by installing godoc locally and running `godoc -http=:6060`

```

...
```

### DIFF
--- a/admin_organization.go
+++ b/admin_organization.go
@@ -72,7 +72,7 @@ type AdminOrganizationList struct {
 	Items []*AdminOrganization
 }
 
-// A list of relations to include. See available resources
+// AdminOrgIncludeOps represents the available options for include query params.
 // https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
 type AdminOrgIncludeOps string
 
@@ -84,10 +84,11 @@ const (
 type AdminOrganizationListOptions struct {
 	ListOptions
 
-	// A query string used to filter organizations.
+	// Optional: A query string used to filter organizations.
 	// Any organizations with a name or notification email partially matching this value will be returned.
 	Query string `url:"q,omitempty"`
-
+	// Optional: A list of relations to include. See available resources
+	// https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
 	Include []AdminOrgIncludeOps `url:"include,omitempty"`
 }
 
@@ -117,6 +118,7 @@ func (s *adminOrganizations) List(ctx context.Context, options *AdminOrganizatio
 	return orgl, nil
 }
 
+// ListModuleConsumers lists specific organizations in the Terraform Enterprise installation that have permission to use an organization's modules.
 func (s *adminOrganizations) ListModuleConsumers(ctx context.Context, organization string, options *AdminOrganizationListModuleConsumersOptions) (*AdminOrganizationList, error) {
 	if !validStringID(&organization) {
 		return nil, ErrInvalidOrg
@@ -138,6 +140,7 @@ func (s *adminOrganizations) ListModuleConsumers(ctx context.Context, organizati
 	return orgl, nil
 }
 
+// Read an organization by its name.
 func (s *adminOrganizations) Read(ctx context.Context, organization string) (*AdminOrganization, error) {
 	if !validStringID(&organization) {
 		return nil, ErrInvalidOrg
@@ -158,6 +161,7 @@ func (s *adminOrganizations) Read(ctx context.Context, organization string) (*Ad
 	return org, nil
 }
 
+// Update an organization by its name.
 func (s *adminOrganizations) Update(ctx context.Context, organization string, options AdminOrganizationUpdateOptions) (*AdminOrganization, error) {
 	if !validStringID(&organization) {
 		return nil, ErrInvalidOrg
@@ -178,6 +182,7 @@ func (s *adminOrganizations) Update(ctx context.Context, organization string, op
 	return org, nil
 }
 
+// UpdateModuleConsumers updates an organization to specify a list of organizations that can use modules from the sharing organization's private registry.
 func (s *adminOrganizations) UpdateModuleConsumers(ctx context.Context, organization string, consumerOrganizationIDs []string) error {
 	if !validStringID(&organization) {
 		return ErrInvalidOrg
--- a/admin_setting_twilio.go
+++ b/admin_setting_twilio.go
@@ -7,7 +7,8 @@ import (
 // Compile-time proof of interface implementation.
 var _ TwilioSettings = (*adminTwilioSettings)(nil)
 
-// TwilioSettings describes all the Twilio admin settings.
+// TwilioSettings describes all the Twilio admin settings for the Admin Setting API.
+// https://www.terraform.io/cloud-docs/api-docs/admin/settings.
 type TwilioSettings interface {
 	// Read returns the Twilio settings.
 	Read(ctx context.Context) (*AdminTwilioSetting, error)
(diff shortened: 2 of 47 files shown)


task_stages_integration_test.go
```diff
@@ -32,9 +32,9 @@ func TestTaskStagesRead(t *testing.T) {
 	rTest, rTestCleanup := createRun(t, client, wkspaceTest)
 	defer rTestCleanup()
 
-	t.Run("without include param", func(t *testing.T) {
+	t.Run("with include param to fetch the task stage of a run", func(t *testing.T) {
 		r, err := client.Runs.ReadWithOptions(ctx, rTest.ID, &RunReadOptions{
-			Include: "task_stages",
+			Include: []RunIncludeOps{RunTaskStages},
```
uturunku1, line 37: These changes in this file are not relevant to this PR's purposes. But this adjustment was needed to avoid a build error pulled from releases branch

helper_test.go
```diff
@@ -164,7 +164,7 @@ func createNotificationConfiguration(t *testing.T, client *Client, w *Workspace,
 			Name:            String(randomString(t)),
 			Token:           String(randomString(t)),
 			URL:             String("http://example.com"),
-			Triggers:        []string{NotificationTriggerCreated},
```
uturunku1, line 167: I think someone at some point meant to create a typed string for the constants they created but they forgot to create the typed string and left it as a string.

team_access.go
```diff
@@ -39,42 +39,46 @@ type teamAccesses struct {
...
-type VariablesPermissionType string
-
-// StateVersionsPermissionType represents the permissiontype to a workspace's state versions.
-type StateVersionsPermissionType string
-
-// SentinelMocksPermissionType represents the permissiontype to a workspace's Sentinel mocks.
-type SentinelMocksPermissionType string
-
-// WorkspaceLockingPermissionType represents the permissiontype to lock or unlock a workspace.
-type WorkspaceLockingPermissionType bool
```
uturunku1, line 55: I deleted this type. It was not being used anywhere

plan.go
```diff
@@ -15,7 +15,7 @@ var _ Plans = (*plans)(nil)
 // Plans describes all the plan related methods that the Terraform Enterprise
 // API supports.
 //
-// TFE API docs: https://www.terraform.io/docs/cloud/api/plan.html
+// TFE API docs: https://www.terraform.io/cloud-docs/api-docs/plans
```
uturunku1, line 18: FYI, this is not suppose to be `plans.html` . This would take you to a "page not found"

admin_organization.go
```diff
@@ -84,10 +84,10 @@ const (
 type AdminOrganizationListOptions struct {
 	ListOptions
 
-	// A query string used to filter organizations.
+	// Optional: A query string used to filter organizations.
 	// Any organizations with a name or notification email partially matching this value will be returned.
 	Query string `url:"q,omitempty"`
-
+	// Optional:
```
brandonc, line 90: This is an interesting and useful parameter that I think we should take better care to document explicitly. Here's an example from WorkspaceReadOptions:

`// Optional: A list of relations to include. See available resources https://www.terraform.io/docs/cloud/api/workspaces.html#available-related-resources`

I think we have good docs around which relations are possible to include, and a link would reinforce the relationship we now have to maintain because they are typed strings.

admin_organization.go
```diff
@@ -117,6 +117,7 @@ func (s *adminOrganizations) List(ctx context.Context, options *AdminOrganizatio
 	return orgl, nil
 }
 
+// List specific organizations in the Terraform Enterprise installation that have permission to use an organization's modules.
```
brandonc, line 120: I think the [godoc convention](https://go.dev/blog/godoc) is to begin the comment with the element it describes:

"ListModuleConsumers lists specific organizations..."

brandonc (commented):
👇

admin_organization.go
sebasslash: ☝️

admin_organization.go
sebasslash: 👆 I noticed there were several comments for functions that don't follow the pattern: `// <FUNCTION> does X and Y` -- something to keep an eye out for.

admin_organization.go
uturunku1: We have a similar comment for the include options in admin_organization.go. You may have not seen it because is in line 75:

```
// A list of relations to include. See available resources
// https://www.terraform.io/docs/cloud/api/admin/organizations.html#available-related-resources
type AdminOrgIncludeOps string
```

So, the issue here is that we have discrepancy where we place that comment and link. Most of the time we do it on top of the type and occasionally we do it the include field definition level, like what you see in workspace.go. I personally, like it a bit more on top of the type because Godoc makes types more visible and will include the typed constants for the include field.

admin_organization.go
uturunku1: I have not ran into that issue before, but I will keep an eye on that.

admin_setting_twilio.go
```diff
@@ -7,7 +7,7 @@ import (
 // Compile-time proof of interface implementation.
 var _ TwilioSettings = (*adminTwilioSettings)(nil)
 
-// TwilioSettings describes all the Twilio admin settings.
+// TwilioSettings describes all the Twilio admin settings for the Admin Setting API https://www.terraform.io/cloud-docs/api-docs/admin/settings.
```
sebasslash, line 10: So I'm curious if we should move the link to a new line for more readability.

uturunku1: Will do.

brandonc (approved):

PR #335, opened 2022-02-28
check if value exist first

uturunku1:
## Description

Attempting to access the value of `*o.Auth`, when this field has no value, will give us the invalid memory address error
And users are allowed to not defined this value because it comes with a default value: https://www.terraform.io/cloud-docs/api-docs/admin/settings#request-body-3

```

...
```

### DIFF
--- a/admin_setting_smtp.go
+++ b/admin_setting_smtp.go
@@ -97,9 +97,12 @@ func (a *adminSMTPSettings) Update(ctx context.Context, options AdminSMTPSetting
 }
 
 func (o AdminSMTPSettingsUpdateOptions) valid() error {
-	_, isValidType := validSMTPAuthType[*o.Auth]
-	if !isValidType {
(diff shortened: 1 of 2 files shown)


brandonc (approved):
Nice! Was it just something you noticed?

uturunku1:
> Nice! Was it just something you noticed?

Not really. The error just pop up in my head by association. 
I was adding an `include` field to listOptions in run_trigger.go and making changes around the `valid()` helper function. 
The value passed is not part of a map then the option passed is invalid, right? This implementation works fine here because if the value is not defined at all,  `_, isValidRunTriggerType := validRunTriggerType[o.RunTriggerType]` this line will just return false, not panic (o.RunTriggerType it's just an empty string). 
Suddenly that reminded me that o.Auth field is not just a string but a pointer to a string, and there you go.

PR #336, opened 2022-02-28
move release instructions to docs directory

uturunku1:
## Description

Recently we created a new docs directory. The instructions for release process, that were placed inside README.md, its long enough that it can live in its own md file inside docs directory.

```

...
```


sebasslash (approved):
🔥

PR #333, opened 2022-02-24
1.x changelog

sebasslash:
The 1.0 changelog -- WIP

Not sure if I properly organized the sections, but the log has all the changes thus far. Let me know if I missed any.

PR #337, opened 2022-02-28
Refactor nested IDs into structs

sebasslash:
## Description
There are a few endpoints which use nested identifiers to identify a resource rather than using the external ID of that resource. As of now the Module Registry API seems to follow this type of pattern, e.g 

`POST /organizations/:organization_name/registry-modules/:registry_name/:namespace/:name/:provider/versions`

In order to keep method signatures readable, I've refactored these nested URL params into structs which represent the set of identifiers used when referencing a resource. The example above would look something like:

```go
type RegistryModuleID struct {
  Organization string
  RegistryName string
  Namespace string
  Name string
  Provider string
}
```
Not only does this help keep our method signatures compact, but it enables us to keep our validations DRY for more consistency.


brandonc (approved):
I like it. Let's update https://github.com/hashicorp/go-tfe/pull/333, too.

PR #339, opened 2022-03-01
add missing include field for listOptions

uturunku1:
## Description

Add "include" field to listOptions structs for API that allow for include values.

**These are the APIs and their include values listed in Atlas:**

agent pool API: "workspaces" 
oauth client API: "oauth_tokens" 
policy check API: "run.workspace", "run" 
run trigger API: "workspace", "sourceable"
organization API: "entitlement_set" 
policy set versions API: "policy_set"
policy API: "policy_sets"  

**I did not implemented the changes for:**
 
organizations API - because the relation for entitlement set was not part of `type Organization struct`. 
policy set version API - I did have the relation present but I did not have the listOptions or listing function present.
Policy API - Policies controller has the include value "policy_sets", but this value does not align with only relation listed under `type Policy struct`: `Organization *Organization `jsonapi:"relation,organization"`` What am I missing here?

**All the other include fields were added. This is how the implementation was made:** 

Step 1 - Add `include` struct field to SomeResourceListOptions
Step 2 - Create constants for the allowed include values that are typed strings
Step 3 - Add a test to make sure what are getting the "included" data

**Bonus Step:**

Change the pre-existing typed string named `SomePrefixIncludeOps` to `SomePrefixIncludeOp`, like `type OAuthClientIncludeOps string` becomes `type OAuthClientIncludeOp string`. I should have made the suffix singular from the beginning, but it is not too late to make that correction. 
To reduce noise coming from this change, you can focus on the changes made to:

1. agent_pool.go
2. agent_pool_integration_test.go
3. oauth_client.go
4. oauth_client_integration_test.go
5. policy_check.go
6. policy_check_integration_test.go
7. run_trigger.go
8. run_trigger_integration_test.go
9. errors.go

```

...
```

### DIFF
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -82,10 +82,17 @@ type OAuthClient struct {
 	OAuthTokens  []*OAuthToken `jsonapi:"relation,oauth-tokens"`
 }
 
+// A list of relations to include
+type OAuthClientIncludeOpt string
+
+const OauthClientOauthTokens OAuthClientIncludeOpt = "oauth_tokens"
+
 // OAuthClientListOptions represents the options for listing
 // OAuth clients.
 type OAuthClientListOptions struct {
 	ListOptions
+
+	Include []OAuthClientIncludeOpt `url:"include,omitempty"`
 }
 
 // OAuthClientCreateOptions represents the options for creating an OAuth client.
--- a/run_task.go
+++ b/run_task.go
@@ -59,13 +59,13 @@ type RunTaskList struct {
 	Items []*RunTask
 }
 
-// RunTaskIncludeOps represents the available options for include query params.
+// RunTaskIncludeOpt represents the available options for include query params.
 // https://www.terraform.io/cloud-docs/api-docs/run-tasks#list-run-tasks
-type RunTaskIncludeOps string
+type RunTaskIncludeOpt string
 
 const (
-	RunTaskWorkspaceTasks RunTaskIncludeOps = "workspace_tasks"
-	RunTaskWorkspace      RunTaskIncludeOps = "workspace_tasks.workspace"
+	RunTaskWorkspaceTasks RunTaskIncludeOpt = "workspace_tasks"
+	RunTaskWorkspace      RunTaskIncludeOpt = "workspace_tasks.workspace"
 )
 
 // RunTaskListOptions represents the set of options for listing run tasks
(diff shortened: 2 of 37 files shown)


oauth_client.go
```diff
@@ -82,10 +82,17 @@ type OAuthClient struct {
 	OAuthTokens  []*OAuthToken `jsonapi:"relation,oauth-tokens"`
 }
 
+// A list of relations to include
+type OAuthClientIncludeOp string
+
+const OauthClientOauthTokens OAuthClientIncludeOp = "oauth_tokens"
```
uturunku1, line 88: Besides "OAuthToken" another relation listed for `type OAuthClient struct` is "organization". However,  I did not included a constant for "organization" because I did not see it listed in atlas's `allowed_values_for_include`

sebasslash:
I have two comments:
1. I have a hard time distinguishing what `IncludeOp` means... Include Operation? Include Option? I've typically seen `Op` for operation and `Opt` for option. 
2. A lot of these strings are not documented and thus will not show up on GoDocs (or am I wrong?)... so while we've prevented a customer from sending an invalid include value, the customer would still need to reference the API docs AND the code to see what include options are available. Should we make this easier for them?

uturunku1:
> I have two comments:
> 
> 1. I have a hard time distinguishing what `IncludeOp` means... Include Operation? Include Option? I've typically seen `Op` for operation and `Opt` for option.
> 2. A lot of these strings are not documented and thus will not show up on GoDocs (or am I wrong?)... so while we've prevented a customer from sending an invalid include value, the customer would still need to reference the API docs AND the code to see what include options are available. Should we make this easier for them?

1. You are right that many people may interpret `Op` for operations. I'll switch it to `IncludeOpt`
2. They are not documented because unfortunately API docs does not have any documentation on it :(

run_task.go
```diff
@@ -59,28 +59,28 @@ type RunTaskList struct {
...
-	Include []RunTaskIncludeOps `url:"include,omitempty"`
+	Include []RunTaskIncludeOpt `url:"include,omitempty"`
 }
 
 // RunTaskReadOptions represents the set of options for reading a run task
 type RunTaskReadOptions struct {
 	// Optional: A list of relations to include with a run task. See available resources:
 	// https://www.terraform.io/cloud-docs/api-docs/run-tasks#list-run-tasks
-	Include []RunTaskIncludeOps `url:"include"`
+	Include []RunTaskIncludeOpt `url:"include"`
```
sebasslash, line 83: Ooo a missing `omitempty` ? 🤔

sebasslash (changes requested):
Last teeny bit, I'm sorry 😢

run_task.go
uturunku1: lol, I don't apply the practices I preach. Making that change now!

PR #341, opened 2022-03-02
Update validations

uturunku1:
## Description

This PR introduces validation check for all "include" values. Previously we had 3 files that had helper functions to check that the values passed as "filter" option and "include" option were valid strings. This release seemed like a good opportunity to standardize this practice beyond those 3 files.

Also, as we have recently made all "include" options a typed string field, it makes validation of those values more convenient.

Keep in mind that throwing errors for invalid include params is a nice bonus that we are adding to go-tfe and not a required one because this behavior does not exactly align with the response you'd get from the platform. To understand this point with a better context, read the comment I left for one of those validations https://github.com/hashicorp/go-tfe/blob/76df95cb33cf241f5d301849ed53f321b5786d20/admin_run.go#L134-L147

## Implementation Details

I added

```
		return nil, err
	}
```
to listing and reading functions that take in options with Include values.

Within the valid() func, I call a helper function that validates that those include values are the listed as constants IncludeOpt typed string

```
if err := validateAdminOrgIncludeParams(o.Include); err != nil {
		return err
	}
```

I added some tests to validate this new functionality.

## Important note

A change not directly related to this PR, but that the new functionality helped surface as an issue, is that a previous change I had introduced on PR #339 was invalid. So I reverted that change in this PR. You can find that change in the following files:

agent_pool.go
agent_pool_integration_test.go

### DIFF
--- a/admin_run.go
+++ b/admin_run.go
@@ -116,37 +116,65 @@ func (s *adminRuns) ForceCancel(ctx context.Context, runID string, options Admin
 }
 
 func (o *AdminRunsListOptions) valid() error {
-	if o == nil { // no need to validate fields
+	if o == nil { // nothing to validate
 		return nil
 	}
-	if validString(&o.RunStatus) {
-		validRunStatus := map[string]int{
-			string(RunApplied):            1,
-			string(RunApplyQueued):        1,
-			string(RunApplying):           1,
-			string(RunCanceled):           1,
-			string(RunConfirmed):          1,
-			string(RunCostEstimated):      1,
-			string(RunCostEstimating):     1,
-			string(RunDiscarded):          1,
-			string(RunErrored):            1,
-			string(RunPending):            1,
-			string(RunPlanQueued):         1,
-			string(RunPlanned):            1,
-			string(RunPlannedAndFinished): 1,
-			string(RunPlanning):           1,
-			string(RunPolicyChecked):      1,
-			string(RunPolicyChecking):     1,
-			string(RunPolicyOverride):     1,
-			string(RunPolicySoftFailed):   1,
-		}
-		runStatus := strings.Split(o.RunStatus, ",")
 
+	if err := validateAdminRunFilterParams(o.RunStatus); err != nil {
+		return err
+	}
+
+	if err := validateAdminRunIncludeParams(o.Include); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateAdminRunFilterParams(runStatus string) error {
+	// For the platform, an invalid filter value is a semantically understood query that returns an empty set, no error, no warning. But for go-tfe, an invalid value is good enough reason to error prior to a network call to the platform:
+	if validString(&runStatus) {
+		sanitizedRunstatus := strings.TrimSpace(runStatus)
+		runStatuses := strings.Split(sanitizedRunstatus, ",")
 		// iterate over our statuses, and ensure it is valid.
-		for _, status := range runStatus {
-			if _, present := validRunStatus[status]; !present {
-				return fmt.Errorf("invalid value %s for run status", status)
+		for _, status := range runStatuses {
+			switch status {
+			case string(RunApplied),
+				string(RunApplyQueued),
+				string(RunApplying),
+				string(RunCanceled),
+				string(RunConfirmed),
+				string(RunCostEstimate),
+				string(RunCostEstimating),
+				string(RunDiscarded),
+				string(RunErrored),
+				string(RunPending),
+				string(RunPlanQueued),
+				string(RunPlanned),
+				string(RunPlannedAndFinished),
+				string(RunPlanning),
+				string(RunPolicyChecked),
+				string(RunPolicyChecking),
+				string(RunPolicyOverride),
+				string(RunPolicySoftFailed),
+				"":
+				// do nothing
+			default:
+				return fmt.Errorf(`invalid value "%s" for run status`, status)
 			}
+		}
+	}
+
+	return nil
+}
+
+func validateAdminRunIncludeParams(params []AdminRunIncludeOpt) error {
+	for _, p := range params {
+		switch p {
+		case AdminRunWorkspace, AdminRunWorkspaceOrg, AdminRunWorkspaceOrgOwners:
+			// do nothing
+		default:
+			return ErrInvalidIncludeValue
 		}
 	}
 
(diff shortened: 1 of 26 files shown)


admin_run.go
```diff
@@ -116,39 +116,48 @@ func (s *adminRuns) ForceCancel(ctx context.Context, runID string, options Admin
...
+	}
+
+	if err := validateAdminRunIncludeParams(o.Include); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateAdminRunFilterParams(runStatus string) error {
```
brandonc, line 134: Is it not possible to define RunStatus as type RunStatus on AdminRunsListOptions so you can avoid the string type here?

agent_pool.go
```diff
@@ -48,18 +51,21 @@ type AgentPool struct {
...
 
+type AgentPoolReadOptions struct {
+	Include []AgentPoolIncludeOpt `url:"include,omitempty"`
+}
+
 // AgentPoolListOptions represents the options for listing agent pools.
 type AgentPoolListOptions struct {
 	ListOptions
-
-	Include []AgentPoolIncludeOpt `url:"include,omitempty"`
```
brandonc, line 62: I thought this was possible on the List as well. Why is it removed?

uturunku1: It is not possible. I was surprised myself. But my digging how agent pool is implemented in Atlas took me to that conclusion.

admin_run.go
uturunku1: Indeed! I'll make it into a typed string.

agent_pool.go
brandonc:
```
$ curl -s -H "Authorization: Bearer $(tfc_local_token)" https://tfcdev-5839ba17.ngrok.io/api/v2/organizations/hashicorp/agent-pools?include=workspaces | jq .included[0].attributes.name
"example"
```

admin_run.go
uturunku1: On second thought, converting RunStatus filter option from a string into a typed string lis going to a larger deal than expected. Filter options is a comma-separated list https://www.terraform.io/cloud-docs/api-docs/admin/runs#query-parameters just like Include options. So I’d have to apply the same types of changes I made for include field: make it into a slice of typed strings, then convert it back into string of comma-separated values when I call client.newRequest . Then I would have to apply the same changes to all filter options not just RunStatus to create consistency in this behavior for all filter options :(

agent_pool.go
uturunku1: ooo OK! Lemme try that as integration test and see what results I get

admin_run.go
```diff
@@ -116,39 +116,48 @@ func (s *adminRuns) ForceCancel(ctx context.Context, runID string, options Admin
...
+		runStatuses := strings.Split(runStatus, ",")
 		// iterate over our statuses, and ensure it is valid.
-		for _, status := range runStatus {
-			if _, present := validRunStatus[status]; !present {
-				return fmt.Errorf("invalid value %s for run status", status)
+		for _, status := range runStatuses {
+			switch status {
+			case string(RunApplied), string(RunApplyQueued), string(RunApplying), string(RunCanceled), string(RunConfirmed), string(RunCostEstimate), string(RunCostEstimating), string(RunDiscarded), string(RunErrored), string(RunPending), string(RunPlanQueued), string(RunPlanned), string(RunPlannedAndFinished), string(RunPlanning), string(RunPolicyChecked), string(RunPolicyChecking), string(RunPolicyOverride), string(RunPolicySoftFailed):
+				// do nothing
+			default:
```
brandonc, line 143: I wonder if we should allow an empty string because one could be split if there was a trailing comma. The API would accept a trailing comma, I think.

brandonc (commented):
Looks good, but I wonder if we are being too strict and not creating a situation that can't be worked around if the client lags behind the API in some of these cases

admin_run.go
uturunku1: Good call! It could also have trailing space. OK! I am writing a new test for this.

admin_run.go
```diff
@@ -116,39 +116,48 @@ func (s *adminRuns) ForceCancel(ctx context.Context, runID string, options Admin
...
+	// For the platform, an invalid filter value is a semantically understood query that returns an empty set, no error, no warning. But for go-tfe, an invalid value is good enough reason to error prior to a network call to the platform:
+	if validString(&runStatus) {
+		runStatuses := strings.Split(runStatus, ",")
 		// iterate over our statuses, and ensure it is valid.
-		for _, status := range runStatus {
-			if _, present := validRunStatus[status]; !present {
-				return fmt.Errorf("invalid value %s for run status", status)
+		for _, status := range runStatuses {
+			switch status {
+			case string(RunApplied), string(RunApplyQueued), string(RunApplying), string(RunCanceled), string(RunConfirmed), string(RunCostEstimate), string(RunCostEstimating), string(RunDiscarded), string(RunErrored), string(RunPending), string(RunPlanQueued), string(RunPlanned), string(RunPlannedAndFinished), string(RunPlanning), string(RunPolicyChecked), string(RunPolicyChecking), string(RunPolicyOverride), string(RunPolicySoftFailed):
```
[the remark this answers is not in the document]
uturunku1: Yes

admin_run_integration_test.go
```diff
@@ -244,6 +252,15 @@ func TestAdminRuns_AdminRunsListOptions_valid(t *testing.T) {
 		err := opts.valid()
 		assert.Error(t, err)
 	})
+
+	t.Run("has trailing comma and trailing space", func(t *testing.T) {
```
uturunku1, line 256: This test and all the ones for `TestAdminRuns_AdminRunsListOptions_valid` don't really need to be part of integration test suite. We are just testing helper functions and it'd be nice if external contributors could at least have access to running these type of tests. I'd create a unit test file, like I did with `tfe_test.go`, but then suddenly we end up with 2 new unit test files, suddenly it feels like they _should_ have their own directory to live..., suddenly we have the urge to reorganize other files. But this is not the time to create new files or reorganize files.
The subject of reorganizing this project for this upcoming has been discussed quite a bit and we decided to postpone. It would involve too many breaking changes on top of the ones we are already bringing. But I really hope that we can find a solution to this organization/creation of new files for this project. It is not scalable as it is :(

brandonc (approved):

PR #347, opened 2022-03-04
Fixed URLto docs in plan.go

atthematyo:
https://www.terraform.io/docs/cloud/api/plan.html leads to a page not found. Fixed so it opens up the correct doc.

## Description

Simple fix to comment so that it has the correct URL to the appropriate doc.

```

...
```


sebasslash (approved):
Thanks for this fix 🔥

sebasslash:
Closing as the link has already been fixed (I apologize).

PR #351, opened 2022-03-04
avoid failing on wrapped errors

uturunku1:
## Description

I noticed logreader.go was using the old way of comparing errors `==`. This make sense because this file was implemented almost 4 years ago when `errors.Is` was not an available standardize practice yet: https://go.dev/blog/go1.13-errors

This could potentially fail an assertion if the error thrown by that external package had wrapped errors. So I change this implementation to use `errors.Is`.

The other small change I am introducing here is that instead of returning a semi-vague error thrown by an external library like so:

```
file, err := os.Stat(path)
	if err != nil {
            return err
```

I am unwrapping the error, adding additional information about the call site where things failed, and then wrapping back the error with `fmt.Errorf(...)`, like this:

```
if err != nil {
		if errors.Is(err, fs.ErrNotExist) {
			return fmt.Errorf(`failed to find terraform configuration files under the path "%v": %w`, path, err)
		}
		return fmt.Errorf(`unable to upload terraform configuration files from the path "%v": %w`, path, err)
	}
``` 

Adding some context to an error coming from a large external package, should make identifying the source of problems easier during debugging.

```

...
```

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -1,6 +1,8 @@
 package tfe
 
 import (
+	"errors"
+	"io/fs"
 	"log"
 	"sort"
 
@@ -884,8 +886,12 @@ func packContents(path string) (*bytes.Buffer, error) {
 
 	file, err := os.Stat(path)
 	if err != nil {
-		return body, err
+		if errors.Is(err, fs.ErrNotExist) {
+			return body, fmt.Errorf(`failed to find files under the path "%v": %w`, path, err)
+		}
+		return body, fmt.Errorf(`unable to upload files from the path "%v": %w`, path, err)
 	}
+
 	if !file.Mode().IsDir() {
 		return body, ErrMissingDirectory
 	}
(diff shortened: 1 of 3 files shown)


sebasslash (approved):
👍

sebasslash:
Disregard, I didn't notice it was a draft.

tfe.go
```diff
@@ -870,8 +872,8 @@ func decodeErrorPayload(r *http.Response) ([]string, error) {
 	return errs, nil
 }
 
-func errorPayloadContains(errors []string, match string) bool {
-	for _, e := range errors {
+func errorPayloadContains(payloadErrors []string, match string) bool {
```
uturunku1, line 875: unrelated change to this PR, fyi. Golangci-lint warned me that the previous name assigned to this variable was clashing with the name of the external library "errors".

sebasslash (approved):
👍

sebasslash:
I wasn't aware of this. Read up on `errors.Is()`, very cool -- I'd been comparing errors using `==` all this time. I'm wondering if the tfe provider might need this change. 🤔

PR #354, opened 2022-03-09
Add AddComment on Run struct

alex-ikse:
## Description

Add the ability to use the endpoint `POST /runs/:run_id/comments` to add comments on a run.

## Testing plan

```
runComment := tfe.RunCommentOptions{
    Body: tfe.String("Test from fork of go-tfe"),
}
err = client.Runs.AddComment(ctx, run.ID, runComment)
if err != nil {
    log.Fatal(err)
}
```

## External links

This endpoint is not documented right now.

```

...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: test
 on:
-  pull_request:
+  push:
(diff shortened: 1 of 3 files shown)


alex-ikse:
Replaced by #355

PR #355, opened 2022-03-09
Add comments object

alex-ikse:
## Description

Adding "comments" object/endpoint to create, list and read run comments

## Testing plan

```go
// Create a context
ctx := context.Background()

run, err := client.Runs.Read(ctx, "run-xxxxxxxxxxxx")
if err != nil {
	log.Fatal(err)
}

commentCreateOptions := tfe.CommentCreateOptions{
	Body: tfe.String("Test from fork of go-tfe v3"),
}
comm, err := client.Comments.Create(ctx, run.ID, commentCreateOptions)
if err != nil {
	log.Fatal(err)
}

fmt.Printf("Comm ID created: %s\n", comm.ID)

commRead, err := client.Comments.Read(ctx, comm.ID)
if err != nil {
	log.Fatal(err)
}
fmt.Printf("Body: %s\n", commRead.Body)

```

## External links

Documentation about Comment doesn't exist right now

```

...
```

### DIFF
--- a/comment.go
+++ b/comment.go
@@ -1,0 +1,140 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+// Compile-time proof of interface implementation.
+var _ Comments = (*comments)(nil)
+
+// Comments describes all the comment related methods that the
+// Terraform Enterprise API supports.
+//
+// TFE API docs:
+// https://www.terraform.io/docs/cloud/api/comments.html
+type Comments interface {
+	// List all comments of the given run.
+	List(ctx context.Context, runID string, options *CommentListOptions) (*CommentList, error)
+
+	// Read a comment by its ID.
+	Read(ctx context.Context, CommentID string) (*Comment, error)
+
+	// Create a new comment with the given options.
+	Create(ctx context.Context, runID string, options CommentCreateOptions) (*Comment, error)
+}
+
+// Comments implements Comments.
+type comments struct {
+	client *Client
+}
+
+// CommentList represents a list of comments.
+type CommentList struct {
+	*Pagination
+	Items []*Comment
+}
+
+// Comment represents a Terraform Enterprise comment..
+type Comment struct {
+	ID   string `jsonapi:"primary,comments"`
+	Body string `jsonapi:"attr,body"`
+}
+
+// CommentListOptions represents the options for listing comments.
+type CommentListOptions struct {
+	ListOptions
+}
+
+type CommentCreateOptions struct {
+	// Type is a public field utilized by JSON:API to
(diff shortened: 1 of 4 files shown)


comment.go
```diff
@@ -0,0 +1,140 @@
...
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,comments"`
+
+	// Required: Body of the comment.
+	Body *string `jsonapi:"attr,body"`
+
+	// Optional: Run where the comment is attached
+	Run *Run `jsonapi:"relation,run"`
```
sebasslash, line 56: Not necessary to include in this struct, the API figures out the parent run through the `:id` URL param.

comment.go
```diff
@@ -0,0 +1,140 @@
...
+	return cl, nil
+}
+
+// Create a new comment with the given options.
+func (s *comments) Create(ctx context.Context, runID string, options CommentCreateOptions) (*Comment, error) {
+	if !validStringID(&runID) {
+		return nil, ErrInvalidRunID
+	}
+
+	if !validString(options.Body) {
```
sebasslash, line 86: Let's refactor these validations to a `func (o CommentCreateOptions) valid() error {}` at the bottom of the file. Then in the `Create()` body:
```go
  return nil, err
}
```

comment.go
```diff
@@ -0,0 +1,124 @@
...
+
+type CommentCreateOptions struct {
+	// Type is a public field utilized by JSON:API to
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,comments"`
+
+	// Required: Body of the comment.
+	Body *string `jsonapi:"attr,body"`
```
sebasslash, line 53: Should be type `string`

sebasslash (changes requested):
Overall, looks great 👍 There's a few things needing to be changed due to some patterns we've established for the upcoming 1.0 release.

comment.go
alex-ikse:
```
Run *Run `jsonapi:"relation,run"
``` 
removed

comment.go
alex-ikse: Changed to `string`

comment.go
alex-ikse: refactor as `options.valid()`

sebasslash (approved):
The Force is ssstrong with this one 🔥

sebasslash (approved):
Only the Sith deal in branch protections!

PR #363, opened 2022-03-18
remove branch from github actions

uturunku1:
We are done using the releases 1.0 branch for now. The branch itself has not been deleted because is possible we want to go back to it in the future. But for now I am removing this branch from our PR checks (specifically the github actions one) before I forget that we need to remove it.
Lots of tests failing right now: Looks like the nightly build for the platform is running.

```

...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: test
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


sebasslash (approved):
Sad to see it go 😢

sebasslash (approved):
👍

PR #371, opened 2022-03-31
update contributing doc

uturunku1:

```

...
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -2,6 +2,14 @@
 
 If you find an issue with this package, please create an issue in GitHub. If you'd like, we welcome any contributions. Fork this repository and submit a pull request.
 
+## Adding new functionality or fixing relevant bugs
+
+If you are adding a new endpoint, make sure to update COVERAGE.md where we keep a list of the TFC APIs that this SDK supports.
+
+If you are making relevant changes that is worth communicating to our users, please include a note about it in our CHANGELOG.md. You can include it as part of the PR where you are submitting your changes.
+
+CHANGELOG.md should have the next minor version listed as `# v1.X.0 (Unreleased)` and any changes can go under there. But if you feel that your changes are better suited for a patch version (like a critical bug fix), you may list a new section for this version. You should repeat the same formatting style introduced by previous versions.
+
 ## Running the Linters Locally
 
 1. Ensure you have have [installed golangci-lint](https://golangci-lint.run/usage/install/#local-installation)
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -23,7 +29,31 @@ You'll need to generate mocks if an existing endpoint method is modified or a ne
...
+  skipIfBeta(t)
+  options := WorkspaceCreateOptions {
+     // rest of required fields here
+     ExampleNewField: Bool(true),
+   }
+  // the rest of your test logic here
+})
+```
+
+Once your PR has been merged you can later come back and remove the skipIfBeta() flag.
```
sebasslash, line 54: I would prefix this with `**Note**:` to make it more prominent.

docs/CONTRIBUTING.md
```diff
@@ -2,6 +2,12 @@
 
 If you find an issue with this package, please create an issue in GitHub. If you'd like, we welcome any contributions. Fork this repository and submit a pull request.
 
+## Adding new functionality or fixing relevant bugs
+
+If you are making any relevant changes that is worth communicating to our users, please include a note about it in our CHANGELOG.md. You can include it as part of the PR where you are submitting your changes.
```
sebasslash, line 7: I might be more specific here. `CHANGELOG.md` should have the next minor version listed as `# v1.X.0 (Unreleased)` and any changes should go under there. If the next version is not listed, it would be encouraged for the contributor to add it. The one caveat being if it's a critical bug fix that needs to be released immediately, one would need to add the next patch version instead (so it's okay if you have two unreleased versions listed in the changelog). In addition we should provide an example of how our changelog is formatted:
```
# v1.X.X (Unreleased)

## Enhancements
* Brief Enhancement description by @contributer ([#XXX](https://github.com/hashicorp/go-tfe/pull/XXX))

## Bug Fixes
* Brief bug fix description by @contributer ([#XXX](https://github.com/hashicorp/go-tfe/pull/XXX))
```

On a side note, I'd love to create an action that automagically updates `CHANGELOG.md` to list the next release whenever a new release is created.

sebasslash (commented):
Overall, looks 👍 🔥 A few comments below for discussion.

docs/CONTRIBUTING.md
sebasslash: Actually once the PR is merged, we don't want to remove `skipIfBeta()` until the feature either reaches GA or the flag is enabled in CI

uturunku1:
@sebasslash I have updated the PR with your suggestions. The only thing I didn't add was an example of the format the contributor must follow for adding to the changelog. That's because I see most, if not all, contributors that update the changelog, are really good at sticking to the format they see in there, and they just copy the style. So I don't see the need of giving them more additional examples than the ones already introduced by the changelog file itself.

sebasslash (approved):
👍 Awesome.

PR #375, opened 2022-04-07
Feature: add support for applying/removing variable set against workspaces

byronwolfman:
## Description

Hello! The Terraform Cloud API has a couple of Variable Set APIs which our team would like to add for the sake of the [Terraform TFE Provider](https://registry.terraform.io/providers/hashicorp/tfe/latest) which `go-tfe` underpins. These APIs are:

* [Apply Variable Set to Workspaces](https://www.terraform.io/cloud-docs/api-docs/variable-sets#apply-variable-set-to-workspaces)
* [Remove a Variable Set from Workspaces](https://www.terraform.io/cloud-docs/api-docs/variable-sets#remove-a-variable-set-from-workspaces)

Both APIs accept a payload of the following format:

```json
{
  "data": [
    {
      "type": "workspaces",
      "id": "ws-YwfuBJZkdai4xj9w"
    },
    {
      "type": "workspaces",
      "id": "ws-YwfuBJZkdai4xj9w"
    },
    ...
  ]
}

```

## Testing plan

* Added test coverage for new methods
* Wrote some tiny programs to consume the new methods and manually verify functionality

## External links

* [Apply Variable Set to Workspaces](https://www.terraform.io/cloud-docs/api-docs/variable-sets#apply-variable-set-to-workspaces)
* [Remove a Variable Set from Workspaces](https://www.terraform.io/cloud-docs/api-docs/variable-sets#remove-a-variable-set-from-workspaces)

N/A; I have access to some live orgs which I'm sure I should not be messing around with!


variable_set.go
```diff
@@ -235,6 +242,72 @@ func (s *variableSets) Delete(ctx context.Context, variableSetID string) error {
 	return s.client.do(ctx, req, nil)
 }
 
+// VariableSetApplyToWorkspacesOptions represents the options for applying variable sets to workspaces.
+type VariableSetApplyToWorkspacesOptions struct {
```
sebasslash, line 246: We've established a convention for v1.0 that lays out the structure for any given file: 1) Interface/Struct definitions, 2) Receiver methods and 3) validate/helper methods.

variable_set_test.go
```diff
@@ -190,6 +190,67 @@ func TestVariableSetsDelete(t *testing.T) {
...
 }
 
+func TestVariableSetsApplyToAndRemoveFromWorkspaces(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	vsTest, _ := createVariableSet(t, client, orgTest, VariableSetCreateOptions{})
```
sebasslash, line 200: While I know deleting an organization will cascade and delete any children relationships, there have been instances where this cascade behavior fails. Therefore, I'd still `defer` each cleanup method to be on the safe side. This one is not a huge deal and is very edge case, so feel free to `defer` this comment (😆).

variable_set.go
sebasslash: Just to be clear, it should be _all_ structs at the top of the file, followed by receiver methods and so on...

sebasslash (changes requested):
All in all, the changes look 🔥 -- There are some minor things mentioned below!

variable_set.go
byronwolfman: Cheers, happy to move these. To be clear: would you like me to move just these newly-added structs, or all previously-existing ones too? It looks like workspace.tf follows the convention you're describing as a matter of prior art.

variable_set_test.go
byronwolfman: My particular discipline is figuring out why impossible things keep happening, so I'm happy to add a `defer` here. :D

variable_set.go
sebasslash: Yes if you would not mind reorganizing the file that would be great! 🚀

variable_set_test.go
byronwolfman: Addressed!

variable_set.go
byronwolfman: Addressed, but this may make the diff somewhat more difficult to read. I've split the re-arrangement into its own commit, here: https://github.com/hashicorp/go-tfe/pull/375/commits/ba9fb5dafb989f5242dfb8bcb7d7fb10e7830be1

sebasslash (approved):
🔥  - Ran integration test and verified the code locally

variable_set.go
```diff
@@ -63,43 +72,39 @@ const (
...
+	// A description to provide context for the variable set.
+	Description *string `jsonapi:"attr,description,omitempty"`
 
-	vl := &VariableSetList{}
-	err = s.client.do(ctx, req, vl)
-	if err != nil {
-		return nil, err
-	}
+	// If true the variable set is considered in all runs in the organization.
+	Global *bool `jsonapi:"attr,global,omitempty"`
```
brandonc, line 98: This appears to be an optional value (pointer, omitempty) but we make sure the value isn't nil within VariableSetCreateOptions valid method

variable_set.go
```diff
@@ -118,17 +123,60 @@ type VariableSetCreateOptions struct {
...
+
+// List all Variable Sets in the organization
+func (s *variableSets) List(ctx context.Context, organization string, options *VariableSetListOptions) (*VariableSetList, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
 	}
-	if !validString(o.Name) {
-		return ErrRequiredName
+	if options != nil {
+		if err := options.valid(); err != nil {
```
brandonc, line 162: 💅 Since nil appears to be allowed for options, you could implement this check (o == nil) within the valid() method

variable_set_test.go
```diff
@@ -190,16 +191,83 @@ func TestVariableSetsDelete(t *testing.T) {
...
+
+	vsTest, vsTestCleanup := createVariableSet(t, client, orgTest, VariableSetCreateOptions{})
+	defer vsTestCleanup()
+
+	wTest1, wTest1Cleanup := createWorkspace(t, client, orgTest)
+	defer wTest1Cleanup()
+	wTest2, wTest2Cleanup := createWorkspace(t, client, orgTest)
+	defer wTest2Cleanup()
+
+	t.Run("with first workspace added", func(t *testing.T) {
```
brandonc, line 209: Would love to see some test cases for invalid options

variable_set.go
byronwolfman: Hey! I can't speak to this particular type. It appears in the diff because I re-ordered everything in variable_set.go at Sebastian's request, but I'm not sure I have the necessary context to understand the ask here or competently resolve it. It might make for a good GH issue to handle in a separate PR.

variable_set.go
byronwolfman: Same as above: I suspect you're right, but I don't have any context for this, and it only appears as a diff because of this file being massively re-arranged.

variable_set_test.go
byronwolfman: Cheers, will add some more test cases!

brandonc (approved):

variable_set_test.go
byronwolfman: Ok, some test cases added for invalid options in https://github.com/hashicorp/go-tfe/pull/375/commits/bd4f215a4bf912756de44e12b4a5aa3e33fbfd73 (and hilariously corrected via https://github.com/hashicorp/go-tfe/pull/375/commits/0fac7ae180594f9ecd4c2cde694ad2a23ea86f04).

sebasslash (approved):
Fuego 🔥

PR #384, opened 2022-04-20
Resolve flaky tests

sebasslash:


helper_test.go
```diff
@@ -1161,6 +1161,38 @@ func createVariableSetVariable(t *testing.T, client *Client, vs *VariableSet, op
 	}
 }
 
+func waitForSVOutputs(t *testing.T, client *Client, svID string) {
+	tick := time.NewTicker(2 * time.Second)
+
+	retries := 1
+	maxRetries := 5
+
+	for {
```
brandonc, line 1170: The solution here is pretty clear, but could it be extracted and reused? terraform-plugin-sdk has a retry function that you can use as inspiration, but essentially I'm thinking that you can provide a function that is retried, and within that you can return an error if len(items) == 0

state_version_output_integration_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const waitForStateVersionOutputs = 1000 * time.Millisecond
+const waitForStateVersionOutputs = 1500 * time.Millisecond
```
brandonc, line 15: Is this still used?

uturunku1 (dismissed):
Let's give these changes a try!
I noticed that there was no PR description, so I took the liberty to add one for you, but please adjust it as you wish.
These `time.sleep` for the test was added about 1 year and half ago. It took a bit of time before these tests started failing consistently. It makes me think that it is really the platform that is pushing those inconsistencies over the edge.
In the past I made a similar-ish change to yours for another set of flaky tests. Except that I made adjustments to the usage of the library  https://github.com/hashicorp/go-retryablehttp (which is imported in this repo). The changes certainly helped to fix that set of flaky tests, but other started failing over time. I see a pattern for how we patch those flaky tests, and that seems to work, at least temporarily.
Thanks for the investigation and these changes!

helper_test.go
```diff
@@ -1161,6 +1164,45 @@ func createVariableSetVariable(t *testing.T, client *Client, vs *VariableSet, op
 	}
 }
 
+func waitForSVOutputs(t *testing.T, client *Client, svID string) {
+	t.Helper()
+	retry(func() (interface{}, error) {
```
brandonc, line 1169: I think you should interpret the return error value of retry and possibly set t.Error or t.Fatal if it's not nil?

brandonc (commented):
This is nicer!

helper_test.go
```diff
@@ -1161,6 +1165,74 @@ func createVariableSetVariable(t *testing.T, client *Client, vs *VariableSet, op
...
+
+		wg.Done()
+		return ws, nil
+	})
+
+	wg.Wait()
+}
+
+func retry(f retryableFn) (interface{}, error) {
+	tick := time.NewTicker(2 * time.Second)
```
annawinkler, line 1214: What's the number 2?

helper_test.go
```diff
@@ -1161,6 +1165,74 @@ func createVariableSetVariable(t *testing.T, client *Client, vs *VariableSet, op
...
+
+	retries := 0
+	maxRetries := 5
```
annawinkler, line 1217: How did we determine to set the max retries to 5? Is this value something we'd want to set in an external config file or via an environment variable?

helper_test.go
sebasslash: It's effectively the quantity "multiplied" by the unit of time, so in this case it's 2 seconds.

helper_test.go
sebasslash: We could, however these retries are for select use cases in our test where some action in the API doesn't happen instantaneously. I think waiting worst case ~10 seconds is ample amount of time. If certain async operations are taking longer than that, there might be an underlying issue with the API itself (I think ?).

helper_test.go
annawinkler: Cool! What do you think about moving the unit of time to a variable external to this loop, e.g., `TICKER_DURATION`?

helper_test.go
annawinkler: Got it - thanks.

Uk1288 (commented):
🎉

Uk1288 (approved):
LGTM

annawinkler (approved):
🥇 Thank you for making these updates and the PR description. Let's try it!

PR #388, opened 2022-05-02
fix: run actions don't accept json-api resource request bodies

brandonc:
## Description

According to the docs and in practice, the run actions group of endpoints commonly accept a request body with a single "comment" attribute. This isn't a json-api resource document, and it shouldn't be encoded into a "data" → "attributes" nesting.

As far as I know, these are the only request bodies that aren't compliant, and it's not worthwhile to convert these endpoints from actions to resources.

## Testing plan

I added an integration test for this in particular, but I also I wrote a simple program to POST a run apply to a [service like requestb.in](https://requestbin.com/r/env12e9eawjlf/28cdloPxOtlJmSk1N4Qk1zIVn1V) so you can see the request body:

```
package main

import (
	"context"

	"github.com/hashicorp/go-tfe"
)

func main() {
		Address: "https://env12e9eawjlf.x.pipedream.net",
		Token:   "xxx.atlasv1.xxx",
	})

	if err != nil {
		panic(err)
	}

	err = client.Runs.Apply(context.Background(), "run-NjC97hxkfWZ4eHTA", tfe.RunApplyOptions{
		Comment: tfe.String("hello, world"),
	})

	if err != nil {
		panic(err)
	}
}
```

## External links

- [API documentation](https://www.terraform.io/cloud-docs/api-docs/run#apply-a-run)

### DIFF
--- a/run.go
+++ b/run.go
@@ -253,13 +253,13 @@ type RunCreateOptions struct {
 // RunApplyOptions represents the options for applying a run.
 type RunApplyOptions struct {
 	// An optional comment about the run.
-	Comment *string `jsonapi:"attr,comment,omitempty"`
+	Comment *string `json:"comment,omitempty"`
 }
 
 // RunCancelOptions represents the options for canceling a run.
 type RunCancelOptions struct {
 	// An optional explanation for why the run was canceled.
-	Comment *string `jsonapi:"attr,comment,omitempty"`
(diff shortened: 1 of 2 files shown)


Uk1288 (approved):
LGTM 🎉 , tested the run actions (Discard, apply, cancel) locally.

PR #399, opened 2022-05-10
Add helper script to setup envvars via envchain

annawinkler:
## Description

Add a helper script to simplify setting env vars via envchain. Script prompts for an envchain workspace name, then invokes envchain with the environment variables.

## Testing plan

1. Pull this branch
2. Run `./scripts/setup-test-envvars.sh`
3. If you don't have envchain, the script will exit and request that it be installed via homebrew (this part could be more generic).
4. If you do have envchain, the script prompts for an envchain workspace name, then calls envchain with that workspace name and these 4 envvars: TFE_ADDRESS TFE_TOKEN GITHUB_TOKEN GITHUB_POLICY_SET_IDENTIFIER

```

...
```

### DIFF
--- a/scripts/setup-test-envvars.sh
+++ b/scripts/setup-test-envvars.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash -e
+
+# setup-test-envvars.sh
+#
+# A helper script that uses envchain (https://github.com/sorah/envchain) to set environment variables for tests.
+# It's required to have TFE_ADDRESS and TFE_TOKEN set, the others are optional.
+#
+type envchain >/dev/null 2>&1 || { echo >&2 "Required executable 'envchain' not found - install it via 'brew install envchain'. Exiting."; exit 1; }
+
+echo "Set environment variables (envvars) for running tests locally"
+echo " envchain will prompt you for values for 4 envvars"
+echo " TFE_ADDRESS and TFE_TOKEN are required, the others are optional,"
+echo " press 'return' to skip them"
+echo ""
+
+read -p "Enter the namespace you want to use in envchain [go-tfe]: " namespace
+namespace=${namespace:-go-tfe}
+envchain --set ${namespace} TFE_ADDRESS TFE_TOKEN GITHUB_TOKEN GITHUB_POLICY_SET_IDENTIFIER
+echo "Done! To see the values: envchain ${namespace} env"


annawinkler:
If we like this script, I can update the testing docs to reference the script.

scripts/setup-test-envvars.sh
```diff
@@ -0,0 +1,17 @@
...
+# It's required to have TFE_ADDRESS and TFE_TOKEN set, the others are optional.
+#
+type envchain >/dev/null 2>&1 || { echo >&2 "Required executable 'envchain' not found - install it via 'brew install envchain'. Exiting."; exit 1; }
+
+echo "Set environment variables (envvars) for running tests locally"
+echo " envchain will prompt you for values for 4 envvars"
+echo " TFE_ADDRESS and TFE_TOKEN are required, the others are optional, hit 'return' to skip them"
+echo ""
+echo "What workspace name do you want to use in envchain?"
+read workspace
```
uturunku1, line 15: Should we name this "namespace" instead to avoid confusion with the popular "workspace" term used by terraform?

uturunku1, line 15: Should we name this "namespace" instead to avoid confusion with the popular "workspace" term used by terraform?

uturunku1 (commented):
Should we consider alternative solutions to envchain? A concern we have that this script is customized to a very specific tool and other may not end up using this script because everyone ends using their own alternative solution.

scripts/setup-test-envvars.sh
```diff
@@ -0,0 +1,17 @@
...
+# A helper script that uses envchain (https://github.com/sorah/envchain) to set environment variables for tests.
+# It's required to have TFE_ADDRESS and TFE_TOKEN set, the others are optional.
+#
+type envchain >/dev/null 2>&1 || { echo >&2 "Required executable 'envchain' not found - install it via 'brew install envchain'. Exiting."; exit 1; }
+
+echo "Set environment variables (envvars) for running tests locally"
+echo " envchain will prompt you for values for 4 envvars"
+echo " TFE_ADDRESS and TFE_TOKEN are required, the others are optional, hit 'return' to skip them"
+echo ""
+echo "What workspace name do you want to use in envchain?"
```
uturunku1, line 14: Should the program name our namespace for us? Create a default name like "go-tfe"

scripts/setup-test-envvars.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# setup-test-envvars.sh
+#
+# A helper script that uses envchain (https://github.com/sorah/envchain) to set environment variables for tests.
+# It's required to have TFE_ADDRESS and TFE_TOKEN set, the others are optional.
+#
+type envchain >/dev/null 2>&1 || { echo >&2 "Required executable 'envchain' not found - install it via 'brew install envchain'. Exiting."; exit 1; }
+
+echo "Set environment variables (envvars) for running tests locally"
+echo " envchain will prompt you for values for 4 envvars"
```
uturunku1, line 11: Nick suggest resizing the spacing, which defaults to 80.

sebasslash (commented):
I won't hit approve cause I'm assuming you're going to address some changes requested by @uturunku1 (so it will dismiss my approval anyways) -- I like the idea of this script, thank you for this 👍 

Once they're addressed, everything looks good to me ✅

PR #401, opened 2022-05-16
🔧 Bump the circleci resource class when we run

annawinkler:
## Description

Bump the resource class to something reasonable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,8 @@ jobs:
         environment:
           TEST_RESULTS_DIR: *test_results_dir
 
+    resource_class: large
+
     steps:
       - checkout
 


brandonc:
I'm curious about the indications that the CI machine doesn't have enough resources. I would imagine testing the client should not be very intense because it is often waiting on the API. Also it may be worth noting that atlas/tflocal is not running in circle and isn't affected by this bump.

annawinkler:
That's a great question! Here is a screenshot of what CircleCI indicated on another build:
Screen Shot 2022-05-17 at 4 43 58 PM

I figured it was worth a try. And it sounds like the larger resource is included in our plan.

brandonc:
I think this is basically a dark pattern to juice sales because, while larger resources are included, they cost more credits per time unit. I think we should leave it as is unless a green run on this branch is substantially faster than main.

annawinkler:
Because we've been having consistent test failures, it's hard to tell if this change would help or not. I guess we could get the test failures fixed and then evaluate the performance.

annawinkler:
I'm going to close this for now. When I've run the tests, they take about 10 minutes. Maybe that's fine? 🔥

PR #403, opened 2022-05-17
Log the warning that agent pool cleanup failed

annawinkler:
## Description

Log the message that the agent pool cleanup failed. The org cleanup will cleanup the agent pool, and if that fails, we rebuild everything every 24 hours. This does not need to be a test failure.

## Testing plan

1. Run the tests
2. They pass! (Right now they are failing)

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -60,7 +60,7 @@ func createAgentPool(t *testing.T, client *Client, org *Organization) (*AgentPoo
 
 	return pool, func() {
 		if err := client.AgentPools.Delete(ctx, pool.ID); err != nil {
-			t.Errorf("Error destroying agent pool! WARNING: Dangling resources "+
+			t.Logf("Error destroying agent pool! WARNING: Dangling resources "+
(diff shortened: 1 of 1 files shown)


helper_test.go
```diff
@@ -60,7 +60,7 @@ func createAgentPool(t *testing.T, client *Client, org *Organization) (*AgentPoo
 
 	return pool, func() {
 		if err := client.AgentPools.Delete(ctx, pool.ID); err != nil {
-			t.Errorf("Error destroying agent pool! WARNING: Dangling resources "+
```
brandonc, line 63: I think we should do the same thing on org cleanup. I think this error may prevent org cleanup until org delete is async

annawinkler: OK - is there a way we could confirm that?

brandonc: No, aside from seeing if we hit any new dangling resource errors without it.

brandonc (approved):

PR #404, opened 2022-05-17
✨ Add registry provider client and integration tests

annawinkler:
## Description

This PR was broken out of this [PR](https://github.com/hashicorp/go-tfe/pull/313).

This PR adds support for the private registry provider API.

## Testing plan

To test these changes locally, create a local go project, and in the `go.mod` include a line similar to the following:

```
```

Then, for a registry provider, you can test:
* Create
* Delete
* Read
* List

## External links

- [Registry Providers API](https://www.terraform.io/cloud-docs/api-docs/private-registry/providers)
- [Private Registry Provider Requirements](https://github.com/hashicorp/go-tfe/pull/399)

## Output from tests

```

...
```

### DIFF
--- a/registry_provider.go
+++ b/registry_provider.go
@@ -1,0 +1,248 @@
+package tfe
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+)
+
+// Compile-time proof of interface implementation.
+var _ RegistryProviders = (*registryProviders)(nil)
+
+// RegistryProviders describes all the registry provider related methods that the Terraform
+// Enterprise API supports.
+//
+// TFE API docs: https://www.terraform.io/docs/cloud/api/providers.html
+type RegistryProviders interface {
+	// List all the providers within an organization.
+	List(ctx context.Context, organization string, options *RegistryProviderListOptions) (*RegistryProviderList, error)
+
+	// Create a registry provider
+	Create(ctx context.Context, organization string, options RegistryProviderCreateOptions) (*RegistryProvider, error)
+
+	// Read a registry provider
+	Read(ctx context.Context, organization string, registryName RegistryName, namespace string, name string, options *RegistryProviderReadOptions) (*RegistryProvider, error)
+
+	// Delete a registry provider
+	Delete(ctx context.Context, organization string, registryName RegistryName, namespace string, name string) error
+}
+
+// registryProviders implements RegistryProviders.
+type registryProviders struct {
+	client *Client
+}
+
+// RegistryName represents which registry is being targeted
+type RegistryName string
+
+// List of available registry names
+const (
+	PrivateRegistry RegistryName = "private"
+	PublicRegistry  RegistryName = "public"
+)
+
+// RegistryProvider represents a registry provider
+type RegistryProvider struct {
+	ID           string                       `jsonapi:"primary,registry-providers"`
+	Namespace    string                       `jsonapi:"attr,namespace"`
+	Name         string                       `jsonapi:"attr,name"`
+	RegistryName RegistryName                 `jsonapi:"attr,registry-name"`
+	Permissions  *RegistryProviderPermissions `jsonapi:"attr,permissions"`
+	CreatedAt    string                       `jsonapi:"attr,created-at"`
+	UpdatedAt    string                       `jsonapi:"attr,updated-at"`
+
+	// Relations
+	Organization             *Organization             `jsonapi:"relation,organization"`
+	RegistryProviderVersions []RegistryProviderVersion `jsonapi:"relation,registry-provider-version"`
+}
+
+type RegistryProviderPermissions struct {
+	CanDelete bool `jsonapi:"attr,can-delete"`
+}
+
+type RegistryProviderListOptions struct {
+	ListOptions
+	// A query string to filter by registry_name
+	RegistryName *RegistryName `url:"filter[registry_name],omitempty"`
+	// A query string to filter by organization
+	OrganizationName *string `url:"filter[organization_name],omitempty"`
+	// A query string to do a fuzzy search
+	Search *string `url:"q,omitempty"`
+}
+
+type RegistryProviderList struct {
+	*Pagination
+	Items []*RegistryProvider
+}
+
+func (o RegistryProviderListOptions) valid() error {
+	return nil
+}
+
+func (r *registryProviders) List(ctx context.Context, organization string, options *RegistryProviderListOptions) (*RegistryProviderList, error) {
+
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+	if options != nil {
+		if err := options.valid(); err != nil {
+			return nil, err
+		}
+	}
+
+	u := fmt.Sprintf("organizations/%s/registry-providers", url.QueryEscape(organization))
+	req, err := r.client.newRequest("GET", u, options)
+	if err != nil {
+		return nil, err
+	}
+
+	pl := &RegistryProviderList{}
+	err = r.client.do(ctx, req, pl)
+	if err != nil {
+		return nil, err
+	}
+
+	return pl, nil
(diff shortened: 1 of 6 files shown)


registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+	*Pagination
+	Items []*RegistryProvider
+}
+
+func (o RegistryProviderListOptions) valid() error {
+	return nil
+}
+
+func (r *registryProviders) List(ctx context.Context, organization string, options *RegistryProviderListOptions) (*RegistryProviderList, error) {
+
```
sebasslash, line 84: unnecessary whitespace ✏️

registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+// RegistryProviderCreateOptions is used when creating a registry provider
+type RegistryProviderCreateOptions struct {
+	// Type is a public field utilized by JSON:API to
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,registry-providers"`
+
+	Namespace    *string       `jsonapi:"attr,namespace"`
```
sebasslash, line 117: These fields should not be pointers unless the field is a struct type or the API accepts a `null` value (and perhaps that may be so since `omitempty` is omitted?)

registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+func (r *registryProviders) Create(ctx context.Context, organization string, options RegistryProviderCreateOptions) (*RegistryProvider, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+	// Private providers must match their namespace and organization name
+	// This is enforced by the API as well
+	if *options.RegistryName == PrivateRegistry && organization != *options.Namespace {
```
sebasslash, line 150: this check should live inside the `RegistryProviderCreateOptions.valid()` function

registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+	prv := &RegistryProvider{}
+	err = r.client.do(ctx, req, prv)
+	if err != nil {
+		return nil, err
+	}
+
+	return prv, nil
+}
+
+type RegistryProviderReadOptions struct {
```
sebasslash, line 171: Empty struct? 👻

registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (r *registryProviders) Read(ctx context.Context, organization string, registryName RegistryName, namespace string, name string, options *RegistryProviderReadOptions) (*RegistryProvider, error) {
```
sebasslash, line 174: Whenever I see a long method signature like that I wonder if we can use a struct to keep the signature concise? An example:
https://github.com/hashicorp/go-tfe/blob/11a4eb9078cef3c9caf389325ddc6327e749ea94/registry_module.go#L75-L82

registry_provider.go
```diff
@@ -0,0 +1,248 @@
...
+	if !validStringID(&organization) {
```
sebasslash, line 175: All these `validStringID()` checks should probably be refactored into a `valid()` function and perhaps should live in the `ReadOptionsStruct` instead of as method params (back to my previous point). What do you think?

registry_provider_integration_test.go
```diff
@@ -0,0 +1,442 @@
...
+		defer orgTestCleanup()
+
+		createN := 10
+		providers := make([]*RegistryProvider, 0)
+		// these providers will be destroyed when the org is cleaned up
+		for i := 0; i < createN; i++ {
+			providerTest, _ := createPublicRegistryProvider(t, client, orgTest)
+			providers = append(providers, providerTest)
+		}
+		for i := 0; i < createN; i++ {
```
sebasslash, line 27: Duplicate for loop?

registry_provider_integration_test.go
```diff
@@ -0,0 +1,442 @@
...
+				ListOptions: ListOptions{
+					PageNumber: 0,
+					PageSize:   providerN,
+				},
+			})
+			require.NoError(t, err)
+			assert.NotEmpty(t, returnedProviders.Items)
+			assert.Equal(t, providerN, returnedProviders.TotalCount)
+			assert.Equal(t, 1, returnedProviders.TotalPages)
+			for _, rp := range returnedProviders.Items {
```
sebasslash, line 45: Alternatively, we can condense this logic to:

```go
for _, p := range providers {
  assert.Contains(t, returnedProviders.Items, p)
}
```

Careful with `assert.Contains()` as it performs deep equality. (I've been burned before with `Included` relationships before 👉 🔥 )

sebasslash: Furthermore, this sort of logic (checking that the desired provider(s) exists) seems better reserved for the `Create` test; for this test we should only care about that endpoint returns the correct # of elements, query params work as intended and that the fields are deserialized properly (although we can reserve checking the deserialization of fields in the `Read` test).

registry_provider_integration_test.go
```diff
@@ -0,0 +1,442 @@
...
+					if rp.ID == p.ID {
+						foundProvider = true
+						break
+					}
+				}
+				assert.True(t, foundProvider, "Expected to find provider %s but did not:\nexpected:\n%v\nreturned\n%v", rp.ID, providers, returnedProviders)
+			}
+		})
+
+		t.Run("returns pages", func(t *testing.T) {
```
sebasslash, line 57: For this kind of test, the usual convention is to set `PageSize` and `PageNumber` to some incredibly large number, which I think is the easiest way to verify pagination is working as intended. Here's an example:

https://github.com/hashicorp/go-tfe/blob/11a4eb9078cef3c9caf389325ddc6327e749ea94/workspace_integration_test.go#L44-L58

registry_provider_integration_test.go
```diff
@@ -0,0 +1,442 @@
...
+				assert.Equal(t, publicName, rp.RegistryName)
+				assert.True(t, foundProvider, "Expected to find provider %s but did not:\nexpected:\n%v\nreturned\n%v", rp.ID, providers, returnedProviders)
+			}
+		})
+
+		t.Run("searches", func(t *testing.T) {
+			expectedProvider := providers[0]
+			returnedProviders, err := client.RegistryProviders.List(ctx, orgTest.Name, &RegistryProviderListOptions{
+				Search: &expectedProvider.Name,
+				ListOptions: ListOptions{
```
sebasslash, line 119: Can we omit the ListOptions field here?

sebasslash (changes requested):
Great work @annawinkler and @joekarl ... I have not been able to review all of it as it is quite a hefty PR, but hopefully a lot the comments translate to the rest of the unreviewed code. 

Files I've reviewed:
helper_test.go
registry_provider.go
And the `List` test in the registry_provider_integration_test.go

Definitely would like @brandonc and @uturunku1 to give it a pass.

registry_provider.go
annawinkler: This is refactored in the next PR. Wondering if it's simpler to go with the large PR that has everything?

annawinkler:
Closing this PR in favor of the one large [one](https://github.com/hashicorp/go-tfe/pull/313) since refactoring has occurred there. I'll address the review comments in PR 313.

registry_provider.go
annawinkler: In PR https://github.com/hashicorp/go-tfe/pull/313 they are not pointers.

registry_provider.go
annawinkler: This has been refactored.

registry_provider_integration_test.go
annawinkler: It looks like it but we create private and public registry providers in the test.

PR #405, opened 2022-05-17
✨ Add private registry provider version

annawinkler:
## Description

This PR was broken out of this [PR](https://github.com/hashicorp/go-tfe/pull/313).

This PR adds support for the private registry provider version API. The first PR for these set of changes is [this](https://github.com/hashicorp/go-tfe/pull/404) one.

## Testing plan

To test these changes locally, create a local go project, and in the `go.mod` include a line similar to the following:

```
```

Then, for a registry provider version, you can test:
* Create
* Delete
* Read
* List

## External links

- [Registry Providers API](https://www.terraform.io/cloud-docs/api-docs/private-registry/providers)
- [Private Registry Provider Requirements](https://github.com/hashicorp/go-tfe/pull/399)
- [Private Provider Versions and Platforms API](https://www.terraform.io/cloud-docs/api-docs/private-registry/provider-versions-platforms)

## Output from tests

```

...
```

---
  - https://app.asana.com/0/0/1202295552965199

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"sync"
 	"testing"
@@ -761,20 +762,29 @@ func createPrivateRegistryProvider(t *testing.T, client *Client, org *Organizati
 
 	ctx := context.Background()
 
-	privateName := PrivateRegistry
-
 	options := RegistryProviderCreateOptions{
-		Name:         String("tst-name-" + randomString(t)),
-		Namespace:    &org.Name,
-		RegistryName: &privateName,
+		Name:         "tst-name-" + randomString(t),
+		Namespace:    org.Name,
+		RegistryName: PrivateRegistry,
 	}
+
 	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
+
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	prv.Organization = org
+
 	return prv, func() {
-		if err := client.RegistryProviders.Delete(ctx, org.Name, prv.RegistryName, prv.Namespace, prv.Name); err != nil {
+		id := RegistryProviderID {
+			OrganizationName: org.Name,
+			RegistryName:     prv.RegistryName,
+			Namespace:        prv.Namespace,
+			Name:             prv.Name,
+		}
+
+		if err := client.RegistryProviders.Delete(ctx, id); err != nil {
 			t.Errorf("Error destroying registry provider! WARNING: Dangling resources\n"+
 				"may exist! The full error is shown below.\n\n"+
 				"Registry Provider: %s/%s\nError: %s", prv.Namespace, prv.Name, err)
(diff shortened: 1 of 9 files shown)


helper_test.go
```diff
@@ -751,6 +753,139 @@ func createRunTask(t *testing.T, client *Client, org *Organization) (*RunTask, f
...
+
+	options := RegistryProviderVersionCreateOptions{
+		Version: randomSemver(t),
+		KeyID:   randomString(t),
+	}
+
+	prvv, err := client.RegistryProviderVersions.Create(ctx, providerID, options)
+
+	if err != nil {
+		t.Fatal(err)
```
brandonc, line 866: same 'return' concern as above

registry_provider.go
```diff
@@ -0,0 +1,265 @@
...
+	PrivateRegistry RegistryName = "private"
+	PublicRegistry  RegistryName = "public"
+)
+
+// RegistryProviderIncludeOps represents which jsonapi include can be used with registry providers
+type RegistryProviderIncludeOps string
+
+// List of available includes
+const (
+	RegistryProviderVersionsInclude RegistryProviderIncludeOps = "registry-provider-versions"
```
brandonc, line 49: I don't see this documented or in the API. Is it possible to include this relationship?

brandonc (commented):
(As I mentioned in slack, I only got about halfway through this review)

helper_test.go
```diff
@@ -751,6 +753,139 @@ func createRunTask(t *testing.T, client *Client, org *Organization) (*RunTask, f
...
+	options := RegistryProviderCreateOptions{
+		Name:         "tst-name-" + randomString(t),
+		Namespace:    org.Name,
+		RegistryName: PrivateRegistry,
+	}
+
+	prv, err := client.RegistryProviders.Create(ctx, org.Name, options)
+
+	if err != nil {
+		t.Fatal(err)
```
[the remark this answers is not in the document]
annawinkler: I think the line does we want it to - [Fatal](https://pkg.go.dev/testing#B.FailNow) is equivalent to Log plus [FailNow](https://pkg.go.dev/testing#B.FailNow). FailNow stops execution, calls runtime.Goexit, and continues execution with the next test or benchmark.

registry_provider.go
annawinkler: 🤔 You're right, when I re-reviewed the documentation, it is not clear that this include is an option. I tested it locally with this request:

```
curl --location --request GET 'https://LOCALURL/api/v2/organizations/amazing-org/registry-providers/private/amazing-org/zebra-provider?include=registry-provider-versions' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Authorization: Bearer TOKEN' \
--header 'Cookie: __profilin=p%3Dt' \
--data-raw '{
  "data": {
    "type": "registry-providers",
    "attributes": {
      "name": "test-registry-provider-5de6cb61-b5ea-e72e-13a2-c9aab0b5c3ac",
      "namespace": "amazing-org",
      "registry-name": "private"
    }
  }
}'
```

And the response does include the registry provider versions (see the attached json). I think we'll need to update the docs. What are your thoughts?

[response.txt](https://github.com/hashicorp/go-tfe/files/8757919/response.txt)

annawinkler:
I'm going to close this PR in favor of the one large [one](https://github.com/hashicorp/go-tfe/pull/313). I'll address the comments here in the one big PR.

registry_provider_version.go
```diff
@@ -0,0 +1,264 @@
...
+type RegistryProviderVersionListOptions struct {
+	ListOptions
+}
+
+type RegistryProviderVersionReadOptions struct{}
+
+type RegistryProviderVersionCreateOptions struct {
+	Version   string   `jsonapi:"attr,version"`
+	KeyID     string   `jsonapi:"attr,key-id"`
+	Protocols []string `jsonapi:"attr,protocols"`
```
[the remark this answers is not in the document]
annawinkler: I like that idea!

PR #407, opened 2022-05-18
Add support for the Audit Trail API

sebasslash:
## Description

Adds support for the Audit Trail API. One thing to note is that this particular API requires an organization token and does not follow the JSON+API spec as most of our endpoints do. Thusly the implementation of the `List()` method will create a new retryable http client and send the request using that client -- I figured this would be better than altering the existing client that is shared across our interfaces. 

## Testing plan

To run the integration test:
```sh
go test -v ./... -run TestAuditTrailsList -tags=integration
```

## Output from tests
```sh
$  go test -v ./... -run TestAuditTrailsList -tags=integration
=== RUN   TestAuditTrailRead
=== RUN   TestAuditTrailRead/with_no_specified_timeframe
=== RUN   TestAuditTrailRead/with_no_specified_timeframe/with_resource_deserialized_correctly
=== RUN   TestAuditTrailRead/with_no_specified_timeframe/with_auth_deserialized_correctly
=== RUN   TestAuditTrailRead/with_no_specified_timeframe/with_request_deserialized_correctly
=== RUN   TestAuditTrailRead/using_since_query_param
--- PASS: TestAuditTrailRead (2.11s)
    --- PASS: TestAuditTrailRead/with_no_specified_timeframe (0.17s)
        --- PASS: TestAuditTrailRead/with_no_specified_timeframe/with_resource_deserialized_correctly (0.00s)
        --- PASS: TestAuditTrailRead/with_no_specified_timeframe/with_auth_deserialized_correctly (0.00s)
        --- PASS: TestAuditTrailRead/with_no_specified_timeframe/with_request_deserialized_correctly (0.00s)
    --- PASS: TestAuditTrailRead/using_since_query_param (1.51s)
PASS
ok      github.com/hashicorp/go-tfe     2.872s
?       github.com/hashicorp/go-tfe/examples/organizations      [no test files]
?       github.com/hashicorp/go-tfe/examples/workspaces [no test files]
?       github.com/hashicorp/go-tfe/mocks       [no test files]
```


audit_trail.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		u.RawQuery = encodeQueryParams(q)
+	}
+
+	fmt.Println(headers)
+
+	fmt.Println(u.String())
```
Uk1288, line 87: to remove `fmt.Println`

audit_trail.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, err
+	}
+	fmt.Println(string(body))
+
+	atl := &AuditTrailList{}
+	if err := json.Unmarshal(body, atl); err != nil {
+		return nil, err
+	}
+
+	fmt.Println(atl)
```
Uk1288, line 120: to remove `fmt.Println`

brandonc (changes requested):
I would love to see all this implemented within a new type of Client, not as a part of existing `type Client`. Naming is tricky but perhaps something like JsonClient or AuditTrailsClient.

Rationale:

1. The endpoint is incompatible with the existing auth and type headers
2. The endpoint requires different token config (always an org token, which is rare)

What if the API was more like

```
auditConfig := DefaultConfig()
auditConfig.Token = orgToken
auditClient, err := NewAuditTrailClient(&auditConfig)

auditTrails, err := auditClient.AuditTrails.List(ctx)
```

one thing to look at to help with naming is if there is ANY OTHER public API that doesn't use json:api.

It just doesn't feel right to drop to retryablehttp, reusing client, within this method.

audit_trail_integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAuditTrailRead(t *testing.T) {
+	skipIfFreeOnly(t)
```
brandonc, line 17: Audit Trails are not available in TFE - I think we have a skipIfEnterprise?

audit_trail_integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+	skipIfNoOrgToken(t)
+
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgToken := os.Getenv("TFE_ORGANIZATION_TOKEN")
```
brandonc, line 23: Why did you decide that creating an org token would not work? I'm sorry I forgot.

sebasslash: Because we currently have no way in go-tfe to upgrade an organization to the business tier. The integration tests in atlas do in fact upgrade the entitlement set for an org but it's an internal endpoint as I understand it. 

I discussed with @taiidani and @annawinkler on potentially configuring tflocal to default any organization that's created to the business plan so we can run the entire test suite. This is non trivial and requires further internal discussion. 👍

audit_trail_integration_test.go
sebasslash: Done 👍

audit_trail_integration_test.go
taiidani: The non go-tfe endpoint that the Atlas integration tests use: https://github.com/hashicorp/atlas/blob/32947de5536a9df4fb79ce18868b0b4ad29bfa07/integration-tests-api/internal/testenv/helpers.go#L775-L838

brandonc:
Is this waiting on #415 ?

audit_trail.go
```diff
@@ -0,0 +1,140 @@
...
+	ImpersonatorID *string `json:"impersonator_id"`
+	OrganizationID string  `json:"organization_id"`
+}
+
+// AuditTrailResource represents the details of the API resource in the audit event.
+type AuditTrailResource struct {
+	ID     string                  `json:"id"`
+	Type   string                  `json:"type"`
+	Action string                  `json:"action"`
+	Meta   *map[string]interface{} `json:"meta"`
```
brandonc, line 51: I think there's no need to use a pointer here since map is already a reference type.

audit_trail.go
```diff
@@ -0,0 +1,140 @@
...
+// AuditTrail represents an event in the TFC audit log.
+type AuditTrail struct {
+	ID        string    `json:"id"`
+	Version   string    `json:"version"`
+	Type      string    `json:"type"`
+	Timestamp time.Time `json:"timestamp"`
+
+	Auth     *AuditTrailAuth     `json:"auth"`
+	Request  *AuditTrailRequest  `json:"request"`
+	Resource *AuditTrailResource `json:"resource"`
```
brandonc, line 63: What criteria do we use when deciding to use a pointer type for something like this? I tend to prefer "values-by-default" as described in this [blog post](https://medium.com/@meeusdylan/when-to-use-pointers-in-go-44c15fe04eac) and in particular I don't see a compelling reason to use them here (performance, mutability, consistency, etc)

audit_trail.go
```diff
@@ -0,0 +1,140 @@
...
+type AuditTrailList struct {
+	*Pagination
+
+	Items []*AuditTrail `json:"data"`
+}
+
+// AuditTrailListOptions represents the options for listing audit trails.
+type AuditTrailListOptions struct {
+	// Optional: Returns only audit trails created after this date
+	Since time.Time `url:"since,omitempty"`
```
brandonc, line 76: On the other hand, this looks like it should be a pointer since it is an emptyable struct ?

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-# v1.2.0 (Unreleased)
+# v1.3.0
+
+## Enhancements
+* Add support for Audit Trail API by @sebasslash [#407](https://github.com/hashicorp/go-tfe/pull/407)
+
+# v1.2.0
```
brandonc, line 6: 1.2.0 is still called Unreleased in main 😱

audit_trail.go
sebasslash: This was a great read, thanks. Will change

audit_trail.go
sebasslash: The zero value for `time.Time{}` is `0001-01-01 00:00:00 +0000 UTC` so it wouldn't be an issue. The only reason I wouldn't make it a pointer is `time.Time` isn't really ever used as a pointer.

README.md
```diff
@@ -66,7 +66,7 @@ This API client covers most of the existing Terraform Cloud API calls and is upd
 - [x] Agent Pools
 - [x] Agent Tokens
 - [x] Applies
-- [ ] Audit Trails
+- [x] Audit Trails
```
brandonc, line 69: 😎

brandonc (commented):
An orgToken parameter still doesn't sit right with me on the interface API. It's confusing because you already supplied an auth token when you created the client, and it feels like configuration as opposed to a list option.

Can we re-open the argument for having a second 'ComplianceClient' or 'PlatformClient' that consumes json instead of jsonapi and is configured with a separate token?

audit_trail.go
```diff
@@ -0,0 +1,141 @@
...
+// List all the audit events in an organization.
+func (s *auditTrails) List(ctx context.Context, options *AuditTrailListOptions) (*AuditTrailList, error) {
+	u, err := s.client.baseURL.Parse("/api/v2/organization/audit-trail")
+	if err != nil {
+		return nil, err
+	}
+
+	headers := make(http.Header)
+	headers.Set("Authorization", "Bearer "+s.client.token)
+	headers.Set("Content-Type", "application/json")
```
brandonc, line 90: we should set the user-agent to `go-tfe`, similar to tfe.go

brandonc (approved):
I've found that a user token will authenticate, returning an empty List and 200 response.

brandonc (approved):

brandonc (approved):

audit_trail.go
annawinkler: @sebasslash Is this something that still needs to be done?

sebasslash: Added in L89: `headers.Set("User-Agent", _userAgent)` where `_userAgent` is `go-tfe`

PR #400, opened 2022-05-13
Feature/add trigger patterns

matejrisek:
## Description

Add `trigger-patterns` to the client.

## Testing plan

N/A

## External links

- [Asana](https://app.asana.com/0/1201287664926243/1201660115329296)

PR #410, opened 2022-05-23
Allow empty trigger-prefixes/trigger-patterns to coexist

matejrisek:
## Description

Allowing both `trigger-patterns` and `trigger-prefixes` if at least one of them is empty.

## Testing plan

To test these changes locally, create a local go project, and in the go.mod include a line similar to the following:

Then, you can test that adding both `trigger-patterns` and `trigger-prefixes` to a workspace is allowed as long one or both are empty arrays. If you populate both with values we should raise an error.

## Output from tests

```

...
```

### DIFF
--- a/workspace.go
+++ b/workspace.go
@@ -1080,7 +1080,8 @@ func (o WorkspaceUpdateOptions) valid() error {
 	if o.AgentPoolID == nil && (o.ExecutionMode != nil && *o.ExecutionMode == "agent") {
 		return ErrRequiredAgentPoolID
 	}
-	if o.TriggerPrefixes != nil && o.TriggerPatterns != nil {
+	if o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 &&
+		o.TriggerPatterns != nil && len(o.TriggerPatterns) > 0 {
 		return ErrUnsupportedBothTriggerPatternsAndPrefixes
 	}
 
--- a/workspace.go
+++ b/workspace.go
@@ -1062,7 +1062,8 @@ func (o WorkspaceCreateOptions) valid() error {
 	if o.AgentPoolID == nil && (o.ExecutionMode != nil && *o.ExecutionMode == "agent") {
 		return ErrRequiredAgentPoolID
 	}
-	if o.TriggerPrefixes != nil && o.TriggerPatterns != nil {
+	if o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 &&
+		o.TriggerPatterns != nil && len(o.TriggerPatterns) > 0 {
 		return ErrUnsupportedBothTriggerPatternsAndPrefixes
 	}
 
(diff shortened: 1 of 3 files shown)


omarismail (dismissed):

annawinkler (changes requested):
Thanks for opening this PR! Could you add a few steps on how one might test this change locally? Thank you!

matejrisek:
> Thanks for opening this PR! Could you add a few steps on how one might test this change locally? Thank you!

@annawinkler 
Hey Anna, yeah no problem, I could add steps, but could you please clarify what should be covered with them?
Do we usually add these kind of steps? Or is it because of the peculiarities of how we enable feature flags that it be better to add them?

uturunku1:
> > Thanks for opening this PR! Could you add a few steps on how one might test this change locally? Thank you!
> 
> @annawinkler Hey Anna, yeah no problem, I could add steps, but could you please clarify what should be covered with them? Do we usually add these kind of steps? Or is it because of the peculiarities of how we enable feature flags that it be better to add them?

Hey Matej! To add to what Anna mention, we usually want contributors to add these kind of steps whenever a PR is adding a new feature or fixing a bug. Our team, the CLI team owns the project but a lot of the times we have other teams bringing in changes to the project that we lack context on. So we try to remind those contributors to fill us in as much as possible on what those changes are about and to leave instructions on how to smoke test their new functionality (this may include instructions on how to smoke test with Atlas as well, if this feature was sitting behind a feature flag).

I notice that this PR is not really bringing new functionality but just making adjustments to this previous PR: https://github.com/hashicorp/go-tfe/pull/400
So I don't think is super needed to add instructions on how we can smoke test your changes for this PR. However, for futures references, the previous PR (#400), is definitely that kind of PR were we like to see instructions on how to smoke test the new feature. Thanks!

workspace.go
```diff
@@ -1062,7 +1062,8 @@ func (o WorkspaceCreateOptions) valid() error {
 	if o.AgentPoolID == nil && (o.ExecutionMode != nil && *o.ExecutionMode == "agent") {
 		return ErrRequiredAgentPoolID
 	}
-	if o.TriggerPrefixes != nil && o.TriggerPatterns != nil {
+	if o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 &&
```
uturunku1, line 1065: Curious! Why do we validate the length of these attributes for the WorkspaceCreateOptions but not the WorkspaceUpdateOptions?

matejrisek: The reason being - I was hasty and forgot to add changes to the update part 🤦🏻‍♂️ 
Good catch!

matejrisek: a290810956ef053e4c08c864bcf0b9c65293346a

matejrisek:
@annawinkler @uturunku1 I've used one of the recent PRs as an example of a manual test steps. Could you please check whether that's sufficient enough? 

@uturunku1 great catch on not replicating the same logic for update as for creation :)

annawinkler (approved):
⭐ Thank you for the updates!

PR #414, opened 2022-05-25
Fix linting errors

annawinkler:
## Description

Fix issues seen when running `golangci-lint run`

## Testing plan

1. When you run `golangci-lint run` on this branch, you should not see any issues.

## Output from tests

```

...
```

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -1302,18 +1302,19 @@ func waitForRunLock(t *testing.T, client *Client, workspaceID string) {
 	wg.Wait()
 }
 
-func retry(f retryableFn) (interface{}, error) {
+func retry(f retryableFn) (interface{}, error) { //nolint
 	tick := time.NewTicker(tickDuration * time.Second)
-
 	retries := 0
 	maxRetries := 5
 
-	for {
+	defer tick.Stop()
+
+	for { //nolint
 		select {
 		case <-tick.C:
 			res, err := f()
 			if err == nil {
-				return res, err
+				return res, nil
 			}
 
 			if retries >= maxRetries {
--- a/tfe.go
+++ b/tfe.go
@@ -854,7 +854,7 @@ func checkResponseCode(r *http.Response) error {
 		return err
 	}
 
-	return fmt.Errorf(strings.Join(errs, "\n"))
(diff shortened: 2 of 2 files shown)


helper_test.go
```diff
@@ -1233,18 +1233,17 @@ func waitForRunLock(t *testing.T, client *Client, workspaceID string) {
...
+func retry(f retryableFn) (interface{}, error) { //nolint
 	tick := time.NewTicker(tickDuration * time.Second)
 
 	retries := 0
 	maxRetries := 5
 
 	for {
-		select {
-		case <-tick.C:
+		if tick.C != nil {
```
brandonc, line 1243: To my eye, this doesn't look equivalent. `<-tick.C` should be getting a value from a channel, but comparing the channel to nil doesn't seem like it would have the same effect?

tfe.go
```diff
@@ -863,7 +863,7 @@ func decodeErrorPayload(r *http.Response) ([]string, error) {
 	errPayload := &jsonapi.ErrorsPayload{}
 	err := json.NewDecoder(r.Body).Decode(errPayload)
 	if err != nil || len(errPayload.Errors) == 0 {
-		return errs, fmt.Errorf(r.Status)
+		return errs, fmt.Errorf("%s", r.Status)
```
brandonc, line 866: For consistency, it seems like this line should be `return errs, errors.New(r.Status)` since no formatting is done and r.Status is already a string (I think)

helper_test.go
```diff
@@ -1302,18 +1302,19 @@ func waitForRunLock(t *testing.T, client *Client, workspaceID string) {
...
+func retry(f retryableFn) (interface{}, error) { //nolint
 	tick := time.NewTicker(tickDuration * time.Second)
-
 	retries := 0
 	maxRetries := 5
 
-	for {
+	defer tick.Stop()
+
+	for { //nolint
```
annawinkler, line 1312: We discussed the two approaches for this block. The linter would have us simplify this block to iterate over the range of the ticker values, but we prefer to stick with the for / select option because it's consistent with how we use tickers in this library.

brandonc (approved):

PR #416, opened 2022-05-27
adds ext-state attribute to StateVersion create

brandonc:
## Description

This will be used in future versions of terraform to supply portable state details to the platform

The type, json.RawMessage, is defined as "a raw encoded JSON value. It implements Marshaler and Unmarshaler and can be used to delay JSON decoding or precompute a JSON encoding."

The encoding of the attribute is meant to be "any json object" and the backend does some basic validation on the message (it must have both "format_version" and "values" attributes.)

## Output from tests

```
$ TF_ACC=1 go test ./... -v -tags integration -run TestStateVersionsCreate

=== RUN   TestStateVersionsCreate
=== RUN   TestStateVersionsCreate/with_valid_options
=== RUN   TestStateVersionsCreate/with_external_state_representation
=== RUN   TestStateVersionsCreate/with_the_force_flag_set
=== RUN   TestStateVersionsCreate/with_a_run_to_associate_with
    state_version_integration_test.go:245: This can only be tested with the run specific token
=== RUN   TestStateVersionsCreate/without_md5_hash
=== RUN   TestStateVersionsCreate/withous_serial
=== RUN   TestStateVersionsCreate/without_state
=== RUN   TestStateVersionsCreate/with_invalid_workspace_id
--- PASS: TestStateVersionsCreate (17.07s)
    --- PASS: TestStateVersionsCreate/with_valid_options (3.59s)
    --- PASS: TestStateVersionsCreate/with_external_state_representation (3.48s)
    --- PASS: TestStateVersionsCreate/with_the_force_flag_set (4.18s)
    --- SKIP: TestStateVersionsCreate/with_a_run_to_associate_with (0.00s)
    --- PASS: TestStateVersionsCreate/without_md5_hash (0.00s)
    --- PASS: TestStateVersionsCreate/withous_serial (0.00s)
    --- PASS: TestStateVersionsCreate/without_state (0.00s)
    --- PASS: TestStateVersionsCreate/with_invalid_workspace_id (0.00s)
PASS
```

### DIFF
--- a/state_version.go
+++ b/state_version.go
@@ -3,6 +3,7 @@ package tfe
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"time"
@@ -138,6 +139,12 @@ type StateVersionCreateOptions struct {
 
 	// Optional: Specifies the run to associate the state with.
 	Run *Run `jsonapi:"relation,run,omitempty"`
+
+	// Optional: The external, json representation of state data.
+	// https://www.terraform.io/internals/json-format#state-representation
+	// Supplying this state representation can provide more details to the platform
+	// about the current terraform state.
+	ExtState json.RawMessage `jsonapi:"attr,ext-state,omitempty"`
 }
 
 // List all the state versions for a given workspace.
--- a/state_version_integration_test.go
+++ b/state_version_integration_test.go
@@ -121,6 +121,11 @@ func TestStateVersionsCreate(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	extState, err := ioutil.ReadFile("test-fixtures/ext-state-version/state.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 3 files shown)


annawinkler (commented):
⭐ This looks straight-forward, but I have one question that I'm curious about 🤔 (see the comment)

sebasslash (commented):
Changes look good. One question: where would I be able to fetch the external state representation? Is it available in `DownloadURL`?

brandonc:
@sebasslash no. At first, it won't be available for consumption in the public API and will be added only to the state parser internal API for use on the platform. Over time, as other useful features are identified, it can become a new download url on the state version resource.

brandonc:
@annawinkler If the error makes the json unparsable, the API will return a bad request response with details (BaseController L56). I don't think it's worthwhile to try and parse it all in this library as a validation.

annawinkler (approved):
:shipit:

PR #417, opened 2022-05-31
Add query option to agent pool list options

JarrettSpiker:
## Description

Add query option to agent pool list options

This will be blocked until the agent pool search feature is enabled (feature flag: `remove-agent-pool-limit`)

## Testing plan

This  code is exercised by an upcoming TFE provider change. It can be used to make the tfe_agent_pool data source much more efficient in organizations with more than 20 agents

## External links

- API documentation: in progress..
- TFE Provider: in progress...

## Output from tests

```
❯ go test .  -run TestAgentPoolsList -tags=integration
ok      github.com/hashicorp/go-tfe     3.740s
```


agent_pool.go
```diff
@@ -70,6 +70,8 @@ type AgentPoolListOptions struct {
 	// Optional: A list of relations to include. See available resources
 	// https://www.terraform.io/cloud-docs/api-docs/agents#available-related-resources
 	Include []AgentPoolIncludeOpt `url:"include,omitempty"`
+
+	Query string `url:"q,omitempty"`
```
sebasslash, line 74: Missing the code documentation for this field 💅

sebasslash (changes requested):
Nice and easy 🔥 ! Just missing the code documentation for the field.

agent_pool.go
JarrettSpiker: @sebasslash added, thanks! I also updated the changelog...I notice that 1.2.0 was still marked as unreleased, so I updated that as well, please let me know if that is not correct!

sebasslash (approved):
🔥 Nice

agent_pool_integration_test.go
```diff
@@ -68,6 +68,21 @@ func TestAgentPoolsList(t *testing.T) {
 		assert.Nil(t, pools)
 		assert.EqualError(t, err, ErrInvalidOrg.Error())
 	})
+
+	t.Run("with query options", func(t *testing.T) {
+
```
sebasslash, line 73: Ah, I believe you mentioned this query param was sitting behind a feature flag... in which case `skipIfBeta(t)` test helper will need to be added to this subtest otherwise the test will fail in CI... I don't think tflocal has the flag enabled?

sebasslash (changes requested):
Missed something ⬇️ , the test is failing in CI which means the query param is not available in tflocal.

agent_pool_integration_test.go
JarrettSpiker: @sebasslash sorry, took a vacation and let this review sit for a bit. The flag is now on in production, and it looks like we are seeing the tests pass!

mpminardi (approved):
This LGTM once the tests are passing (the one failure looks like a flake).

sebasslash (approved):
🔥 Looks good to me

PR #419, opened 2022-06-02
Testing the agent pool tests and their flakiness.

annawinkler:
## Description

The agent pool tests are one of the set of flaky tests we have. Upgrade the org to a biz account for the tests to work. I looked in atlas and nothing stood out to me to cause this issue, so I'm a little confused about why it's happening, but if we can create an agent pool, then the other tests will work (list, update, delete).

## Testing plan

1. Run the test: `go test --run TestAgentPoolsCreate -v ./... -tags=integration` and it should pass.

## Output from tests

```

...
```

### DIFF
--- a/agent_pool_integration_test.go
+++ b/agent_pool_integration_test.go
@@ -81,6 +81,8 @@ func TestAgentPoolsCreate(t *testing.T) {
 	orgTest, orgTestCleanup := createOrganization(t, client)
 	defer orgTestCleanup()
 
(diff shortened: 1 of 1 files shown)


brandonc (approved):

brandonc:
I wonder if we can get rid of `skipIfFree` if we just upgrade all the orgs in the paid feature tests.

annawinkler:
I agree - I don't understand how this would have ever worked.

PR #422, opened 2022-06-08
Add info re support for Private Registry Provider etc APIs

annawinkler:
## Description

Update changelog with addition of support for private registry provider, version, and platform APIs.


sebasslash (approved):
I love to see it 🔥

PR #425, opened 2022-06-10
add support for new organization-specific workspace limit setting

SwiftEngineer:
## Description

This change allows go-tfe to read/set/unset a newly created setting on the organization model. This setting is only available in TFE.

## Testing plan

1. Run local development version of TFE
2. Run the newly updated tests in [admin_organization_integration_test.go](admin_organization_integration_test.go)

# Output from tests

It's worth noting that I had to make a few changes to the existing test (specifically, assertions about the GlobalModuleSharing flag) in order to get these tests to pass. Since those tests don't get run in CI, I'm thinking they probably haven't run in a long while.

Looking at https://github.com/hashicorp/go-tfe/pull/278, I can't figure out how [the assertions](https://github.com/hashicorp/go-tfe/pull/278/files#diff-4527a001d099fd64c39587b6779f05a56736553af6d47bd1fd6fddddb8e90b25R197) could have passed, since `globalModuleSharing` was changed from a `bool` to `*bool`, shouldn't the assertions reflect that? Maybe the test library changed? I have no idea! 😅 This is my first PR for Golang, so if anyone could correct me here, I can happily revert my changes there and fix whatever is going on.

```
go tool test2json -t tmp/path-from/GoLand/___TestAdminOrganizations_Update_fetches_and_updates_organization_in_github_com_hashicorp_go_tfe.test -test.v -test.paniconexit0 -test.run ^\QTestAdminOrganizations_Update\E$/^\Qfetches_and_updates_organization\E$
=== RUN   TestAdminOrganizations_Update
--- PASS: TestAdminOrganizations_Update (3.37s)
=== RUN   TestAdminOrganizations_Update/fetches_and_updates_organization
    --- PASS: TestAdminOrganizations_Update/fetches_and_updates_organization (2.45s)
PASS
```

### DIFF
--- a/admin_organization_integration_test.go
+++ b/admin_organization_integration_test.go
@@ -106,6 +106,7 @@ func TestAdminOrganizations_Read(t *testing.T) {
 		assert.NotNilf(t, adminOrg.NotificationEmail, "NotificationEmail is not nil")
 		assert.NotNilf(t, adminOrg.SsoEnabled, "SsoEnabled is not nil")
 		assert.NotNilf(t, adminOrg.TerraformWorkerSudoEnabled, "TerraformWorkerSudoEnabledis not nil")
+		assert.Nilf(t, adminOrg.WorkspaceLimit, "WorkspaceLimit is nil")
 	})
 }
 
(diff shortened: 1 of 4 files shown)


nullable.go
```diff
@@ -0,0 +1,39 @@
...
+
+// int version of Nullable
+type NullableInt struct {
+	value *int
+}
+
+func (i NullableInt) MarshalJSON() ([]byte, error) {
+	return Nullable{i.value}.MarshalJSON()
+}
+
```
SwiftEngineer, line 31: I added a string version of nullable as well (even though it isn't in use right now) so that folks would have an easier time understanding how they might extend Nullable to work with their own types.

JarrettSpiker (dismissed):
LGTM.
Im assuming that 0 is handles separately from null on the backend, which is why we want the value to be either?

My only nitpick with that it is `nil` and not `null` in go. I guess it shows up as `null` in the JSON though, so 🤷

SwiftEngineer:
> LGTM. Im assuming that 0 is handles separately from null on the backend, which is why we want the value to be either?

Can confirm, that's a valid and true assumption 👍 

> My only nitpick with that it is `nil` and not `null` in go. I guess it shows up as `null` in the JSON though, so 🤷

Tell me about it. I did write a test to verify that at least!

docs/TESTS.md
```diff
@@ -12,7 +12,7 @@ Your registry module repository will need to be a [valid module](https://www.ter
 It will need the following: 
 1. To be named `terraform-<PROVIDER>-<NAME>`
 1. At least one valid SemVer tag in the format `x.y.z`
-[terraform-random-module](ttps://github.com/caseylang/terraform-random-module) is a good example repo. 
+[terraform-random-module](https://github.com/caseylang/terraform-random-module) is a good example repo.
```
SwiftEngineer, line 15: small little typo fix that I didn't think was worth a second PR, so I snuck it in here 🥷

nullable.go
```diff
@@ -0,0 +1,39 @@
...
+// Nullable acts as a wrapper to enable using "omitempty" on fields that should retain the ability to explicitly be set
+// to null.
+//
+// This is particularly useful for attributes that are set to null if provided as null in a json request body.
+// Conversely, if they are omitted from the request body, they will retain their existing value.
+//
+// This struct should use generics, but this project doesn't support them yet :(
+type Nullable struct {
+	value interface{}
+}
```
SwiftEngineer, line 14: All the research I did into this problem was pretty fruitless. The only helpful thing I learned is that the concept of an `Optional` is very "un-go-like", which certainly made me feel less great about this solution.

If anyone has any alternative suggestions, I'd be more than happy to hear them!

barrettclark:
@SwiftEngineer Would you mind squashing the commits once you're ready for this to be merged in?

SwiftEngineer:
> @SwiftEngineer Would you mind squashing the commits once you're ready for this to be merged in?

No sweat! will do

admin_organization_integration_test.go
```diff
@@ -238,39 +239,47 @@ func TestAdminOrganizations_Update(t *testing.T) {
...
 			IsDisabled:          &isDisabled,
+			WorkspaceLimit:      &workspaceLimit,
 		}
 
 		adminOrg, err = client.Admin.Organizations.Update(ctx, org.Name, opts)
 		assert.NoError(t, err)
 		assert.NotNilf(t, adminOrg, "Org returned as nil when it shouldn't be.")
 
-		assert.Equal(t, adminOrg.GlobalModuleSharing, globalModuleSharing)
+		assert.Equal(t, &globalModuleSharing, adminOrg.GlobalModuleSharing)
```
brandonc, line 279: I'm a little confused by the argument swapping here and on L241. The other assertions seem to put the parameter second and the model first. And also by the address operator. What's going on?

brandonc (dismissed):
Thanks! Be sure to update the changelog and it looks g2g

admin_organization_integration_test.go
brandonc: > It's worth noting that I had to make a few changes to the existing test (specifically, assertions about the GlobalModuleSharing flag) in order to get these tests to pass. Since those tests don't get run in CI, I'm thinking they probably haven't run in a long while.

OK ! Nevermind !

annawinkler:
Please squash the commits & then this looks great!

brandonc (approved):

PR #429, opened 2022-06-13
Add support for GPG Keys API

sebasslash:
Adds support for the [GPG Keys API ](https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys)

This PR also modifies the client to support the registry base path: `/api/registry`. To keep the client's internal interface consistent I've added an extra check in `newRequest()` that will parse the `path` variable; if a request is made to the Private Registry API the url must explicitly contain the base path so that the client can set the appropriate registry base path. As far as the TFC/E API is concerned, the GPG Key endpoints are the only ones that use the registry's default path. 

To run individual tests:

```sh
go test -run TestGPGKeyCreate -v ./... -tags=integration
go test -run TestGPGKeyRead -v ./... -tags=integration
go test -run TestGPGKeyUpdate -v ./... -tags=integration
go test -run TestGPGKeyDelete -v ./... -tags=integration
```


gpg_key.go
```diff
@@ -0,0 +1,184 @@
...
+	Type      string `jsonapi:"primary,gpg-keys"`
+	Namespace string `jsonapi:"attr,namespace"`
+}
+
+func (s *gpgKeys) Create(ctx context.Context, options GPGKeyCreateOptions) (*GPGKey, error) {
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := s.client.newRequest("POST", "/api/registry/private/v2/gpg-keys", &options)
```
annawinkler, line 72: When we talked about these URLs last week, did we say that we would create a new request type for the private registry API endpoint? 🤔 I didn't take notes so I'm not remembering unfortunately.

annawinkler (commented):
Thanks for adding support for this API! I've left a few comments and questions. ✨

gpg_key_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+			Namespace: provider2.Organization.Name,
+		}
+
+		updatedKey, err := client.GPGKeys.Update(ctx, keyID, opts)
+		require.NoError(t, err)
+
+		assert.Equal(t, gpgKey.KeyID, updatedKey.KeyID)
+		assert.Equal(t, updatedKey.Namespace, provider2.Organization.Name)
+
+		// Cleanup
```
annawinkler, line 139: 💯 Nice!

gpg_key_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+	t.Cleanup(orgCleanup)
+
+	upgradeOrganizationSubscription(t, client, org)
+
+	provider, providerCleanup := createRegistryProvider(t, client, org, PrivateRegistry)
+	t.Cleanup(providerCleanup)
+
+	gpgKey, _ := createGPGKey(t, client, org, provider)
+
+	t.Run("when a key exists", func(t *testing.T) {
```
annawinkler, line 162: What do you think about adding a test if the key doesn't exist or is invalid?

gpg_key.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	if err := keyID.valid(); err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("/api/registry/private/v2/gpg-keys/%s/%s",
+		url.QueryEscape(keyID.Namespace),
```
annawinkler, line 119: When I was reading the [docs](https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys#update-a-gpg-key), I noticed this line: `Only the namespace attribute can be updated, and namespace has to match an organization the user has permission to access.` I'm curious if the API returns an error if a namespace doesn't match an org the user has permission to access? Would this be something we would want to add a test for? The reason I suggest adding a test is because this detail could be overlooked by someone using this API, and they might see a test. I also admit that we don't want to cross a boundary if it doesn't make sense. What do you think? 🤔

gpg_key.go
sebasslash: Yep! but these improvements will be made after the fact since they are only internal changes.

gpg_key.go
```diff
@@ -0,0 +1,184 @@
...
+type GPGKeyID struct {
+	Namespace string
+	KeyID     string
+}
+
+// GPGKeyCreateOptions represents all the available options used to create a GPG key.
+type GPGKeyCreateOptions struct {
+	Type      string `jsonapi:"primary,gpg-keys"`
+	Namespace string `jsonapi:"attr,namespace"`
+	AsciiAmor string `jsonapi:"attr,ascii-armor"`
```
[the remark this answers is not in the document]
sebasslash: Ascii Love is not a bad field name ! 😆

gpg_key_integration_test.go
sebasslash: This will return a 404, however we don't need to add a test for this. The reason being is the client in `client.do()` decodes the headers and returns the appropriate generic error. In this case that would be `ErrResourceNotFound`.

gpg_key.go
sebasslash: Good catch and this makes way more sense here as we want to test for an error that's specific to this resource! I've added a check in the error handler for `client.do()` inside `GPGKeys.Update()` where if the body's message contains `namespace is not authorized` it will return `ErrNamespaceNotAuthorized`. I've also added the corresponding test.

gpg_key.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return g, nil
+}
+
+func (s *gpgKeys) Read(ctx context.Context, keyID GPGKeyID) (*GPGKey, error) {
+	if err := keyID.valid(); err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("/api/registry/private/v2/gpg-keys/%s/%s",
```
annawinkler, line 92: I have mixed thoughts on not taking a parameter for the `registry-name` (which must be `private`). The API [docs](https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys#get-gpg-key) specify this as a parameter, and this is a parameter for the private registry provider APIs. What are your thoughts about this @sebasslash ?

sebasslash: Hmm, well since it's always going to be `private`, I'd rather not expose it as part of the interface.

annawinkler (approved):
🥇 Looks great!

PR #433, opened 2022-06-16
temporarily disable flaky config versions test.

SwiftEngineer:
## Description

it will be re-enabled shortly! we just need a bit more time before we can fix it.


mpminardi (approved):

nfagerlund:
@radditude filled me in! Plan approved.

PR #439, opened 2022-06-21
Extend config interface to support setting `retryServerErrors`

sebasslash:
## Description

There were instances where users would experience a network failure during the metadata or `/ping` request when calling `NewClient()`. 

This PR adds a new field `RetryServerErrors` to the `Config` object, which sets the value for `c.retryServerErrors` during client initialization and has a default value of `false`. In short, the client exposes a receiver method `RetryServerErrors(bool)` which allows you to set the value of `c.retryServerErrors` which toggles the retry logic, however it requires an initialized client. This became a problem due to the network request made in `NewClient()` that wouldn't be retried should it fail -- there was no way for the user to enable the retry logic before calling `NewClient()`!

Making it an option in the client's configuration process will circumvent this issue.

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -66,17 +66,21 @@ type Config struct {
 
 	// RetryLogHook is invoked each time a request is retried.
 	RetryLogHook RetryLogHook
+
+	// RetryServerErrors enables the retry logic in the client.
+	RetryServerErrors bool
 }
 
 // DefaultConfig returns a default config structure.
 
 func DefaultConfig() *Config {
 	config := &Config{
-		Address:    os.Getenv("TFE_ADDRESS"),
-		BasePath:   DefaultBasePath,
-		Token:      os.Getenv("TFE_TOKEN"),
-		Headers:    make(http.Header),
-		HTTPClient: cleanhttp.DefaultPooledClient(),
+		Address:           os.Getenv("TFE_ADDRESS"),
+		BasePath:          DefaultBasePath,
+		Token:             os.Getenv("TFE_TOKEN"),
+		Headers:           make(http.Header),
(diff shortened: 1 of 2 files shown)


tfe.go
```diff
@@ -441,6 +441,9 @@ func (c *Client) getRawAPIMetadata() (rawAPIMetadata, error) {
 	req.Header.Set("Accept", "application/vnd.api+json")
 	req.Header.Set("Authorization", "Bearer "+c.token)
 
+	// Temporarily enable retryServerErrors in case the metadata request fails
+	c.retryServerErrors = true
+
 	// Make a single request to retrieve the rate limit headers.
 	resp, err := c.http.HTTPClient.Do(req)
 	if err != nil {
```
hashi-derek, line 449: Thank you for this change.

If it errors here, then it looks like it could get permanently stuck with the retry enabled. I don't know if `getRawAPIMetadata()` will be called somewhere else in the code in the future, but it could result in unexpected side-effects as the codebase grows.

sebasslash: I considered this initially but two things jumped out at me that made me hesitate:
- This method is not exported. 
- This method is only called once in go-tfe and that's in `NewClient()`. The user can only modify `c.retryServerErrors` using the exported receiver method `c.RetryServerErrors(bool)` and so we wouldn't be altering any set values since the client has not yet been created.

Uk1288 (approved):
LGTM! 🎉

brandonc (changes requested):
I don't think this does what it says in the description?

annawinkler (dismissed):
We reviewed this PR yesterday. Looks good! Please update the description accordingly. 🌟

brandonc (approved):

brandonc (approved):

PR #446, opened 2022-06-27
re-add configuration version test, which was fixed by adding another configuration version

SwiftEngineer:
## Description

This commit re-adds a test that was removed in the past due to a change in TFE, the issue was discovered by @radditude, who described it thusly:

> This is likely due to the [recent change](https://github.com/hashicorp/atlas/pull/12523) that prevents archiving a config version if it's the "current configuration version" for the workspace (aka the config version that's either been mostly recently uploaded, or the one that was used in the most recent run). We'll probably need to upload a second config version during that test so that the one we're trying to archive isn't the workspace's current config version.

## Output from tests

```
=== RUN   TestConfigurationVersionsArchive
--- PASS: TestConfigurationVersionsArchive (4.91s)
=== RUN   TestConfigurationVersionsArchive/when_the_configuration_version_exists_and_has_been_uploaded
    --- PASS: TestConfigurationVersionsArchive/when_the_configuration_version_exists_and_has_been_uploaded (2.02s)
=== RUN   TestConfigurationVersionsArchive/when_the_configuration_version_does_not_exist
    --- PASS: TestConfigurationVersionsArchive/when_the_configuration_version_does_not_exist (0.27s)
=== RUN   TestConfigurationVersionsArchive/with_invalid_configuration_version_id
    --- PASS: TestConfigurationVersionsArchive/with_invalid_configuration_version_id (0.00s)
```

---
  - https://app.asana.com/0/0/1202446447509971

### DIFF
--- a/configuration_version_integration_test.go
+++ b/configuration_version_integration_test.go
@@ -201,7 +201,7 @@ func TestConfigurationVersionsUpload(t *testing.T) {
 		)
 		require.NoError(t, err)
 
-		WaitUntilStatus(t, client, ctx, cv, ConfigurationUploaded)
+		WaitUntilStatus(t, client, cv, ConfigurationUploaded, 60)
 	})
 
 	t.Run("without a valid upload URL", func(t *testing.T) {
@@ -241,7 +241,7 @@ func TestConfigurationVersionsArchive(t *testing.T) {
 		)
 		require.NoError(t, err)
 
-		WaitUntilStatus(t, client, ctx, cv, ConfigurationUploaded)
+		WaitUntilStatus(t, client, cv, ConfigurationUploaded, 60)
 
 		// configuration version should not be archived, since it's the latest version
 		err = client.ConfigurationVersions.Archive(ctx, cv.ID)
@@ -257,12 +257,12 @@ func TestConfigurationVersionsArchive(t *testing.T) {
 		)
 		require.NoError(t, err)
 		defer newCvCleanup()
-		WaitUntilStatus(t, client, ctx, newCv, ConfigurationUploaded)
+		WaitUntilStatus(t, client, newCv, ConfigurationUploaded, 60)
 
 		err = client.ConfigurationVersions.Archive(ctx, cv.ID)
 		require.NoError(t, err)
 
-		WaitUntilStatus(t, client, ctx, cv, ConfigurationArchived)
+		WaitUntilStatus(t, client, cv, ConfigurationArchived, 60)
 	})
 
 	t.Run("when the configuration version does not exist", func(t *testing.T) {
@@ -353,22 +353,4 @@ func TestConfigurationVersions_Unmarshal(t *testing.T) {
 	assert.Equal(t, cv.Status, ConfigurationUploaded)
 	assert.Equal(t, cv.StatusTimestamps.FinishedAt, finishedParsedTime)
 	assert.Equal(t, cv.StatusTimestamps.StartedAt, startedParsedTime)
-}
-
-// helper to wait until a configuration version has reached a certain status
-func WaitUntilStatus(t *testing.T, client *Client, ctx context.Context, cv *ConfigurationVersion, desiredStatus ConfigurationStatus) {
-	for i := 0; ; i++ {
-		refreshed, err := client.ConfigurationVersions.Read(ctx, cv.ID)
-		require.NoError(t, err)
-
-		if refreshed.Status == desiredStatus {
-			break
-		}
-
-		if i > 30 {
-			t.Fatal("Timeout waiting for the configuration version to be archived")
-		}
-
-		time.Sleep(1 * time.Second)
-	}
 }
--- a/configuration_version_integration_test.go
+++ b/configuration_version_integration_test.go
@@ -257,7 +257,7 @@ func TestConfigurationVersionsArchive(t *testing.T) {
 		)
 		require.NoError(t, err)
 		defer newCvCleanup()
-		WaitUntilStatus(t, client, ctx, cv, ConfigurationUploaded)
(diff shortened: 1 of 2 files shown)


configuration_version_integration_test.go
```diff
@@ -242,6 +242,73 @@ func TestConfigurationVersionsArchive(t *testing.T) {
...
+			}
+
+			if i > 10 {
+				t.Fatal("Timeout waiting for the configuration version to be uploaded")
+			}
+
+			time.Sleep(1 * time.Second)
+		}
+
+		// configuration version should not be achievable, since it's the latest version
```
mpminardi, line 276: Nit: typo
```suggestion
		// configuration version should not be archived, since it's the latest version
```

configuration_version_integration_test.go
```diff
@@ -242,6 +242,73 @@ func TestConfigurationVersionsArchive(t *testing.T) {
...
+
+		// create subsequent version, since the latest configuration version cannot be archived
+		newCv, newCvCleanup := createConfigurationVersion(t, client, w)
+		err = client.ConfigurationVersions.Upload(
+			ctx,
+			newCv.UploadURL,
+			"test-fixtures/config-version",
+		)
+		require.NoError(t, err)
+		defer newCvCleanup()
```
mpminardi, line 289: I think we might want a similar loop here to what is done for the first uploaded configuration version for similar reasons (e.g., to make sure we attempt the archive only after it is actually uploaded).

configuration_version_integration_test.go
```diff
@@ -242,6 +242,73 @@ func TestConfigurationVersionsArchive(t *testing.T) {
...
+		for i := 0; ; i++ {
+			refreshed, err := client.ConfigurationVersions.Read(ctx, cv.ID)
+			require.NoError(t, err)
+
+			if refreshed.Status == ConfigurationUploaded {
+				break
+			}
+
+			if i > 10 {
+				t.Fatal("Timeout waiting for the configuration version to be uploaded")
```
annawinkler, line 270: I notice that there is a similar loop later in the code. Any chance of refactoring to avoid the duplication?

Also, I'm concerned about adding a test that might have flaky behavior. Specifically, I'm wondering where the 10 second wait comes from? What if we need 15 seconds for this to pass? Could other tests running in tflocal impact the performance of this test, for example?

annawinkler (changes requested):
Thank you for adding this test! I have a few questions about the test that I've added in a separate comment.

configuration_version_integration_test.go
SwiftEngineer: > I notice that there is a similar loop later in the code. Any chance of refactoring to avoid the duplication?

Fair enough! I'll deduplicate the code now.

> Also, I'm concerned about adding a test that might have flaky behavior. Specifically, I'm wondering where the 10 second wait comes from? What if we need 15 seconds for this to pass? Could other tests running in tflocal impact the performance of this test, for example?

Also fair! Unfortunately, I don't know anything we can do here besides increasing the timeout to something absurd like a full minute maybe? That feels super icky to me, but for this specific case it doesn't seem like we have any other alternatives, unless you're advocating for keeping the test deleted? I'm 50/50, heck I wasn't even the person who originally wrote this, so I imagine whoever that is they'd probably be the ones to ask.

SwiftEngineer: It'd be awesome if we could somehow materialize the promise to archive/upload the config versions so we could avoid all the polling, but since that archiving/uploading happens on a background job, we're kinda hosed there 😢

annawinkler: You're right - a minute doesn't seem awesome. What do you think about adding a comment where the timeout is so future us knows that if we see some flakiness there it might be that we need a longer timeout?

configuration_version_integration_test.go
```diff
@@ -331,3 +354,21 @@ func TestConfigurationVersions_Unmarshal(t *testing.T) {
 	assert.Equal(t, cv.StatusTimestamps.FinishedAt, finishedParsedTime)
 	assert.Equal(t, cv.StatusTimestamps.StartedAt, startedParsedTime)
 }
+
+// helper to wait until a configuration version has reached a certain status
+func WaitUntilStatus(t *testing.T, client *Client, ctx context.Context, cv *ConfigurationVersion, desiredStatus ConfigurationStatus) {
```
annawinkler, line 359: 💞 I like this refactor and I love how the tests read now! 

I'm still new to this repo, so I'm going to ping @brandonc and @sebasslash because I'm not sure what we typically do with test helper methods. In this case, I'm not sure it makes sense to put it in helper_test.go, but that seems like the pattern we have used in the past. 🤔

configuration_version_integration_test.go
SwiftEngineer: Good catch 👍 Added

helper_test.go
```diff
@@ -174,26 +175,29 @@ func createUploadedConfigurationVersion(t *testing.T, client *Client, w *Workspa
 		t.Fatal(err)
 	}
 
+	WaitUntilStatus(t, client, cv, ConfigurationUploaded, 15)
+
+	return cv, cvCleanup
+}
+
+// helper to wait until a configuration version has reached a certain status
+func WaitUntilStatus(t *testing.T, client *Client, cv *ConfigurationVersion, desiredStatus ConfigurationStatus, timeoutSeconds int) {
```
annawinkler, line 184: One minor suggestion - I don't think we need to export the function to other packages, so it could instead be `waitUntilStatus`

annawinkler (commented):
Once the commits are squashed I think this looks good to go! 🚀

annawinkler (approved):
Unblocking this PR since I'm off tomorrow. Please squash the commits before merging. Thank you! 💖

PR #450, opened 2022-06-29
Update run-tests to large resource class in config.yml

annawinkler:
## Description

With successful runs taking about 20 minutes in CircleCI, I'd like to see if we get any benefit from a larger resource class in CircleCI.

## Testing plan

1. Run the build in CircleCI and see how long it takes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,14 +22,15 @@ jobs:
       - run:
           name: Run tests
           command: |
-              export GO111MODULE=on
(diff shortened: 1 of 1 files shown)


annawinkler:
Nope! Verified it takes the same amount of time, so we're good sticking with a medium resource class. See the screenshot, and the build at the bottom to compare build times.

Screen Shot 2022-06-30 at 10 19 46 AM

PR #451, opened 2022-06-30
Add Terraform Datasource for TFC Agents

laurenolivia:
## Output from tests

```

...
```

---
  - https://app.asana.com/0/0/1202267776134371

### DIFF
--- a/agent_pool.go
+++ b/agent_pool.go
@@ -31,6 +31,9 @@ type AgentPools interface {
 
 	// Delete an agent pool by its ID.
 	Delete(ctx context.Context, agentPoolID string) error
+
+	// List ip-address by agent pool ID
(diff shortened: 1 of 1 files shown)


agent_pool.go
```diff
@@ -214,6 +217,28 @@ func (s *agentPools) Delete(ctx context.Context, agentPoolID string) error {
...
+		return nil, ErrInvalidAgentPoolID
+	}
+
+	u := fmt.Sprintf("agent-pools/%s/agents", url.QueryEscape(agentPoolID))
+	req, err := s.client.newRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	poolList := &AgentPoolList{}
```
Uk1288, line 232: The IP address attribute mapping should be added to the `AgentPool` type in `AgentPoolList` since the intent is to retrieve the IPs. Good progress 🔥

PR #455, opened 2022-07-05
Tests: Wait for async deletion APIs

JarrettSpiker:
## Description

We are seem some intermittent test failures on the TestOAuthClientsDelete/with_valid_options test, which I am pretty sure is because the Delete API for OAuth Clients on TFC [has become asynchronous ](https://github.com/hashicorp/atlas/commit/b586f5b8da939d8dfe0a8b6ab7f50db214469a13) which the test doesnt account for.

I am pretty sure that the same is true for workspace deletion.

This adds a retries to the checks that deletion has completed for those APIs. The existing `retry` function used will make 5 attempts, with a space of 2s between attempts, which should be plenty of time to allow the async deletion to complete server side

## Testing plan

Run the tests locally, and check that the failures go away. Also check that errors TFC-side still cause the tests to fail

### DIFF
--- a/admin_workspace_integration_test.go
+++ b/admin_workspace_integration_test.go
@@ -190,10 +190,19 @@ func TestAdminWorkspaces_Delete(t *testing.T) {
 		err = client.Admin.Workspaces.Delete(ctx, adminWorkspace.ID)
 		assert.NoError(t, err)
 
+		// Admin Workspace client deletion can happen asynchronously
+		// Retry the read a few times until we get an error
+		workspaceReadError, _ := retry(func() (interface{}, error) {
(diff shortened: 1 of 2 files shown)


JarrettSpiker:
> 🌟 Thank you for this PR! Would we still expect to see the errors in the build with this update? 🤔

No, we should not :( It should have fixed the exact error which is appearing in the CircleCI failure, even though it has completely eliminated the error locally. I am trying to figure that out...

JarrettSpiker:
https://github.com/hashicorp/atlas/pull/12820 is the more correct solution for the error we are seeing, so I am going to close this PR

PR #456, opened 2022-07-06
add agent data source

laurenolivia:
## Description

This PR adds a new data source to the Client. The purpose is to allow users to Read, List agents. Please see the Agent Pool and Agents documentation, linked below, for more information. 

Please be aware that this PR downloads the [tfc-agent binary](https://releases.hashicorp.com/tfc-agent/), which only supports Linux AMD64 architecture.

## Testing plan

1. Checkout this branch
2. Ensure you have an existing agent pool (Org -> Settings -> Agent Pool)
3. Ensure you have an agent pool token
4. Create an agent
5. Make an API call to read, list an agent

-- or --

1. With the required architecture
2. Run the test command noted, below
3. The test suite should pass ✅ 

## External links

- [Agent Pools and Agents API documentation](https://www.terraform.io/cloud-docs/api-docs/agents#agent-pools-and-agents-api)

## Output from tests
PASS
ok  	github.com/hashicorp/go-tfe

See example of how to run integration tests, below. Refer to [TESTS.md](https://github.com/hashicorp/go-tfe/blob/main/docs/TESTS.md) for further details.

```
envchain tfe.local go test ./... -v -tags=integration -run TestAgents

...
```

---
  - https://app.asana.com/0/0/1202267776134371


agent.go
```diff
@@ -0,0 +1,138 @@
...
+// TFE API docs: https://www.terraform.io/docs/cloud/api/agents.html
+type Agents interface {
+	// Read an agent by its ID.
+	Read(ctx context.Context, agentID string) (*Agent, error)
+
+	// Read an agent by its ID with the given options.
+	ReadWithOptions(ctx context.Context, agentID string, options *AgentReadOptions) (*Agent, error)
+
+	// List all the agents of the given pool.
+	List(ctx context.Context, agentPoolID string, options *AgentPoolListOptions) (*AgentList, error)
```
laurenolivia, line 23: Updated `AgentPoolListOptions` to `AgentListOptions`

agent.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Read a single agent by its ID
+func (s *agents) Read(ctx context.Context, agentID string) (*Agent, error) {
+	return s.ReadWithOptions(ctx, agentID, nil)
+}
+
+// Read a single agent by its ID with options.
+func (s *agents) ReadWithOptions(ctx context.Context, agentID string, options *AgentReadOptions) (*Agent, error) {
+	if !validStringID(&agentID) {
+		return nil, ErrInvalidAgentID //undeclared var name
```
sebasslash, line 77:
```suggestion
		return nil, ErrInvalidAgentID
```

agent.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, err
+	}
+
+	agent := &Agent{}
+	err = req.Do(ctx, agent)
+	if err != nil {
+		return nil, err
+	}
+
+	return agent, nil //cannot use agent as *Agent value in return statement
```
sebasslash, line 95: Another comment to remove? 🧹

agent.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *agents) ReadWithOptions(ctx context.Context, agentID string, options *AgentReadOptions) (*Agent, error) {
+	if !validStringID(&agentID) {
+		return nil, ErrInvalidAgentID //undeclared var name
+	}
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("agents/%s", url.QueryEscape(agentID))
+	req, err := s.client.NewRequest("GET", u, nil)
```
sebasslash, line 84: I think we should pass `options` here instead of `nil`.

agent_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func TestAgentsRead(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	agent, _, agentCleanup := createAgent(t, client, nil, nil, nil)
+	defer agentCleanup()
+	t.Log("log agent: ", agent)
+	// t.Log("log pool: ", pool)
```
sebasslash, line 21:
```suggestion
```

agent_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAgentsRead(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	agent, _, agentCleanup := createAgent(t, client, nil, nil, nil)
+	defer agentCleanup()
```
sebasslash, line 19: We've begun using `t.Cleanup(agentCleanup)` in favor of `defer`. `t.Cleanup()` is useful for cleaning up resources that are allocated by helper functions since it ensures all subsequent subtests have finished running before invoking.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
 	return _testAccountDetails
 }
 
+func createAgent(t *testing.T, client *Client, org *Organization, agentPool *AgentPool, agentPoolToken *AgentToken) (*Agent, *AgentPool, func()) {
```
sebasslash, line 92: Okay so this method signature is a bit long (apologies for the pickiness):

So I understand you've expanded the interface so that you wouldn't have to call:

```go
agentPool, agentPoolCleanup := createAgentPool(stuff, here)
```

within each test. But if I'm reading for the first time what this helper method is doing: 
```go
agent, _, agentCleanup := createAgent(t, client, nil, nil, nil)
```
I wouldn't really be sure what's going on, and I also wouldn't be able to deduce that `createAgent()` also creates an agent pool. 

Here's what I'm proposing:

```go
func createAgent(t *testing.T, client *Client, org *Organization, pool *AgentPool) (*Agent, func()) {}
```

We don't need to pass an `AgentToken` since we only ever use it upon creation. I guess there may be some subsequent test added to validate the token, but I don't think that falls under the scope of the Agents API (we know a token is valid if the agent is registered).

So we can simply create a new token inside of `createAgent()` and include its deletion in the cleanup function. 

The second thing is to remove `AgentPool` from the return value. If I wanted to use the `AgentPool`'s values in tests, I should create the agent pool beforehand and pass it into `createAgent()` otherwise why would we allow passing an agent pool as a parameter if we can simply use the one the method creates?

sebasslash (changes requested):
Some items for discussion down below ⬇️ . Great work so far 👍

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+		"--env", "TFC_ADDRESS="+DefaultConfig().Address,
+		"docker.mirror.hashicorp.services/hashicorp/tfc-agent:latest")
+
+	go func() {
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			t.Logf("Could not run container: %s", err)
+		}
+
+		t.Log("Logging container output: ", (string)(output))
```
nfagerlund, line 127: With `--detach`, I believe `run` just returns the unique container ID, unless it fails and prints an error message instead. It wouldn't actually log any container output.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+	}()
+
+	defer func() {
+		t.Log("Cleaning up agent docker container: ")
+		cmd := exec.Command("docker", "rm", "-f")
+		_ = cmd.Run()
+	}()
+
+	i, err := retry(func() (interface{}, error) {
```
nfagerlund, line 136: This is going to be vulnerable to the same impatient flake conditions you experienced from the existing tests while working on your other PR, since it uses the same hardcoded 12-second limit in `retry`. If it ends up being convenient, you might want to rebase onto the branch where Brandon and/or Luces started revising that behavior.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+
+		agentList, err := client.Agents.List(ctx, agentPool.ID, nil)
```
nfagerlund, line 138: It looks like there's an unspoken contract here: you can't ever use the same AgentPool for two concurrent executions of this function, otherwise you might get incoherent results depending on the test parallelism. Since we allow the caller to pass in their own agent pool, it might be worth stating that contract up in the function doc comment.

nfagerlund (changes requested):
Impressive work! I think I spotted some stuff that isn't right, plus a few things that _might_ not be intended.

agent.go
laurenolivia: @sebasslash 👍 
@nfagerlund Can you elaborate on why it's a sign that the method, possibly, isn't needed❔

agent_integration_test.go
laurenolivia: @sebasslash Got it. Thanks for sharing that knowledge.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+
+	if org == nil {
+		org, orgCleanup = createOrganization(t, client)
+	}
+
+	upgradeOrganizationSubscription(t, client, org)
+
+	if agentPool == nil {
+		agentPool, agentPoolCleanup = createAgentPool(t, client, org)
+		t.Log("create, log agentPool: ", agentPool)
```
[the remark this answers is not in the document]
laurenolivia: Nope. Removed.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+			t.Logf("Could not run container: %s", err)
+		}
+
+		t.Log("Logging container output: ", (string)(output))
+	}()
+
+	defer func() {
+		t.Log("Cleaning up agent docker container: ")
+		cmd := exec.Command("docker", "rm", "-f")
+		_ = cmd.Run()
```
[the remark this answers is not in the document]
laurenolivia: This was for testing purposes only, when the container was not behaving as expected. Will remove.

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+		}
+		return nil, errors.New("No agent found.")
+	})
+
+	if err != nil {
+		t.Fatalf("Could not return an agent %s", err)
+	}
+
+	agent = i.(*Agent)
+	t.Log("log agent, after type assertion: ", agent)
```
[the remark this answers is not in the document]
laurenolivia: No. Removed.

agent_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+		assert.Nil(t, k)
+		assert.EqualError(t, err, ErrInvalidAgentID.Error())
+	})
+}
+
+func TestAgentsList(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	agent, agentPool, agentCleanup := createAgent(t, client, nil, nil, nil)
```
laurenolivia, line 46: @sebasslash I'm curious what you think here. Per your recommendation not to return *AgentPool in the createAgent helper fn, when you get here, you want to pass an agent pool that is connected to the agent created in the helper. Does that make sense?

agent.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Agents describes all the agent-related methods that the
+// Terraform Cloud API supports.
+// TFE API docs: https://www.terraform.io/docs/cloud/api/agents.html
+type Agents interface {
+	// Read an agent by its ID.
+	Read(ctx context.Context, agentID string) (*Agent, error)
+
+	// Read an agent by its ID with the given options.
+	ReadWithOptions(ctx context.Context, agentID string, options *AgentReadOptions) (*Agent, error)
```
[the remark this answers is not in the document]
laurenolivia: This `ReadWithOptions` structure seems ubiquitous, so I based the decision on previous examples, but am open to removing. I'm unsure as to whether there is a general policy of always having a "with options" variant.

@sebasslash what do you think?

helper_test.go
```diff
@@ -88,6 +89,85 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
...
+
+	ctx := context.Background()
+	cmd := exec.Command("docker",
+		"run", "-d",
+		"--env", "TFC_AGENT_TOKEN="+agentPoolToken.Token,
+		"--env", "TFC_AGENT_NAME="+"this-is-a-test-agent",
+		"--env", "TFC_ADDRESS="+DefaultConfig().Address,
+		"docker.mirror.hashicorp.services/hashicorp/tfc-agent:latest")
+
+	go func() {
```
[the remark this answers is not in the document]
laurenolivia: This implementation was recommended to me, as feedback. What do you recommend in place of a goroutine? @nfagerlund

agent_integration_test.go
laurenolivia: Nvm! We chatted!

helper_test.go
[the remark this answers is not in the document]
taiidani: If you're referring to me I actually [recommended the opposite](https://hashicorp.slack.com/archives/C01BR9FR8KZ/p1658428391943639?thread_ts=1657747733.956509&cid=C01BR9FR8KZ) 😅 

> One more thing to pile on -- since you're specifying the -d (detach) property in your docker run command you don't need to spin up a goroutine on the cmd.CombinedOutput. That command will finish as soon as the container has been created.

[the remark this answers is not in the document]
laurenolivia: Lol no, not you @taiidani. Thanks for linking your note. Working through lots of feedback.

agent.go
```diff
@@ -0,0 +1,89 @@
...
+	Name string `jsonapi:"attr,name"`
+	IP   string `jsonapi:"attr,ip-address"`
+
+	// Relations
+	Organization *Organization `jsonapi:"relation,organization"`
+	Workspaces   []*Workspace  `jsonapi:"relation,workspaces"`
+}
+
+type AgentListOptions struct {
+	ListOptions
```
brandonc, line 46: Missing `filter[last-ping-since]` parameter https://www.terraform.io/cloud-docs/api-docs/agents#query-parameters-1

agent.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Agent represents a Terraform Cloud agent.
+type Agent struct {
+	ID   string `jsonapi:"primary,agents"`
+	Name string `jsonapi:"attr,name"`
+	IP   string `jsonapi:"attr,ip-address"`
+
+	// Relations
+	Organization *Organization `jsonapi:"relation,organization"`
+	Workspaces   []*Workspace  `jsonapi:"relation,workspaces"`
```
brandonc, line 42: I think these can be removed because these relations aren't available on agents according to the docs

agent.go
```diff
@@ -0,0 +1,89 @@
...
+type AgentList struct {
+	*Pagination
+	Items []*Agent
+}
+
+// Agent represents a Terraform Cloud agent.
+type Agent struct {
+	ID   string `jsonapi:"primary,agents"`
+	Name string `jsonapi:"attr,name"`
+	IP   string `jsonapi:"attr,ip-address"`
```
brandonc, line 38: missing attributes: status, last-ping-at

agent_integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAgentsRead(t *testing.T) {
+	//skipIfFreeOnly(t)
```
brandonc, line 15: ✂️

agent_integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	t.Run("without a valid agent ID", func(t *testing.T) {
+		k, err := client.Agents.Read(ctx, badIdentifier)
+		assert.Nil(t, k)
+		assert.EqualError(t, err, ErrInvalidAgentID.Error())
+	})
+}
+
+func TestAgentsList(t *testing.T) {
+	//skipIfFreeOnly(t)
```
brandonc, line 51: ✂️

agent_integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	org, orgCleanup := createOrganization(t, client)
+	t.Cleanup(orgCleanup)
+
+	upgradeOrganizationSubscription(t, client, org)
+
+	agentPool, agentPoolCleanup := createAgentPool(t, client, org)
+	t.Cleanup(agentPoolCleanup)
+
+	agent, agentCleanup := createAgent(t, client, org, agentPool)
```
brandonc, line 28: I think we can circumvent the agentPool reuse problem that @nfagerlund highlighted by always creating a new agentpool within createAgent and returning the pool, the agent, and the cleanup function.

brandonc (changes requested):
Be sure to add this file to the mocks script and then generate mocks

brandonc:
Tests passing 🎉

agent_integration_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	client := testClient(t)
+	ctx := context.Background()
+
+	org, orgCleanup := createOrganization(t, client)
+	t.Cleanup(orgCleanup)
+	upgradeOrganizationSubscription(t, client, org)
+
+	agentPool, agentPoolCleanup := createAgentPool(t, client, org)
+	t.Cleanup(agentPoolCleanup)
```
brandonc, line 57: This looks redundant!

helper_test.go
```diff
@@ -86,6 +93,188 @@ func fetchTestAccountDetails(t *testing.T, client *Client) *TestAccountDetails {
 	return _testAccountDetails
 }
 
+func downloadFile(filepath string, url string) error {
+
```
brandonc, line 97: 💅 extra linebreak

laurenolivia: Got it!

agent.go
```diff
@@ -0,0 +1,90 @@
...
+	IP         string `jsonapi:"attr,ip-address"`
+	Status     string `jsonapi:"attr,status"`
+	LastPingAt string `jsonapi:"attr,last-ping-at"`
+}
+
+type AgentListOptions struct {
+	ListOptions
+
+	//Optional:
+	Filter string `url:"filter[last-ping-since],omitempty"`
```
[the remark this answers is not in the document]
brandonc: This field value is supposed to be a date in ISO8601 format (ex. 2020-08-11T10:41:23Z), not just any string. It would be helpful to know what was expected. Try this encoding instead:

LastPingSince        time.Time            `url:"filter[last-ping-since],omitempty,iso8601"`

brandonc (changes requested):
I made the change that I recommended and smoke tested it and it worked well. So close!

agent.go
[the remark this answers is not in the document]
laurenolivia: Lol that's actually a mistake. Thanks for catching!

brandonc (approved):
giphy-2

PR #459, opened 2022-07-08
Speculative Flag inaccurate on Configuration Version List 

straubt1:
## Description

When Listing Configuration Versions, the Speculative flag was always returning false.
It appears this was due to a trailing space in the struct.

`go-tfe` version 1.3.0

This change removes that space and allows the function call to return accurately if the CV is speculative.

## Testing plan

To replicate, create a CV that is marked as speculative, then list CV's for that workspace.

```go
	cv, err := client.ConfigurationVersions.List(ctx, w.ID, &tfe.ConfigurationVersionListOptions{
		ListOptions: tfe.ListOptions{PageSize: 100},
		Include:     []tfe.ConfigVerIncludeOpt{},
	})
```

The resulting `cv` variable will not be marked as Speculative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * [beta] Add support for triggering Workspace runs through matching Git tags [#434](https://github.com/hashicorp/go-tfe/pull/434)
 
+## Bug fixes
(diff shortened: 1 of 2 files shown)


sebasslash (approved):
Good catch 👍 . Approved !

sebasslash (approved):
🏇 💨

PR #466, opened 2022-07-15
Add Name to the Oauth Client Payload

### LINKED ISSUE #461: OAuthClient missing Name attribute
The [oauth-clients endpoint](https://www.terraform.io/cloud-docs/api-docs/oauth-clients#sample-response) returns the `name` attribute of each OAuth Client, but the [OAuthClient struct](https://github.com/hashicorp/go-tfe/blob/main/oauth_client.go#L67-L83) doesn't expose this attribute. Without a human-friendly name, there isn't a particularly reliable way for a user to distinguish between clients, making this endpoint a bit challenging to use in the current implementation.

barrettclark:
## Description
The [oauth-clients endpoint](https://www.terraform.io/cloud-docs/api-docs/oauth-clients#sample-response) returns the name attribute of each OAuth Client, but the [OAuthClient struct](https://github.com/hashicorp/go-tfe/blob/main/oauth_client.go#L67-L83) doesn't expose this attribute. This pull request adds the Name field to the OAuthClient struct.

Fixes #461

## Testing plan

This is how you see the name in the Oauth Client payload:
1.  Make sure you have a GitHub [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
1.  Make sure you have a TF API token that has admin access
1.  Configure a [GitHub Oauth linkage](https://www.terraform.io/cloud-docs/vcs/github). If you give the Oauth linkage a name, you will see that name in the API response in the next step.
2. Now you can fetch the [list of OAuth Clients from the API](https://www.terraform.io/cloud-docs/api-docs/oauth-clients#sample-response)

In order to run the integration test you also need to do the first two steps from above.

## Output from tests

```
$ GITHUB_TOKEN=$GITHUB_TOKEN TF_ACC=1 TFE_TOKEN=$TFE_TOKEN TFE_ADDRESS=$TFE_ADDRESS go test -v -run TestOAuthClientsCreate ./... --tags=integration
```

---
  - https://app.asana.com/0/0/1202588229476830

### DIFF
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -73,6 +73,7 @@ type OAuthClient struct {
 	HTTPURL             string              `jsonapi:"attr,http-url"`
 	Key                 string              `jsonapi:"attr,key"`
 	RSAPublicKey        string              `jsonapi:"attr,rsa-public-key"`
+	Name                *string             `jsonapi:"attr,name"`
 	Secret              string              `jsonapi:"attr,secret"`
 	ServiceProvider     ServiceProviderType `jsonapi:"attr,service-provider"`
 	ServiceProviderName string              `jsonapi:"attr,service-provider-display-name"`
--- a/oauth_client_integration_test.go
+++ b/oauth_client_integration_test.go
@@ -109,6 +109,7 @@ func TestOAuthClientsCreate(t *testing.T) {
 		oc, err := client.OAuthClients.Create(ctx, orgTest.Name, options)
 		require.NoError(t, err)
 		assert.NotEmpty(t, oc.ID)
+		assert.Nil(t, oc.Name)
 		assert.Equal(t, "https://api.github.com", oc.APIURL)
 		assert.Equal(t, "https://github.com", oc.HTTPURL)
 		assert.Equal(t, 1, len(oc.OAuthTokens))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+# v1.6.0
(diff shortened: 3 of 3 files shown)


oauth_client.go
```diff
@@ -73,6 +73,7 @@ type OAuthClient struct {
 	HTTPURL             string              `jsonapi:"attr,http-url"`
 	Key                 string              `jsonapi:"attr,key"`
 	RSAPublicKey        string              `jsonapi:"attr,rsa-public-key"`
+	Name                string              `jsonapi:"attr,name"`
```
sebasslash, line 76: Given the default value of this field is `null`, we should make this a pointer.

oauth_client_integration_test.go
```diff
@@ -109,6 +109,7 @@ func TestOAuthClientsCreate(t *testing.T) {
 		oc, err := client.OAuthClients.Create(ctx, orgTest.Name, options)
 		require.NoError(t, err)
 		assert.NotEmpty(t, oc.ID)
+		assert.Empty(t, oc.Name)
```
sebasslash, line 112: `assert.Nil()` 💅

sebasslash (changes requested):
Really minor ⬇️ . Don't forget to include a changelog entry and given the OAuthClient interface has changed, you'll want to run `./generate_mocks.sh`.

oauth_client_integration_test.go
barrettclark: It's actually not `nil` - I thought it would be, but it comes through as `""`.

sebasslash (approved):
Looks good to me 👍  -- I was incorrect though, you only need to regenerate mocks when a method signature changes.

PR #468, opened 2022-07-16
Allow FileTriggersEnabled to be set to false when Git tags are present

hashimoon:
## Description

File triggers must be disabled when enabling Git tags, so both `FileTriggersEnabled` and `TagsRegex` need to support being configured at the same time.

### Creating A Workspace
1.  Create a `WorkspaceCreateOptions` struct
1. Set `FileTriggersEnabled` to `true`
1. Create a `VCSRepoOptions` struct
1. Set `TagsRegex` to "v-1"
1. On the `WorkspaceCreateOptions` struct, set `VCSRepo` equal to `VCSRepoOptions`
1. Verify that an exception is raised
1. Set `FileTriggersEnabled` to `false`
1. Verify an exception is not raised

### Updating A Workspace
1.  Create a `WorkspaceUpdateOptions` struct
1. Set `FileTriggersEnabled` to `true`
1. Create a `VCSRepoOptions` struct
1. Set `TagsRegex` to "v-1"
1. On the `WorkspaceUpdateOptions` struct, set `VCSRepo` equal to `VCSRepoOptions`
1. Verify that an exception is raised
1. Set `FileTriggersEnabled` to `false`
1. Verify an exception is not raised

## Output from tests

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Adds additional Task Stage and Run Statuses for Pre-plan run tasks by @glennsarti [#469](https://github.com/hashicorp/go-tfe/pull/469)
 * Adds `stage` field to the create and update methods for Workspace Run Tasks by @glennsarti [#469](https://github.com/hashicorp/go-tfe/pull/469)
 * Adds `ResourcesProcessed`, `StateVersion`, `TerraformVersion`, `Modules`, `Providers`, and `Resources` fields to the State Version struct by @laurenolivia [#484](https://github.com/hashicorp/go-tfe/pull/484)
+* Add support for disabled file trigger runs when setting Git tags [#468](https://github.com/hashicorp/go-tfe/pull/468)
 
 # v1.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Adds additional Task Stage and Run Statuses for Pre-plan run tasks by @glennsarti [#469](https://github.com/hashicorp/go-tfe/pull/469)
 * Adds `stage` field to the create and update methods for Workspace Run Tasks by @glennsarti [#469](https://github.com/hashicorp/go-tfe/pull/469)
 * Adds `ResourcesProcessed`, `StateVersion`, `TerraformVersion`, `Modules`, `Providers`, and `Resources` fields to the State Version struct by @laurenolivia [#484](https://github.com/hashicorp/go-tfe/pull/484)
-* Add support for disabled file trigger runs when setting Git tags [#468](https://github.com/hashicorp/go-tfe/pull/468)
+* Add support for disabled file trigger runs when setting Git tags by @hashimoon [#468](https://github.com/hashicorp/go-tfe/pull/468)
 
 # v1.6.0
 
--- a/workspace.go
+++ b/workspace.go
@@ -1080,7 +1080,7 @@ func (o WorkspaceCreateOptions) valid() error {
 		o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 {
 		return ErrUnsupportedBothTagsRegexAndTriggerPrefixes
 	}
-	if o.VCSRepo != nil && o.VCSRepo.TagsRegex != nil &&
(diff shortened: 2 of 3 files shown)


workspace.go
```diff
@@ -1080,8 +1080,8 @@ func (o WorkspaceCreateOptions) valid() error {
 		o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 {
 		return ErrUnsupportedBothTagsRegexAndTriggerPrefixes
 	}
-	if o.VCSRepo != nil && o.VCSRepo.TagsRegex != nil &&
-		o.FileTriggersEnabled != nil && *o.FileTriggersEnabled {
+	if o.VCSRepo != nil && o.VCSRepo.TagsRegex != nil && *o.VCSRepo.TagsRegex != "" &&
+		o.FileTriggersEnabled != nil && *o.FileTriggersEnabled == true {
```
matejrisek, line 1084: Could we not just leave it as `*o.FileTriggersEnabled`

workspace_integration_test.go
```diff
@@ -414,6 +413,18 @@ func TestWorkspacesCreate(t *testing.T) {
 		assert.EqualError(t, err, ErrUnsupportedBothTagsRegexAndFileTriggersEnabled.Error())
 	})
 
+	t.Run("when options include both non-empty tags-regex and file-triggers-enabled as false an error is not returned", func(t *testing.T) {
```
matejrisek, line 416: suggestion: We could use table driven tests and merge together this and the above 3 tests

workspace_integration_test.go
```diff
@@ -890,6 +900,18 @@ func TestWorkspacesUpdate(t *testing.T) {
 		assert.EqualError(t, err, ErrUnsupportedBothTagsRegexAndFileTriggersEnabled.Error())
 	})
 
+	t.Run("when options include both non-empty tags-regex and file-triggers-enabled an error is returned", func(t *testing.T) {
```
matejrisek, line 903: And this one could be a table driven test

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # v1.6.0
 * Add `Name` field to `OAuthClient` by @barrettclark [#466](https://github.com/hashicorp/go-tfe/pull/466)
 * Add `ConfigurationSourceAdo` configuration source option by @mjyocca [#467](https://github.com/hashicorp/go-tfe/pull/467)
+* Add support for disabled file trigger runs when setting Git tags [#468](https://github.com/hashicorp/go-tfe/pull/468)
```
annawinkler, line 6:
```suggestion
* Add support for disabled file trigger runs when setting Git tags by @hashimoon [#468](https://github.com/hashicorp/go-tfe/pull/468)
```

workspace.go
```diff
@@ -1080,7 +1080,7 @@ func (o WorkspaceCreateOptions) valid() error {
 		o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 {
 		return ErrUnsupportedBothTagsRegexAndTriggerPrefixes
 	}
-	if o.VCSRepo != nil && o.VCSRepo.TagsRegex != nil &&
+	if o.VCSRepo != nil && o.VCSRepo.TagsRegex != nil && *o.VCSRepo.TagsRegex != "" &&
```
annawinkler, line 1083: In looking at this if expression and the similar one when checking the update options, I was thinking it might be clearer to refactor the expression to a helper method that could be called in both places. What do you think?

workspace_integration_test.go
annawinkler: @hashimoon We're good with adding table driven tests here - agree that it would help readability.

hashimoon: Given the rest of the file is not utilizing table driven tests, can we add them in a future refactor? @annawinkler

hashimoon: Given the rest of the file is not utilizing table driven tests and the style is consistent with what is already present, can we add them in a future refactor? @annawinkler

annawinkler (approved):
:shipit: I'm OK with refactoring the tests in a separate PR.

PR #474, opened 2022-07-22
Initial tflocal configuration for CI

sebasslash:
This configuration makes use of the tflocal module that provisions the necessary TFE installation and AWS resources to run CI.

Once merged, I can connect the go-tfe repo to our TFC workspace where this configuration will live.

PR #475, opened 2022-07-22
Workflow job that rebuilds tflocal

sebasslash:
Adds the workflow action that runs a Go script that triggers a run to rebuild the tflocal instance. This workflow is run nightly but can also be used to refresh an unhealthy instance.


.github/workflows/tflocal-instance.yml
```diff
@@ -0,0 +1,21 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+
+      - name: Rebuild tflocal 
+        env:
+          TFE_TOKEN: ${{ secrets.TF_WORKFLOW_TFLOCAL_CLOUD_TFC_TOKEN }}
+        run: |
+          cd .github/scripts/rebuild-tflocal
+          go run main.go hashicorp-v2 tflocal-go-tfe
```
brandonc, line 21: This workspace doesn't exist. Should it?

brandonc: 💅 btw I think you could remove the dependency on go 1.18 and go-tfe by using curl:

```
curl \
  --header "Authorization: Bearer ${{ secrets.TF_WORKFLOW_TFLOCAL_CLOUD_TFC_TOKEN }}" \
  --header "Content-Type: application/vnd.api+json" \
  --request POST \
  --data '{"data": {"attributes": {"is-destroy":false,"message":"Queued nightly from tflocal-cloud GH Actions via go-tfe","replace-addrs":["module.tflocal.module.tfbox.aws_instance.tfbox"]},"type":"runs","relationships": {"workspace": {"data": {"type": "workspaces","id": "ws-LLGHCr4SWy28wyGN"}}}}}' \
  https://app.terraform.io/api/v2/runs
```
I didn't try this but I think it would work

sebasslash: 1) Yeah it exists now, I had to wait for #474 to be merged before creating it.

2) The only reason I think we should keep it so we can easily extend the `build-tflocal` script to modify certain workspace specific variables, e.g `env` (so we can toggle FFs as well as other env variables dynamically). Of course, we can do it solely using curl, but I find it harder to maintain -- But that's a personal opinion.

PR #479, opened 2022-07-25
[Beta] Add detailed-type field to state_version_output.go

brandonc:
## Description

This beta field is conditionally provided by the API depending on feature flags. It is needed for an upcoming release of terraform, so I believe it should be released as beta.

## Testing plan

See https://github.com/hashicorp/terraform/pull/31507 to test this


annawinkler (approved):
🚀 Do we also want to update the Changelog with this new addition?

brandonc:
Thanks! A release is being prepped and I'll add a release note for it.

PR #482, opened 2022-07-25
Add new run creation attributes

sebasslash:
This PR adds three new fields to the `RunCreateOptions` struct:
* `allow-empty-apply`
* `terraform-version`
* `plan-only`

Reviewer Notes:
- Error validation is performed by the client when a run-specific terraform version is set and the run is not configured with `PlanOnly: true`. It will return `ErrVersionValidForPlanOnly` (personally not a huge fan of this name, so hopefully there may be some other naming options here) 
- The version itself is not validated by the client to follow semver... I decided to offload this to the API which also checks if the desired version is available to the organization. If however, we have strong opinions that the client should validate the version, we can always use the [go-version](https://github.com/hashicorp/go-version) package.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -304,4 +304,6 @@ var (
 	ErrInvalidAsciiArmor = errors.New("ascii armor is invalid")
 
 	ErrRequiredNamespace = errors.New("namespace is required for public registry")
+
+	ErrTerraformVersionValidForPlanOnly = errors.New("setting terraform-version is only valid when plan-only is set to true")
 )
--- a/run.go
+++ b/run.go
@@ -106,6 +106,7 @@ type Run struct {
 	ID                     string               `jsonapi:"primary,runs"`
 	Actions                *RunActions          `jsonapi:"attr,actions"`
 	AutoApply              bool                 `jsonapi:"attr,auto-apply,omitempty"`
+	AllowEmptyApply        bool                 `jsonapi:"attr,allow-empty-apply"`
 	CreatedAt              time.Time            `jsonapi:"attr,created-at,iso8601"`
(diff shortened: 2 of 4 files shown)


errors.go
```diff
@@ -304,4 +304,6 @@ var (
 	ErrInvalidAsciiArmor = errors.New("ascii armor is invalid")
 
 	ErrRequiredNamespace = errors.New("namespace is required for public registry")
+
+	ErrVersionValidForPlanOnly = errors.New("configuring a run specific terraform version is only valid for plan-only runs")
```
brandonc, line 308: I think we should be explicit here

```suggestion
	ErrTerraformVersionValidForPlanOnly = errors.New("setting terraform-version is only valid when plan-only is set to true")
```

run.go
```diff
@@ -233,6 +236,18 @@ type RunCreateOptions struct {
 	// https://jsonapi.org/format/#crud-creating
 	Type string `jsonapi:"primary,runs"`
 
+	// AllowEmptyApply specifies whether Terraform can apply the run even when the plan contains no changes.
+	// Often used to upgrade state after upgrading a workspace to a new terraform version.
+	AllowEmptyApply *bool `jsonapi:"attr,allow-empty-apply"`
```
brandonc, line 241: I think you should use `,omitempty` on these (instead of sending null)

brandonc (changes requested):
I think you made the right decision to defer terraform version validation to the API. Otherwise, just some minor feedback about omitempty usage.

brandonc (dismissed):

brandonc:
This will end up in 1.7.0 so please rebase 😁

brandonc (approved):

PR #483, opened 2022-07-26
Remove unnecessary skip if beta

matejrisek:
## Description

Just remove unnecessary beta flag for one test - since we've already create organization that will flip the feature flag on 

## Output from tests

```

...
```


lilincmu (approved):
Looks good to me!

brandonc (approved):

PR #484, opened 2022-07-26
expose additional fields in state versions struct

laurenolivia:
## Description

This PR exposes additional attributes in the State Versions API.

## External links

- [State Versions API](https://www.terraform.io/cloud-docs/api-docs/state-versions#state-versions-api)

## Output from tests

```

...
```


state_version.go
```diff
@@ -55,12 +55,18 @@ type StateVersionList struct {
...
+	DownloadURL        string                   `jsonapi:"attr,hosted-state-download-url"`
+	Serial             int64                    `jsonapi:"attr,serial"`
+	VCSCommitSHA       string                   `jsonapi:"attr,vcs-commit-sha"`
+	VCSCommitURL       string                   `jsonapi:"attr,vcs-commit-url"`
+	ResourcesProcessed bool                     `jsonapi:"attr,resources-processed"`
+	StateVersion       int                      `jsonapi:"attr,state-version"`
+	TerraformVersion   string                   `jsonapi:"attr,terraform-version"`
+	Modules            *StateVersionModules     `jsonapi:"attr,modules"`
+	Providers          *StateVersionProviders   `jsonapi:"attr,providers"`
+	Resources          []*StateVersionResources `jsonapi:"attr,resources"`
```
brandonc, line 69: Can these be tested?

laurenolivia: @brandonc I can confirm the new attributes are exposed in the response body. Is that sufficient? Do you have any suggestions.

```
{
   "data": {
      "id": "sv-DYXpEz7SQBrRJEHV",
      "type": "state-versions",
      "attributes": {
         "created-at": "2022-07-06T17:54:23.254Z",
         "size": 699,
         "hosted-state-download-url": "https://tfcdev-2c13224a.ngrok.io/_archivist/v1/object/dmF1bHQ6djE6MkpXVGFwQmhqWnRheGdGZEpjUVV4UGFCRFhyVm9nTTV3a044SVF3eldHTGtFVHNpSXpYOVRBWHRNNkU0SVlxMVVObjZPajRId0drUGlXN0N4dTdxRmpTeGV6Yk56SmtIUHRCbVJsN3pGNGhpSWswMnNJUWJ4RDhMcEdnMERwZk5ycFRpemhaVWlyVmZXWmZsYkJxYnVQU0tpOGJrOTYvL21NOWswU1ZQRXkxeWF2WTBxTElJK0gzTlBYYmRpbThsUlpxeDZldUVJQ1FYMEN3VHJXS1d4cmVqRzMrT2xHWVl4NktYWlNpTStTWWhzZ2lDYXExMGZLSjN4ekxJVWNYdjVwaDJtM1hCYWlDUEUzNGtaV0luSDZwbTVzZmcxNG5ZQWVQVkJHZzhUWjFBQk1KRFFrd3BDY0dya3ZjRlM4UktITmpKNjlEMG01NXI1QTJtT3dtd3JCMFlwek50bzY5bXkrMnk1V1FPMW9TM2t1c3BITFpocmNMaFNlOTJ4VjVOVmlhSTVKbnVjdHlNTERaU0gxbkRpMWpCTy9XSlNYQT0",
         "modules": {
            "root": {
               "random_pet": 3
            }
         },
         "providers": {
            "provider[\"registry.terraform.io/hashicorp/random\"]": {
               "random_pet": 3
            }
         },
         "resources": [
            {
               "name": "animal_trio",
               "type": "random_pet",
               "count": 3,
               "module": "root",
               "provider": "provider[\"registry.terraform.io/hashicorp/random\"]"
            }
         ],
         "resources-processed": true,
         "serial": 0,
         "state-version": 4,
         "terraform-version": "1.2.4",
         "vcs-commit-url": null,
         "vcs-commit-sha": null
      },
      "relationships": {
         "run": {
            "data": {
               "id": "run-T3EUTgV3yzSR91GE",
               "type": "runs"
            }
         },
         "next-state-version": {
            "data": null
         },
         "previous-state-version": {
            "data": null
         },
         "created-by": {
            "data": {
               "id": "user-wJVRjZegtRxpexoa",
               "type": "users"
            },
            "links": {
               "self": "/api/v2/users/user-wJVRjZegtRxpexoa",
               "related": "/api/v2/runs/run-T3EUTgV3yzSR91GE/created-by"
            }
         },
         "workspace": {
            "data": {
               "id": "ws-LRczoXxb9a3XSmLh",
               "type": "workspaces"
            }
         },
         "outputs": {
            "data": [
               {
                  "id": "wsout-cJjt3xxikeKDGA1g",
                  "type": "state-version-outputs"
               }
            ],
            "links": {
               "related": "/api/v2/state-versions/sv-DYXpEz7SQBrRJEHV/outputs"
            }
         }
      },
      "links": {
         "self": "/api/v2/state-versions/sv-DYXpEz7SQBrRJEHV"
      }
   }
}
```

state_version_integration_test.go
```diff
@@ -333,6 +333,12 @@ func TestStateVersionsRead(t *testing.T) {
 		sv.Outputs = nil
 
 		assert.Equal(t, svTest, sv)
+		assert.NotEmpty(t, svTest, svTest.ResourcesProcessed)
```
brandonc, line 336: This is what I meant! Thanks.

I am slightly worried that some of these will indeed be empty if the backend hasn't yet processed the state. The way to fix this would be to retry while ResourcesProcessed was false. Or we could just wait and see if it flakes!

brandonc (approved):

state_version_integration_test.go
laurenolivia: @brandonc Wouldn't it stand true that if ResourcesProcessed were false, it would still validate on NotEmpty? Like regardless of which bool, it will still always contain a value?

laurenolivia: Ah, the attr returning `false` could be an indicator that the other exposed attributes may be empty. Got it.

brandonc: `false` is actually empty as well because it's the default value of a boolean. This is the source of a lot of bugs, I think! Check this out:

https://go.dev/play/p/Qsz5Q7FdAWU

laurenolivia: Could this example in the [State Version API docs](https://www.terraform.io/cloud-docs/api-docs/state-versions#list-state-versions-for-a-workspace) be incorrect? The response has `resources-processed: false` but the other attributes I exposed are populated. 

```
"modules": {
          "root": {
              "null-resource": 1,
               "data.terraform-remote-state": 1
                }
            },
            "providers": {
                "provider[\"terraform.io/builtin/terraform\"]": {
                    "data.terraform-remote-state": 1
                },
                "provider[\"registry.terraform.io/hashicorp/null\"]": {
                    "null-resource": 1
                }
            },
            "resources": [
                {
                    "name": "other_username",
                    "type": "data.terraform_remote_state",
                    "count": 1,
                    "module": "root",
                    "provider": "provider[\"terraform.io/builtin/terraform\"]"
                },
                {
                    "name": "random",
                    "type": "null_resource",
                    "count": 1,
                    "module": "root",
                    "provider": "provider[\"registry.terraform.io/hashicorp/null\"]"
                }
            ],
            "resources-processed": false,
```

brandonc: Yes, it is. If resources_processed is false, you would expect the resources key to be empty.

laurenolivia: Ah, ok, according to the [docs](https://pkg.go.dev/github.com/realistschuckle/testify/assert?utm_source=godoc#NotEmpty), there's a `func NotEmpty` workaround. 
```
Empty asserts that the specified object is NOT empty. I.e. not nil, "", false, 0 or a slice with len == 0. 
Returns whether the assertion was successful (true) or not (false).
```


```
if assert.NotEmpty(t, obj) {
  assert.Equal(t, "two", obj[1])
}
```

So maybe `assert.Equal()` that the value is either `true` or `false`, using both `assert.Empty` and `assert.NotEmpty`. What do you think?

PR #485, opened 2022-07-27
Splits the retry helper so it can be parameterized

brandonc:
adds retryPatiently for longer, slower retries

retry: 6 tries, 2 second wait = 12 seconds
retryPatiently: 40 tries, 3 second wait = 2 minutes


nfagerlund (approved):
This seems great, and should reduce flakes. 

As a "nice to have" that I won't block merge on: the base `retry()` still seems like a bit of an attractive nuisance, given that the main type of thing we seem to be retrying involves asynchronous jobs running in a TFC instance. What do you think about bumping the "impatient" version to like 30 or 40 seconds, to make flakes less likely when people inevitably reach for it in the future? 

Also, not sure why the run tasks tests failed on you, but if they pass on re-run, :shipit:.

brandonc:
@nfagerlund I agree. Bumped to 10 retries over 30 seconds

nfagerlund (approved):

nfagerlund (approved):

PR #486, opened 2022-07-27
Add PullRequestOutputsEnabled to Workspace settings

lilincmu:
## Description

Add `PullRequestOutputsEnabled` to Workspace settings.

The PR is not ready to be released yet due to other dependencies. Will unmark its draft status when it's ready to be reviewed.

## External links
- [Asana](https://app.asana.com/0/1201287664926243/1202307454909375/f)
- [RFC](https://docs.google.com/document/d/1EtPNCPfiPQfSEwe3ZoAGS5tk6PahmTROD20OtgmUYwc/edit#)

### DIFF
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -434,9 +434,8 @@ func TestWorkspacesCreate(t *testing.T) {
 	})
 
 	t.Run("when both speculative-enabled and pull-request-outputs-enabled are true (behind a feature flag)", func(t *testing.T) {
-		skipIfBeta(t)
 		options := WorkspaceCreateOptions{
-			Name:                      String("foobar"),
+			Name:                      String("tst-foobar"),
 			SpeculativeEnabled:        Bool(true),
 			PullRequestOutputsEnabled: Bool(true),
 		}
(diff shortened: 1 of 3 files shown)


workspace_integration_test.go
```diff
@@ -432,6 +432,50 @@ func TestWorkspacesCreate(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, options.TriggerPatterns, w.TriggerPatterns)
 	})
+
+	t.Run("when both speculative-enabled and pull-request-outputs-enabled are true (behind a feature flag)", func(t *testing.T) {
+		skipIfBeta(t)
```
matejrisek, line 437: Let's not skip tests, but rather set the feature flag variant in `integration_flags.json`.
This file is read by the CI Atlas instance and in it you can define set of rules when to flip the feature flag on - e.g. if your organization name ends with `-ff-on`. ([an example](https://github.com/hashicorp/go-tfe/blob/6b5c37fbd622f39bd9784ff730f3595d213d58da/workspace_integration_test.go#L304-L332))

lilincmu: I see. Thanks for letting me know!

Aside from using suffix like `-ff-on`, may I just blindly flip the feature on for every org with `tst-` prefix? Is it necessary to tests both on/off cases under the `tst-` orgs? I created a draft [PR](https://github.com/hashicorp/atlas/pull/12998/files) to update `integration_flags.json` and will update it accordingly.

PR #488, opened 2022-08-01
Workflow test runner CI 

sebasslash:
## Description
These changes introduce the final steps of migrating from CircleCI to Github Actions. Upon approval, we'll be removing the CircleCI integration from this repository (which may break some open PRs -- **they'll need to rebase to run the new workflow**).

`fetch_outputs`: This script will fetch the current state version outputs from the TFC workspace that provisions our CI instance and pass them to `$GITHUB_ENV`. We can then read these variables in our test runner step and pass them to `go test` (which is wrapped by `gotestsum`) 

What will this enable?
- More granular control over our CI instance. Coverage was limited due to an inflexible way to toggle features behind a FF or features related to TFE. 
- We no longer share the same CI instance across different repositories. This will help reduce flakiness. 
- We can more easily extend the test harness in Go to add more custom functionality to our tests. Perhaps we need an external service like a TFC agent? 
- Nightly tests

## For the reviewer
Currently the `fetch_outputs` script will grab the latest SV outputs and write them to `$GITHUB_ENV`. Those values are then passed to `go test`. However, I'm wondering if it would be inherently better to wrap this entire process within one Go script instead? We would invoke the test runner process directly using `exec.Command()` enabling us to dynamically set environment variables, as well as circumventing the need to write to `$GITHUB_ENV`. What do you think?


.github/scripts/fetch_outputs/main.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// We need to create a new tfe client that will talk to the tflocal instance
+	// since we need to fetch TFE_USER1 and TFE_USER2
+	client, err := tfe.NewClient(tfeCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	oml, err := client.OrganizationMemberships.List(ctx, "hashicorp", &tfe.OrganizationMembershipListOptions{
```
brandonc, line 67: What is the significance of this org? Can this be extracted to a named const so the meaning is more apparent?

brandonc (changes requested):

brandonc (commented):
FWIW I like the usage of GITHUB_ENV instead of running go test directly.

.github/workflows/ci.yml
```diff
@@ -40,4 +40,46 @@ jobs:
...
+          path: ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+      - name: Set local Go version
+        run: |
+          VERSION=`cat .go-version | awk '{printf$1}'`
+          echo "go_version=$VERSION" >> $GITHUB_ENV
+      - name: Setup Go environment
+        uses: actions/setup-go@v2
+        with:
+          go-version: "${{ env.go_version }}"
```
taiidani, line 63: Non-blocker -- setup-go can actually
* Automatically cache Go modules between runs, no `actions/cache` step required.
* Automatically detect the [major.minor] version of Go from the go.mod file.

See the [internal] https://github.com/hashicorp/citool-example-project/blob/main/.github/workflows/test.yml#L11-L15 for an example of this in use.

.github/workflows/ci.yml
```diff
@@ -40,4 +40,46 @@ jobs:
...
+          TFE_TOKEN: "${{ secrets.TF_WORKFLOW_TFLOCAL_CLOUD_TFC_TOKEN }}"
+        run: |
+          go run .github/scripts/fetch_outputs/main.go hashicorp-v2 tflocal-go-tfe
+      - name: Run Tests
+        env:
+          TFE_TOKEN: "${{ env.TFE_TOKEN }}"
+          TFE_ADDRESS: "${{ env.TFE_ADDRESS }}"
+          TFC_RUN_TASK_URL: "https://httpstat.us/200"
+        run: |
+          export GO111MODULE=on
```
taiidani, line 81: Non-blocker -- slightly more consistent?

```suggestion
          TFC_RUN_TASK_URL: "https://httpstat.us/200"
          GO111MODULE: "on"
        run: |
```

taiidani (commented):
All non-blockers! My team is just here to consult and advise 😄

PR #489, opened 2022-08-01
remove references to circle CI

brandonc:
## Output from tests

```

...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 Terraform Cloud/Enterprise Go Client
 ==============================
 
-[![Build Status](https://circleci.com/gh/hashicorp/go-tfe.svg?style=shield)](https://circleci.com/gh/hashicorp/go-tfe)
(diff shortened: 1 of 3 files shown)


sebasslash (dismissed):
:+1:

brandonc:
@sebasslash I don't think its true anymore that people outside the org cannot run CI workflows. I'm going to reword some of this. They just have to be approved once.

README.md
```diff
@@ -1,7 +1,6 @@
 Terraform Cloud/Enterprise Go Client
 ==============================
 
-[![Build Status](https://circleci.com/gh/hashicorp/go-tfe.svg?style=shield)](https://circleci.com/gh/hashicorp/go-tfe)
```
annawinkler, line 4: What do you think about adding a [Github Actions build badge](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) instead?

annawinkler (commented):
Maybe we could add a badge for GH actions? 💖

annawinkler (approved):
👍 My comment was addressed in another [PR](https://github.com/hashicorp/go-tfe/pull/509)

PR #490, opened 2022-08-02
add no-code to module CREATE api, introduce PATCH api for module

dsa0x:
## Output from tests

```

...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -44,6 +44,9 @@ type RegistryModules interface {
 	// requires a path to the configuration files on disk, which will be packaged by
 	// hashicorp/go-slug before being uploaded.
 	Upload(ctx context.Context, rmv RegistryModuleVersion, path string) error
+
+	// Update a registry module
+	Update(ctx context.Context, moduleID RegistryModuleID, options RegistryModuleUpdateOptions) (*RegistryModule, error)
 }
 
 // registryModules implements RegistryModules.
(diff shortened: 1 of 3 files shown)


registry_module.go
```diff
@@ -265,6 +282,45 @@ func (r *registryModules) Create(ctx context.Context, organization string, optio
...
+	}
+
+	u := fmt.Sprintf(
+		"organizations/%s/registry-modules/%s/%s/%s/%s",
+		url.QueryEscape(moduleID.Organization),
+		url.QueryEscape(string(moduleID.RegistryName)),
+		url.QueryEscape(moduleID.Namespace),
+		url.QueryEscape(moduleID.Name),
+		url.QueryEscape(moduleID.Provider),
+	)
```
brandonc, line 308: (nonblocking) 💅 this could be moved to a new RegistryModuleID method, like `func (r RegistryModuleID) URL() string` since it is also duplicated in Read

registry_module.go
```diff
@@ -265,6 +282,45 @@ func (r *registryModules) Create(ctx context.Context, organization string, optio
...
+
+	if moduleID.RegistryName == "" {
+		log.Println("[WARN] Support for using the RegistryModuleID without RegistryName is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the RegistryName in RegistryModuleID.")
+		moduleID.RegistryName = PrivateRegistry
+	}
+
+	if moduleID.RegistryName == PrivateRegistry && strings.TrimSpace(moduleID.Namespace) == "" {
+		log.Println("[WARN] Support for using the RegistryModuleID without Namespace is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the Namespace in RegistryModuleID.")
+		moduleID.Namespace = moduleID.Organization
+	}
```
brandonc, line 299: (nonblocking) 💅 Since this logic and warnings are repeated in the Read function, it could be moved to a shared *RegistryModuleID method

brandonc (commented):
Is this beta API? See [this section of the contributing guide](https://github.com/hashicorp/go-tfe/blob/main/docs/CONTRIBUTING.md#adding-api-changes-that-are-not-generally-available) for more details about introducing non-GA features to this lib. (This is also the reason why tests are failing!)

registry_module.go
dsa0x: Thank you @brandonc. Addressed in 0362848

PR #498, opened 2022-08-10
[Fix] Remove incorrect parameter to read go.mod version, use go-version-file

sebasslash:
The workflow to rebuild the tflocal instance was [failing](https://github.com/hashicorp/go-tfe/actions/workflows/tflocal-instance.yml) due to an incorrect parameter used to set the Go version with `go.mod`. 

The workflow was setting `go-version: go.mod` when it should be `go-version-file: go.mod`.


annawinkler (approved):
🚀

annawinkler (approved):
🚀

PR #503, opened 2022-08-17
Update workflow to include pull_request_target

annawinkler:
## Description

Add [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) to the ci workflow so forks from this repository will also be able to run the ci integration tests.

## Testing plan

1. We could create a fork from this branch and see if the tests run.
2. Do we tests workflows any other way? (I was reading about [act](https://github.com/nektos/act)

## External links

- Related [article](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks)

## Output from tests

```

...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: test
-on:
-  pull_request:
-    branches: [ main ]
+
+on: [pull_request, pull_request_target]
+  branches: [ main ]
+
 jobs:
   codegen:
     name: Codegen
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: test
 
-on: [pull_request, pull_request_target]
-  branches: [ main ]
+on:
+  pull_request_target:
+    branches: [ main ]
 
 jobs:
   codegen:


sebasslash (approved):
I've tested in #504 :heavy_check_mark: however I don't want to conflate the fact I have write access to this repository, so we need to ensure the workflow requires approval to run (it should). Would need someone w/o write access for a sanity check.

annawinkler:
I'm not seeing the test workflow running though 🤔

sebasslash:
Re-reading through the docs, it looks like there is no way a fork can access Github secrets (I assumed manual approval would be sufficient). I think our solution here  is to either create a local branch with the fork's changes manually or introduce some automated way to do it.

annawinkler:
> Re-reading through the docs, it looks like there is no way a fork can access Github secrets (I assumed manual approval would be sufficient). I think our solution here is to either create a local branch with the fork's changes manually or introduce some automated way to do it.

Hm... It sounded like we could based on this [post](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks) 🤔

laurenolivia:
My understanding is `pull_request_target` was introduced because the `pull_request` event prevents write permissions and secrets access to the target repository. Thus, workflows triggered via `pull_request_target` have write permissions to the target repository. [Additionally](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), workflows triggered by `pull_request_target` events are run in the context of the base branch, and the base branch is considered trusted, so those workflows will always run.

On the contrary, I've [come across](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) a few [warnings](https://blog.gitguardian.com/github-actions-security-cheat-sheet/) regarding the `pull_request_target` event.

Security Lab @ GitHub noted:
```
You may ask yourself: if the pull_request_target workflow only checks out and builds the 
PR, i.e. runs untrusted code but doesn’t reference any secrets, is it still vulnerable?

Yes it is, because a workflow triggered on pull_request_target still has the read/write repository 
token in memory that is potentially available to any running program.
```

annawinkler:
Closing because we're still working on ideas.

PR #508, opened 2022-08-23
Remove tflocal admin token variable

sebasslash:
Removing this variable from the module configuration, enabling the instance to generate one dynamically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,13 +69,12 @@ jobs:
 
       - name: Run Tests
         env:
-          TFE_TOKEN: "${{ env.TFE_TOKEN }}"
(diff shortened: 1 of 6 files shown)


brandonc (approved):
Cool! I think [the documentation](https://hashicorp.atlassian.net/wiki/spaces/TDP/pages/2400911434/Configuring+a+tflocal-cloud+instance+for+CI+GitHub+Actions) may need to be modified?

.github/workflows/ci.yml
```diff
@@ -69,13 +69,12 @@ jobs:
 
       - name: Run Tests
         env:
-          TFE_TOKEN: "${{ env.TFE_TOKEN }}"
```
brandonc, line 72: I just found out that if this were fetched from the github secrets config, it would output `***` in the action log. Just FYI for the future

PR #513, opened 2022-08-25
Set tflocal build script to trigger a run with auto apply

sebasslash:
Last night, `instance` workflow triggered a run that ended up stuck in the waiting confirmation state. Given any runs triggered by this workflow won't capture any new configuration changes, we can safely set it to auto-apply, eliminating the need for someone to intervene manually.


annawinkler (approved):
:shipit:

PR #514, opened 2022-08-26
add the optional StateVersion field JSONState and changes to string for base64 encoding

megan07:
## Description

add JSONState attribute to be base64 encoded (a string)
(basically adds this PR back in: https://github.com/hashicorp/go-tfe/pull/444)

## Output from tests

```

...
```

### DIFF
--- a/state_version_integration_test.go
+++ b/state_version_integration_test.go
@@ -177,7 +177,7 @@ func TestStateVersionsCreate(t *testing.T) {
 			MD5:              String(fmt.Sprintf("%x", md5.Sum(state))),
 			Serial:           Int64(1),
 			State:            String(base64.StdEncoding.EncodeToString(state)),
-			JSONState:        String(base64.StdEncoding.EncodeToString(state)),
+			JSONState:        String(base64.StdEncoding.EncodeToString(jsonState)),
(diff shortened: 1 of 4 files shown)


state_version_integration_test.go
```diff
@@ -120,6 +120,11 @@ func TestStateVersionsCreate(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	jsonState, err := ioutil.ReadFile("test-fixtures/ext-state-version/state.json")
```
brandonc, line 123: Youll need to add this fixture, but also can you rename it to `json-state/state.json` ?

brandonc (changes requested):

state_version_integration_test.go
```diff
@@ -172,6 +177,7 @@ func TestStateVersionsCreate(t *testing.T) {
 			MD5:              String(fmt.Sprintf("%x", md5.Sum(state))),
 			Serial:           Int64(1),
 			State:            String(base64.StdEncoding.EncodeToString(state)),
+			JSONState:        String(base64.StdEncoding.EncodeToString(state)),
```
brandonc, line 180: this should be EncodeToString(jsonState)

brandonc (changes requested):

brandonc (approved):

PR #517, opened 2022-08-30
Create a TaskResult object for the client

glennsarti:
## Description

This commit actually creates a TaskResult struct. Without this change users will receive
nil pointer errors when attempting to use this part of the client.

## Testing plan

N/A


annawinkler (approved):
Thank you for this change! I think we might benefit from having a test for TaskResult too, so I'll add an issue for us to follow up with that.

laurenolivia:
@glennsarti very nit, but it looks like those services are listed alphabetically, and `client.TaskResults = &taskResults{client: client}` should be shifted up 1 row.

glennsarti:
@laurenolivia Raised PR #522 to reorder things.

PR #521, opened 2022-08-30
Fix struct tag for varset relationship

### LINKED ISSUE #491: VariableSetVariable struct has null value for VariableSet
`VariableSetVariables.List()` returns a `[]VariableSetVariable`. 

`VariableSetVariable` has a `VariableSet` property, but it is always set to null in the output from `VariableSetVariables.List()`.

I'm trying to generate a combined list of variables for a workspace that includes both workspace variables and variableset variables, and I would like to indicate the source of each variable in my program's output. 

Without the `VariableSet` property being set on `VariableSetVariable`, it's rather difficult to know where a varset variable has come from without passing around extra context.

sebasslash:
This PR fixes #491, where the struct tag definition for the variable set relationship was incorrectly named thus resulting in a null value.

### DIFF
--- a/variable_set_variable.go
+++ b/variable_set_variable.go
@@ -49,7 +49,7 @@ type VariableSetVariable struct {
 	Sensitive   bool         `jsonapi:"attr,sensitive"`
 
 	// Relations
-	VariableSet *VariableSet `jsonapi:"relation,configurable"`
+	VariableSet *VariableSet `jsonapi:"relation,varset"`
 }
 
 type VariableSetVariableListOptions struct {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Bug Fixes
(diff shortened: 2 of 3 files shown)


variable_set_variable.go
```diff
@@ -49,7 +49,7 @@ type VariableSetVariable struct {
 	Sensitive   bool         `jsonapi:"attr,sensitive"`
 
 	// Relations
-	VariableSet *VariableSet `jsonapi:"relation,configurable"`
+	VariableSet *VariableSet `jsonapi:"relation,varset"`
```
laurenolivia, line 52: Are `varset.data.id` and `varset.data.type` being exposed?

Screen Shot 2022-08-30 at 5 27 48 PM

sebasslash: Yep! The relationship is automatically included in the response body.

annawinkler (approved):
🥇

PR #523, opened 2022-09-01
Update for pre apply

glennsarti:
## Output from tests

```

...
```


glennsarti:
Hrmm, the ` TestStateVersionsRead` is being flaky

laurenolivia (commented):
I think there's an unofficial refactor where we're replacing `defer t.Cleanup()` with `t.Cleanup()`

laurenolivia:
I ran these tests locally, and got the error pasted below, but it looks like that env var is declared [here](https://github.com/hashicorp/go-tfe/blob/main/.github/workflows/ci.yml#L72). Am I missing something? 🤔 

```
Cannot create a run task with an empty URL. You must set TFC_RUN_TASK_URL for run task related tests.
invalid url for run task URL
```

glennsarti:
Err... the tests definitely work in CI. I assume you set your local env with that variable?

glennsarti:
Ah ha, kept rekicking the CI job until it was green!

glennsarti:
Yeah, I didn't bother with a changelog entry as it was only an extra const being added, and some code comment removals.

PR #526, opened 2022-09-02
fix stateversion flaky test

laurenolivia:
## Description

This PR fixes a flaky test.

## Testing plan

1. `svTest.ResourcesProcessed` and `sv.ResourcesProcessed` should both either return true, or fail the test.
2. Running tests locally should suffice.

## Output from tests

```

...
```

### DIFF
--- a/state_version_integration_test.go
+++ b/state_version_integration_test.go
@@ -318,26 +318,39 @@ func TestStateVersionsRead(t *testing.T) {
 	t.Cleanup(svTestCleanup)
 
 	t.Run("when the state version exists", func(t *testing.T) {
-		sv, err := client.StateVersions.Read(ctx, svTest.ID)
-		require.NoError(t, err)
+		if !svTest.ResourcesProcessed {
+			_, err := retry(func() (interface{}, error) {
+				svTest, err := client.StateVersions.Read(ctx, svTest.ID)
+				require.NoError(t, err)
 
-		// Don't compare the DownloadURL because it will be generated twice
-		// in this test - once at creation of the configuration version, and
-		// again during the GET.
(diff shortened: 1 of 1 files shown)


state_version_integration_test.go
```diff
@@ -318,26 +318,39 @@ func TestStateVersionsRead(t *testing.T) {
...
-		assert.NotEmpty(t, svTest, svTest.Providers)
-		assert.NotEmpty(t, svTest, svTest.Resources)
+		if !svTest.ResourcesProcessed {
+			_, err := retry(func() (interface{}, error) {
+				svTest, err := client.StateVersions.Read(ctx, svTest.ID)
+				require.NoError(t, err)
+
+				if !svTest.ResourcesProcessed {
+					return nil, fmt.Errorf("resources not processed %s", err)
+				} else {
```
brandonc, line 328: I think it would be clearer to move this entire block outside of and right after retry (line 352 in this version)

That way, the retry can concentrate on awaiting ResourcesProcessed and then the test can proceed

The problem here is that if resources are processed on the very first attempt, the assertions are not run at all.

brandonc (changes requested):

state_version_integration_test.go
laurenolivia: The problem I ran into here, was accessing `svTest.ResourcesProcessed` when it becomes `true`. It will reference the global instance that is `false` rather than the updated value.

laurenolivia: Any suggestions?

brandonc: You can just assume it is true unless retry returns a non-nil error, because retry always returns an error if it remains false

PR #528, opened 2022-09-07
Add/Update Community Health Files

barrettclark:
* Move CONTRIBUTING.md
* Add CODE_OF_CONDUCT.md
* Move SECURITY.md


barrettclark:
@jptoto Is this what y'all had in mind? Are there other repo configuration settings that we also need to update? I ask because  it looks like we'd manually added some of the files that it looks like we should have had in the `.github` directory.

PR #525, opened 2022-09-01
Resource scaffolding script

sebasslash:
This script introduces a quick way for contributors to generate boilerplate when adding an entirely new set of endpoints to go-tfe. 

Usage:

```sh
./generate_resource.sh <resource name>
```

Example:
```sh
./generate_resource.sh Foobar
```

It will create a file, `foobar.go` and `foobar_integration_test.go` that look as follows:

`foobar.go`
```go
package tfe

import (
        "context"
)

var _ Foobars = (*foobars)(nil)

// Foobars describes all the foobar related methods that the Terraform
// Enterprise API supports
//
// TFE API docs: (TODO: ADD DOCS URL)
type Foobars interface {
        // List all foobars.
        List(ctx context.Context, options *FoobarListOptions) (*FoobarList, error)

        // Create a foobar.
        Create(ctx context.Context, options FoobarCreateOptions) (*Foobar, error)

        // Read a foobar by its ID.
        Read(ctx context.Context, foobarID string) (*Foobar, error)

        // Read a foobar by its ID with options.
        ReadWithOptions(ctx context.Context, foobarID string, options *FoobarReadOptions) (*Foobar, error)

        // Update a foobar.
        Update(ctx context.Context, foobarID string, options FoobarUpdateOptions) (*Foobar, error)

        // Delete a foobar.
        Delete(ctx context.Context, foobarID string) error
}

// foobars implements Foobars
type foobars struct {
        client *Client
}

// FoobarList represents a list of foobars
type FoobarList struct {
        *Pagination
        Items []*Foobar
}

// Foobar represents a Terraform Enterprise foobar
type Foobar struct {
        ID string `jsonapi:"primary,foobars"`
        // Add more fields here
}

// FoobarListOptions represents the options for listing foobars
type FoobarListOptions struct {
        ListOptions

        // Add more list options here
}

// FoobarCreateOptions represents the options for creating foobars
type FoobarCreateOptions struct {
        Type string `jsonapi:"primary,foobars"`
        // Add more create options here
}

// FoobarReadOptions represents the options for reading a foobar
type FoobarReadOptions struct {
        // Add more read options here
}

// FoobarUpdateOptions represents the options for updating a foobar
type FoobarUpdateOptions struct {
        ID string `jsonapi:"primary,foobars"`

        // Add more update options here
}

// List all foobars
func (s *foobars) List(ctx context.Context, options *FoobarListOptions) (*FoobarList, error) {
        panic("not implemented")
}

// Create a new foobar with the given options.
[...]

PR #530, opened 2022-09-12
Resource scaffolding script v2

sebasslash:
An extension of #525, where I've ported the script `generate_resource` over to Go, giving us greater flexibility in code generation. Namely, the biggest benefit (aside from writing it in Go) is that we can properly case the resource name without having to write a complex bash script. 

I've also introduced a `Makefile` since we've been writing a few scripts to help improve the workflow for contributors. 

You can run `generate_resource`:
```sh
RESOURCE=example_resource make generate
```

Will create `ExampleResource`, stub out the interface and struct methods, and generate an integration test file. 

Other Makefile targets include:
* `make vet`: runs `go vet`
* `make lint`: runs `golangci-lint run .`
* `make test`: runs `go test`, e.g

```sh
TESTARGS="-run TestRunsList" make test
```

* `make mocks`: automates mock creation by appending a `mockgen` command to `generate_mocks.sh` and then running that script, e.g

```sh
FILENAME=example_resource.go make mocks
```

* `make envvars`: runs our `./script/setup-test-envvars.sh` script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,7 @@
-# This rule runs the resource scaffolding script
-# $1 is the name of the resource to generate
-check_defined = \
-    $(strip $(foreach 1,$1, \
-        $(call __check_defined,$1,$(strip $(value 2)))))
-__check_defined = \
-    $(if $(value $1),, \
-      $(error Undefined $1$(if $2, ($2))))
-
 .PHONY: vet fmt lint test mocks envvars generate
 
-generate:
-	$(call check_defined, RESOURCE)
+# Make target to generate resource scaffolding for specified RESOURCE
+generate: check-resource
 	@cd ./scripts/generate_resource; \
 	go mod tidy; \
 	go run . $(RESOURCE) ;
@@ -29,11 +20,21 @@ lint:
 test:
 	go test ./... $(TESTARGS) -tags=integration -timeout=30m
 
-mocks:
-	$(call check_defined, FILENAME)
+# Make target to generate mocks for specified FILENAME
+mocks: check-filename
 	@echo "mockgen -source=$(FILENAME) -destination=mocks/$(FILENAME) -package=mocks" >> generate_mocks.sh
 	./generate_mocks.sh
 
 envvars:
 	./scripts/setup-test-envvars.sh
+
+check-filename:
+ifndef FILENAME
+	$(error Missing FILENAME param. Example usage: FILENAME=example_resource.go make mocks)
+endif
+
+check-resource:
+ifndef RESOURCE
+	$(error Missing RESOURCE param. Example usage: RESOURCE=foo_bar make generate)
+endif
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,39 @@
+# This rule runs the resource scaffolding script
+# $1 is the name of the resource to generate
+check_defined = \
+    $(strip $(foreach 1,$1, \
+        $(call __check_defined,$1,$(strip $(value 2)))))
+__check_defined = \
+    $(if $(value $1),, \
+      $(error Undefined $1$(if $2, ($2))))
+
+.PHONY: vet fmt lint test mocks envvars generate
+
+generate:
+	$(call check_defined, RESOURCE)
+	@cd ./scripts/generate_resource; \
+	go mod tidy; \
+	go run . $(RESOURCE) ;
+
+fmt:
+
+vet:
+	go vet
+
+fmt:
+	go fmt ./...
+
+lint:
+	golangci-lint run .
+
+test:
+	go test ./... $(TESTARGS) -tags=integration -timeout=30m
+
+mocks:
+	$(call check_defined, FILENAME)
+	@echo "mockgen -source=$(FILENAME) -destination=mocks/$(FILENAME) -package=mocks" >> generate_mocks.sh
+	./generate_mocks.sh
+
+envvars:
+	./scripts/setup-test-envvars.sh
+
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -19,7 +19,7 @@ There are instances where several new resources being added (i.e Workspace Run T
 ## Running the Linters Locally
 
 1. Ensure you have [installed golangci-lint](https://golangci-lint.run/usage/install/#local-installation)
-2. From the CLI, run `golangci-lint run`
+2. From the CLI, run `make lint`
 
 ## Writing Tests
 
@@ -36,6 +36,12 @@ You'll need to generate mocks if an existing endpoint method is modified or a ne
 
 ```
 mockgen -source=example_resource.go -destination=mocks/example_resource_mocks.go -package=mocks
+```
+
+Alternatively, you can use the Makefile target `mocks` to automate the steps:
+
+```
+FILENAME=example_resource.go make mocks
 ```
 
 ## Adding API changes that are not generally available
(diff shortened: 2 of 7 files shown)


scripts/generate_resource/templates.go
```diff
@@ -0,0 +1,205 @@
...
+Not Allowed: policy123, #user, my cool resource
+
+If your resource contains multiple terms, e.g policy set, you must use an underscore delimiter for each term in order to generate proper casing in your code. For example, if you wanted to generate the policy set resource as PolicySet, you would pass policy_set as your argument.
+
+Example usage: go run ./scripts/generate_resource/main.go policy_set`
+}
+
+func sourceTemplate() string {
+	return `
+package tfe
```
sebasslash, line 47: This should probably be defined as a constant instead.

scripts/generate_resource/main.go
```diff
@@ -0,0 +1,99 @@
...
+		Resource:          strcase.ToCamel(name),
+		ResourceInterface: strcase.ToCamel(pluralName),
+		ResourceStruct:    strcase.ToLowerCamel(pluralName),
+		ResourceID:        fmt.Sprintf("%sID", strcase.ToLowerCamel(name)),
+		ListOptions:       fmt.Sprintf("%sListOptions", strcase.ToCamel(name)),
+		ReadOptions:       fmt.Sprintf("%sReadOptions", strcase.ToCamel(name)),
+		CreateOptions:     fmt.Sprintf("%sCreateOptions", strcase.ToCamel(name)),
+		UpdateOptions:     fmt.Sprintf("%sUpdateOptions", strcase.ToCamel(name)),
+	}
+}
```
sebasslash, line 41: `strcase.ToCamel(name)` , `stracse.ToCamel(pluralName)`, should probably be refactored and stored as a variable.

Makefile
```diff
@@ -0,0 +1,27 @@
+# This rule runs the resource scaffolding script
+# $1 is the name of the resource to generate
+
+generate:
+	@cd ./scripts/generate_resource; \
+	go mod download; \
+	go mod tidy; \
```
brandonc, line 7: I thought go mod tidy also effectively does a go mod download. Does doing a separate download step do something different?

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	go run . $(resource) ;
+
+vet:
+	go vet
+
+lint:
```
brandonc, line 13: what about `go fmt -c` ? https://github.com/hashicorp/terraform-provider-tfe/blob/main/scripts/gofmtcheck.sh

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	golangci-lint run .
+
+test:
+	go test ./... $(TESTARGS) -tags=integration -timeout=30m
+
+mocks:
+	@echo "mockgen -source=$(filename) -destination=mocks/$(filename) -package=mocks" >> generate_mocks.sh
```
brandonc, line 20: If I type "make mocks" it will add an empty line to this file: `mockgen -source= -destination=mocks/ -package=mocks`

Is it possible to test these make parameters to make sure they exist?

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	go run . $(resource) ;
+
+vet:
+	go vet
+
+lint:
+	golangci-lint run .
+
+test:
+	go test ./... $(TESTARGS) -tags=integration -timeout=30m
```
brandonc, line 17: FWIW I prefer this type of parameterization to --filename="blah.go"

brandonc (commented):
Initial feedback

Makefile
sebasslash: :laughing: I didn't even consider this since I have `go fmt` running on save! Will add :+1:

Makefile
```diff
@@ -0,0 +1,39 @@
...
+check_defined = \
+    $(strip $(foreach 1,$1, \
+        $(call __check_defined,$1,$(strip $(value 2)))))
+__check_defined = \
+    $(if $(value $1),, \
+      $(error Undefined $1$(if $2, ($2))))
+
+.PHONY: vet fmt lint test mocks envvars generate
+
+generate:
```
annawinkler, line 12: When I run `make generate` it would be amazing if we could print a usage message instead of `Makefile:13: *** Undefined RESOURCE.  Stop.`

docs/CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ There are instances where several new resources being added (i.e Workspace Run T
 ## Running the Linters Locally
 
 1. Ensure you have [installed golangci-lint](https://golangci-lint.run/usage/install/#local-installation)
-2. From the CLI, run `golangci-lint run`
+2. From the CLI, run `make lint`
```
annawinkler, line 22: 💯 This is nice!

docs/CONTRIBUTING.md
```diff
@@ -64,11 +70,35 @@ t.Run("with nested changes trigger", func (t *testing.T) {
...
+
+#### Running the script directly
+```sh
+cd ./scripts/generate_resource
+go run . example_resource
+```
+
+#### Running the Makefile target `generate`
+```sh
+RESOURCE=example_resource make generate
```
annawinkler, line 87: This command is different from what is in the PR description. Do both options work?

scripts/generate_resource/main.go
```diff
@@ -0,0 +1,102 @@
...
+	"os"
+	"regexp"
+	"strings"
+	"text/template"
+
+	"github.com/gertd/go-pluralize"
+	"github.com/iancoleman/strcase"
+)
+
+var validResourceName = regexp.MustCompile(`^[a-zA-Z_]+$`).MatchString
```
annawinkler, line 15: There is a difference in the validation between invoking the command from the Makefile and invoking with go.

When I use the make file I get this error:

```
[awinkler@awinkler-NJ1XWDG2N0 go-tfe (resource-scaffolding-v2 ✗)]$ make generate resource=✨
Makefile:13: *** Undefined RESOURCE.  Stop.
```

But then if I cd into `scripts/generate_resource` and run the following, I get a more helpful error message:

```
4156 scripts/generate_resource git:(resource-scaffolding-v2) ✗ » go run . ✨
2022/09/26 14:48:10 resource name can only contain letters or underscores.
exit status 1
```

It would be nice if the error from the makefile invocation was the same as the go version.

annawinkler: Ah! I tested this again:

```
[awinkler@awinkler-NJ1XWDG2N0 go-tfe (resource-scaffolding-v2 ✗)]$ RESOURCE=✨ make generate
2022/09/26 14:59:32 resource name can only contain letters or underscores.
exit status 1
make: *** [generate] Error 1
```

And the error is the same. I think I was using the wrong format of the command.

annawinkler (commented):
Thank you for adding this functionality! I think this is helpful. I have one question about if we can get the Makefile to output a usage message if someone enters `make generate` without the `RESOURCE` - if that's not possible I'm good with merging this PR as it is. 🥇

docs/CONTRIBUTING.md
sebasslash: Just updated!

docs/CONTRIBUTING.md
```diff
@@ -64,11 +70,35 @@ t.Run("with nested changes trigger", func (t *testing.T) {
...
+* We require that each resource interface provides compile-time proof that it has been implemented.
+* You'll need to add an integration test that covers each method of the resource's interface.
+* Option structs serve as a proxy for either passing query params or request bodies:
+    - `ListOptions` and `ReadOptions` are values passed as query parameters.
+    - `CreateOptions` and `UpdateOptions` represent the request body.
+* URL parameters should be defined as method parameters.
+* Any resource specific errors must be defined in `errors.go`
 
-In general, an interface should cover one RESTful resource, which sometimes involves two or more endpoints. Add all new modules to the tfe package.
+Here is a more comprehensive example of how a resource will look like when implemented. The helper script `generate_resource` will generate a subset of this example, focusing only on the core details that are required across all resources in go-tfe.
```
annawinkler, line 101:
```suggestion
Here is a more comprehensive example of what a resource looks like when implemented. The helper script `generate_resource` generates a subset of this example, focusing only on the core details that are required across all resources in go-tfe.
```

annawinkler (dismissed):
🥳

annawinkler (approved):
:shipit:

Issue #380, opened 2022-04-15
RunVariable in `/runs` request payload is not marshalled as expected


alessio-form3:
# Description

Release https://github.com/hashicorp/go-tfe/releases/tag/v0.21.0 added the ability to pass run variable to workspaces' run. 

## Expected behaviour

As for [official documentation](https://www.terraform.io/cloud-docs/api-docs/run#request-body), the expected payload should be something like:

```
"attributes": {
  "variables": [
    { "key": "replicas", "value": "2" },
    { "key": "access_key", "value": "\"ABCDE12345\"" }
  ]
}
```

## Actual behaviour

```
"variables": [
        {
          "Key": "version",
          "Value": "\"dummy\""
        }
]
```

The case of `Key` and `Value` does not seem to be respected.

brandonc:
Hello, I'm sorry to hear that run variables are not behaving as you expected. Do you mean that go-tfe is capitalizing Key and Value unnecessarily? If you can clarify what you mean I can update the documentation to be more clear.

I do think run variables are behaving as expected. Here is some test code I used to test run variables:

```go
package main

import (
	"context"

	"github.com/hashicorp/go-tfe"
)

func main() {
	client, err := tfe.NewClient(&tfe.Config{
		Address: "https://app.terraform.io",
		Token:   "TEST_TOKEN",
	})

	if err != nil {
		panic(err)
	}

	client.Runs.Create(context.Background(), tfe.RunCreateOptions{
		Workspace: &tfe.Workspace{
			ID: "ws-MY_WORKSPACE_ID",
		},
		Variables: []*tfe.RunVariable{
			{
				Key:   "foo",
				Value: "\"bar\"",
			},
		},
	})
}
```

Let me know how I can help troubleshoot!

alessio-form3:
Hello Brandon,
I am sorry not have explained clearly enough.

> Do you mean that go-tfe is capitalizing Key and Value unnecessarily?
This is exactly what I meant. The payload is marshalled ignoring the `jsonapi` meta tag definition of the `struct`.

It's also worth mentioning that our Terraform enterprise (we are on version `v202203-1`) is responding to requests successfully, so it seems that the case is insensitive on the Terraform enterprise API server side. 
I wasn't able to find any reference to JSON payload sensitiveness in the API reference, for this reason I decided to open the issue to let maintainers of this repository aware: while the client is working as expected as it is, it's not guaranteed that a "fix" on the server side wouldn't break it.  
[...]

PR #531, opened 2022-09-13
Fix ignored json meta tags for run variables in RunCreateOptions

Uk1288:
## Description: 

**NOTE: The`Variables` field in `Run` has been changed from type `[]*RunVariable` to `[]*RunVariableAttr` **

Ref PR: https://github.com/hashicorp/go-tfe/pull/204

Run variables in RunCreateOptions are not marshalled in line with the API (https://www.terraform.io/cloud-docs/api-docs/run#request-body). The variables are currently marshalled as:

```
"variables": [
        {
          "Key": "aKey",
          "Value": "\"aValue\""
        }
]
```

instead of 

```
"variables": [
        {
          "key": "aKey",
          "value": "\"aValue\""
        }
]
```

This PR adds changes that marshals and un-marshals the run variables correctly.

## Testing plan

1.  _Create a run with options_
```
	newRun, err := client.Runs.Create(context.Background(), tfe.RunCreateOptions{
		Workspace: &tfe.Workspace{
			ID: "ws-ID",
		},
		Variables: []*tfe.RunVariableOption{
			{
				Key:   "name",
				Value: "\"aName\"",
			},
		},
	})
```
1.  _`Run` should be created with correct variable key, name attributes,_
1.  _`Run` value in go client should have Key, Value fields populated as well._

## External links

- [API documentation](https://www.terraform.io/cloud-docs/api-docs/run#request-body)
- [Issue](https://github.com/hashicorp/go-tfe/issues/380)

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" TF_ACC="1" go test ./... -v -tags=integration -run TestRunCreateOptions_Marshal

=== RUN   TestRunCreateOptions_Marshal
--- PASS: TestRunCreateOptions_Marshal (1.47s)
PASS
ok 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 * Fixes null value returned in variable set relationship in `VariableSetVariable` by @sebasslash [#521](https://github.com/hashicorp/go-tfe/pull/521)
+* Fix marshalling of run variables in `RunCreateOptions`. The `Variables` field type in `Run` struct has changed from `[]*RunVariable` to `[]*RunVariableAttr` by @Uk1288 [#531](https://github.com/hashicorp/go-tfe/pull/531)
 
 # v1.9.0
 
--- a/run.go
+++ b/run.go
@@ -127,7 +127,7 @@ type Run struct {
 	StatusTimestamps       *RunStatusTimestamps `jsonapi:"attr,status-timestamps"`
 	TargetAddrs            []string             `jsonapi:"attr,target-addrs,omitempty"`
 	TerraformVersion       string               `jsonapi:"attr,terraform-version"`
-	Variables              []*RunVariable       `jsonapi:"attr,variables"`
+	Variables              []*RunVariableAttr   `jsonapi:"attr,variables"`
 
 	// Relations
 	Apply                *Apply                `jsonapi:"relation,apply"`
@@ -304,7 +304,7 @@ type RunCreateOptions struct {
 	// user confirmation. It defaults to the Workspace.AutoApply setting.
 	AutoApply *bool `jsonapi:"attr,auto-apply,omitempty"`
 
-	// RunVariables allows you to specify terraform input variables for
+	// Variables allows you to specify terraform input variables for
 	// a particular run, prioritized over variables defined on the workspace.
 	Variables []*RunVariable `jsonapi:"attr,variables,omitempty"`
 }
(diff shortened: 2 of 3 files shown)


run.go
```diff
@@ -329,6 +329,11 @@ type RunVariable struct {
 	Value string `jsonapi:"attr,value"`
 }
 
+type RunVariableOption struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
```
brandonc, line 336: Would it bet better to just redefine the RunVariable struct to be decorated as json: instead of jsonapi: ? I think that was the intent and it just got decorated incorrectly. By me, probably!

Uk1288: I initially tried redefining the RunVariable, in that case it will marshal correctly from Go value to json but un-marshalling from json to Go value stops working. The only way I found to make both direction work is if I create the two distinct types.

annawinkler (commented):
If this is a breaking change, is this a major version bump then? 
cc @brandonc

brandonc:
@annawinkler It's not a breaking change because the behavior doesn't match the documentation. (It capitalizes the keys in the request, which technically works but it's not expected) So this is just a bug fix.

brandonc:
@annawinkler I see now that you mean modifying the actual type. Let me think about it. It would probably be less impactful to add a new output type and redefine the existing input type.

run.go
brandonc: Yeah, I see what you mean, now. Since RunVariable was used as a create option, which more often times requires callers to specify the exact type name ([example](https://github.com/hashicorp/terraform/blob/76d40d281d5565d805017aeba4750191f063ab12/internal/cloud/backend_plan.go#L238-L247)), I think it would be better to redefine that as "json:" and create a new type for use in the Run struct. Like [this](https://github.com/hashicorp/go-tfe/compare/brandonc/reverse_type_fix) WDYT?

I think then we could get away with not reving the major version... Although we should still note this type change in the release notes in the context of a bug fix.

Uk1288: Yes, this makes sense. I have renamed attributes to `RunVariable` and `RunVariableAttr`. This way all existing `Create` run calls will still work.

brandonc (approved):
Don't forget to add CHANGELOG

laurenolivia:
Hmm looks like StateVersion is flaking on your PR 🤔 But I checked out this branch and you have the latest changes.

laurenolivia:
Testing plan works as expected! ✅

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 * Fixes null value returned in variable set relationship in `VariableSetVariable` by @sebasslash [#521](https://github.com/hashicorp/go-tfe/pull/521)
+* Fix marshalling of run variables in `RunCreateOptions` by @Uk1288 [#531](https://github.com/hashicorp/go-tfe/pull/531)
```
brandonc, line 5: Let's mention the type that changed as well.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 * Fixes null value returned in variable set relationship in `VariableSetVariable` by @sebasslash [#521](https://github.com/hashicorp/go-tfe/pull/521)
+* Fix marshalling of run variables in `RunCreateOptions`. The `Variables` field type in `Run` struct has changed from `[]*RunVariable` to `[]*RunVariableAttr` by @Uk1288 [#531](https://github.com/hashicorp/go-tfe/pull/531)
```
annawinkler, line 5: Nice! 🌟

annawinkler (approved):
:shipit:

PR #534, opened 2022-09-21
Improve test execution time with test splitting

sebasslash:
## Description

Currently, our CI workflow completes after about ~22 minutes which introduces unnecessary friction for community and internal pull requests. This PR introduces test splitting into our CI workflow in an effort to reduce our test execution time. While Github Actions does not support test splitting directly like other services such as CircleCI, it does support running [multiple nodes for a job](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix). Therefore we had to introduce our own bespoke strategy to split tests across these nodes. From these changes we were able to reduce our workflow execution time to ~11 minutes, or roughly a **50% reduction**. 

### Node Index Modulo Strategy
In order to evenly balance the number of tests across X number of CI nodes (currently 2), we decided to use a modulo strategy where we count each test parsing the output from `go test -v ./... -list=. -tags=integration`, assigning an index to each test prior to executing the tests. We then take that index and calculate the value of the node index it should execute on by computing the modulo of the total number of nodes. If the following is true, the test will execute on the given node. 

```go
CURRENT_TEST_INDEX % CI_NODE_TOTAL == CI_NODE_INDEX
```
`CURRENT_TEST_INDEX`: the index assigned to a particular test, from [0, X total number of tests]
`CI_NODE_TOTAL`: value is assigned in `ci.yml`, currently 2
`CI_NODE_INDEX`: the index of the node that's currently executing the tests, [0, CI_NODE_TOTAL - 1]

## Impact for Maintainers
Each top level test **must** call ` skipIfNotCINode(t)` before any other `skipIfX()` flag or test logic. If this is omitted from a test, the test will run across all nodes. Ideally, we shouldn't have this be manually called and should be part of a test harness, but we currently don't provide one for our integration tests. 

## Next Steps
We noticed some tests were unusually slow:
```
TestStateVersionsDownload (379.52s)
TestRunsCreate (8.50s)
TestRunsRead_CostEstimate (28.35s)
TestRunsApply (17.92s)
TestRunsDiscard (29.96s)
TestPlanExportsCreate (79.04s)
TestPlanExportsRead (55.31s)
TestPlanExportsDelete (85.68s)
TestPlanExportsDownload (42.42s)
TestPolicyChecksRead (15.95s)
TestPolicyChecksOverride/when_the_policy_failed (16.80s)
TestPolicyChecksOverride/when_the_policy_passed (16.33s)
TestLogReader_withoutMarkers (37.19s)
TestLogReader_withoutEndOfTextMarker (47.36s)
TestLogReader_withMarkersMulti (33.77s)
```
Future investigation will be needed to improve these long running tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: test
 on:
   pull_request:
-    branches: [ main ]
+    branches:
+      - main
 jobs:
   codegen:
     name: Codegen
@@ -42,6 +43,12 @@ jobs:
   tests:
     name: Integration Tests
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ci_node_total: [2]
+        ci_node_index: [0, 1]
+
     timeout-minutes: 30
     steps:
       - name: Checkout code
@@ -74,6 +81,8 @@ jobs:
           GITHUB_REGISTRY_MODULE_IDENTIFIER: "hashicorp/terraform-random-module"
           OAUTH_CLIENT_GITHUB_TOKEN: "${{ secrets.OAUTH_CLIENT_GITHUB_TOKEN }}"
           GO111MODULE: "on"
+          CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
+          CI_NODE_INDEX: ${{ matrix.ci_node_index }}
         run: |
           source $HOME/.env
           gotestsum --format short-verbose -- -timeout=40m -tags=integration
--- a/admin_organization_integration_test.go
+++ b/admin_organization_integration_test.go
@@ -13,6 +13,7 @@ import (
 )
 
 func TestAdminOrganizations_List(t *testing.T) {
+	skipIfNotCINode(t)
 	skipIfCloud(t)
 
 	client := testClient(t)
@@ -70,6 +71,7 @@ func TestAdminOrganizations_List(t *testing.T) {
 }
 
 func TestAdminOrganizations_Read(t *testing.T) {
+	skipIfNotCINode(t)
 	skipIfCloud(t)
 
 	client := testClient(t)
@@ -111,6 +113,7 @@ func TestAdminOrganizations_Read(t *testing.T) {
 }
 
 func TestAdminOrganizations_Delete(t *testing.T) {
+	skipIfNotCINode(t)
 	skipIfCloud(t)
 
 	client := testClient(t)
@@ -148,6 +151,7 @@ func TestAdminOrganizations_Delete(t *testing.T) {
 }
 
 func TestAdminOrganizations_ModuleConsumers(t *testing.T) {
+	skipIfNotCINode(t)
 	skipIfCloud(t)
 
 	client := testClient(t)
@@ -192,6 +196,7 @@ func TestAdminOrganizations_ModuleConsumers(t *testing.T) {
 }
 
 func TestAdminOrganizations_Update(t *testing.T) {
+	skipIfNotCINode(t)
 	skipIfCloud(t)
 
 	client := testClient(t)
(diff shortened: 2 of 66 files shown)


admin_organization_integration_test.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 func TestAdminOrganizations_List(t *testing.T) {
+	checkTestNodeEnv(t)
```
annawinkler, line 16: After reading the PR description about adding a `checkTestNodeEnv(t)` to all files, and seeing the impact, I'm wondering if it would be beneficial to explore another option for checking the test node environment. In the Ruby world, I would add a test class or module that does this checking, and all of the test classes would inherit from this new test class. I haven't done something similar in Go, so curious what your thoughts are about it? 🤔

annawinkler:
Really excited 🤩 about the performance improvements!! Nice job on the PR description too 💯

admin_organization_integration_test.go
sebasslash: Absolutely! While we can't exactly have top-level tests inherit some general behavior, we can potentially introduce our own "test suite" to go-tfe, which would reduce boilerplate setup across tests. This would require a pretty sizable refactor of our tests currently. 

An example might look like:
```go
func TestWorkspacesList(t *testing.T) {
   // setupCloudTest() 
   // calls skipIfTFE(), checkTestNodeEnv(), etc
   // creates a go-tfe client
   // calls createOrganization, t.Cleanup(orgCleanup), upgrades an organization
  client, org := setupCloudTest(t) 

  // or alternatively

   // setupTFETest() 
   // calls skipIfCloud(), checkTestNodeEnv(), etc
   // creates a go-tfe client
   // calls createOrganization(), t.Cleanup(orgCleanup)
  client, org := setupTFETest(t) 
  
  // run rest of test logic here
  
}
```

annawinkler: Could you add a follow on issue for us to refactor the go-tfe tests like you've described? I'm good with this PR addressing the test execution time, and then a subsequent PR addressing test refactoring.

annawinkler (dismissed):
Yay for faster tests! 🥳

Uk1288:
Awesome work!! 🎉 

Minor comment, let's update the update the PR description strategy to match the new modulus divisor strategy.

Also, I agree with @annawinkler comment regarding adding a generic initializer, I can see how one can miss adding `checkTestNodeEnv(t)` to a test.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,9 @@
 name: test
 on:
   pull_request:
-    branches: [ main ]
+    branches:
+      - main
+      - improve-test-execution-time
```
brandonc, line 6: This can be removed now

admin_organization_integration_test.go
brandonc, line 16: I would love if you had the word skip in this function name. skipIfNotCINode ?

brandonc: When you rebase, you should add this line to https://github.com/hashicorp/go-tfe/blob/main/scripts/generate_resource/templates.go#L165-L196

brandonc (changes requested):

.github/workflows/ci.yml
```diff
@@ -42,6 +43,12 @@ jobs:
   tests:
     name: Integration Tests
     runs-on: ubuntu-latest
+    strategy:
```
annawinkler, line 46: I think it would be helpful to add a comment about the modulo strategy so we know what we're looking at in the future.

brandonc (approved):

.github/workflows/ci.yml
sebasslash: Will add in a followup PR 👍

PR #537, opened 2022-09-26
Nightly go tfe tests

sebasslash:
## Description
This PR introduces a Github workflow that builds a TFE instance and runs our TFE-related integration tests by toggling `ENABLE_TFE=1`. This workflow is run nightly and does not affect our normal test runner or regular tflocal instance which is also recreated nightly. This workflow checks out the `tflocal/` working directory, creates an empty cloud backend configuration `backend.tf`, and triggers a run to create the instance using the non-interactive CLI workflow. To read more on configuring your `cloud` backend non-interactively, see [cloud environment variables](https://www.terraform.io/cli/cloud/settings#environment-variables). We are unable to use the expected VCS workflow due to a current limitation where runs are triggered on any PR merge even if there are no changes to your working directory. Since this workspace most of the time has "empty" state (due to our requirement for our instance being ephemeral), TFC thinks it needs to create the instance. If we persisted on using a VCS workflow, we'd have to find a way to discard each run triggered by a merge to `main` which is obviously more complicated. Once the run has completed, the tests job will fire off. 

Once the nightly tests finish (be it succeeding or failing) the cleanup job triggers a destroy run to remove the instance using the `build-tflocal` script.

These changes also include an update to the `build-tflocal` script our workflows depend on:
* Awaits the completion of a run before exiting. 
* Allow for the creation of a destroy run since it will be used in the cleanup phase of this workflow.
* Extend the interface for this script to support flags instead of relying on positional arguments. We can now pass `-o` for organization, `-w` for workspace, and `-d` to mark as destroy, including default values for each. 

Aside from this, the workflow is nearly identical to [ci.yml](https://github.com/hashicorp/go-tfe/blob/main/.github/workflows/ci.yml). 

## Testing (For Maintainers Only)
1. Ensure you have access to `hashicorp-v2` on prod and a user token available as `TFE_TOKEN`. Also double check there aren't any runs locking the `tflocal-go-tfe-nightly` workspace. 

2. To test creating the instance using `build-tflocal`
```sh
cd .github/scripts/build-tflocal \ 
  && go run . -w tflocal-go-tfe-nightly \ 
  && cd -
```
3. To test the `cleanup` job
```sh
cd .github/scripts/build-tflocal \ 
  && go run . -w tflocal-go-tfe-nightly -d \ 
  && cd -
```


.github/workflows/nightly-ci.yml
```diff
@@ -0,0 +1,126 @@
+name: test
```
annawinkler, line 1: What do you think about naming this file something like `nightly-tfe-ci.yml` so then it's clearer in the GH workflows UI what the workflow is for?

annawinkler:
Wonder why we've got test failures 🤔

annawinkler (dismissed):
:shipit:

sebasslash:
I can't merge this just yet. I realized that TFC is trying to trigger a run to create the nightly instance each time a PR is merged. This instance needs to be ephemeral, so I think using a VCS workflow might not accomplish what we're after. I need to rethink how we want to achieve creating this instance.

annawinkler:
> I can't merge this just yet. I realized that TFC is trying to trigger a run to create the nightly instance each time a PR is merged. This instance needs to be ephemeral, so I think using a VCS workflow might not accomplish what we're after. I need to rethink how we want to achieve creating this instance.

Hi @sebasslash ! I wasn't sure if this PR is ready for a rereview? Could you please indicate in the PR description what the problem was and what was rethought or reworked? Thanks!

annawinkler (approved):
:shipit: Awesome!

PR #539, opened 2022-09-29
Add safe-delete workspace API, and org setting to restrict force-delete

JarrettSpiker:
## Description

Adds the new workspace safe-delete API (and tests)

Adds the can-force-delete permission hash to the workspace permissions hash. This value is a `*bool` which will allow clients to detect if safe vs force delete is available

Adds the allow_force_delete_workspaces setting to organizations

These are all part of a change which adds some guardrails around workspace deletion in TFC. There will be a new safe-delete API to remove a workspace **only** if it does not have resources under management. The existing delete API will continue to function as a force delete, but may be restricted to org owners based on an org level setting

This change is in draft as we complete the feature in TFC and prepare for it in a TFE release

## Testing plan

1.  Test that Safe Delete works when a workspace does not have RUM, and fails when it does
2. Check that force delete still works
3. Check that can-force-delete is returned in workspace GETs
4. Check that can-force-delete is nil when communicating with older versions of TFE
5. Check that allow_force_delete is settable/readable through go-tfe

## External links

- [API documentation](https://github.com/hashicorp/terraform-docs-common/pull/146) (in progress)
- [RFC](https://hashidocs.hashicorp.services/document/1SwO0PmgIu2KxQl6vr7Z55v3veLjcHBbv4YdhjL67lBs)
- [JIRA](https://hashicorp.atlassian.net/browse/TF-719)

## Output from tests

```
❯ go test .  -run TestOrganizationsAllowForceDeleteSetting -tags=integration -v
=== RUN   TestOrganizationsAllowForceDeleteSetting
=== RUN   TestOrganizationsAllowForceDeleteSetting/creates_and_updates_allow_force_delete
--- PASS: TestOrganizationsAllowForceDeleteSetting (1.52s)
    --- PASS: TestOrganizationsAllowForceDeleteSetting/creates_and_updates_allow_force_delete (1.23s)
PASS
ok      github.com/hashicorp/go-tfe     1.658s

❯ go test .  -run TestWorkspacesSafeDelete -tags=integration -v 
=== RUN   TestWorkspacesSafeDelete
=== RUN   TestWorkspacesSafeDelete/with_valid_options
=== RUN   TestWorkspacesSafeDelete/when_organization_is_invalid
=== RUN   TestWorkspacesSafeDelete/when_workspace_is_invalid
=== RUN   TestWorkspacesSafeDelete/when_workspace_is_locked
=== RUN   TestWorkspacesSafeDelete/when_workspace_has_resources_under_management
--- PASS: TestWorkspacesSafeDelete (4.33s)
    --- PASS: TestWorkspacesSafeDelete/with_valid_options (0.35s)
    --- PASS: TestWorkspacesSafeDelete/when_organization_is_invalid (0.00s)
    --- PASS: TestWorkspacesSafeDelete/when_workspace_is_invalid (0.00s)
    --- PASS: TestWorkspacesSafeDelete/when_workspace_is_locked (0.78s)
    --- PASS: TestWorkspacesSafeDelete/when_workspace_has_resources_under_management (1.77s)
=== RUN   TestWorkspacesSafeDeleteByID
=== RUN   TestWorkspacesSafeDeleteByID/with_valid_options
=== RUN   TestWorkspacesSafeDeleteByID/without_a_valid_workspace_ID
=== RUN   TestWorkspacesSafeDeleteByID/when_workspace_is_locked
=== RUN   TestWorkspacesSafeDeleteByID/when_workspace_has_resources_under_management
--- PASS: TestWorkspacesSafeDeleteByID (3.99s)
    --- PASS: TestWorkspacesSafeDeleteByID/with_valid_options (0.34s)
    --- PASS: TestWorkspacesSafeDeleteByID/without_a_valid_workspace_ID (0.00s)
    --- PASS: TestWorkspacesSafeDeleteByID/when_workspace_is_locked (0.78s)
    --- PASS: TestWorkspacesSafeDeleteByID/when_workspace_has_resources_under_management (1.39s)
PASS
ok      github.com/hashicorp/go-tfe     8.436s

❯ go test .  -run TestWorkspacesDelete -tags=integration -v 
=== RUN   TestWorkspacesDelete
=== RUN   TestWorkspacesDelete/with_valid_options
=== RUN   TestWorkspacesDelete/when_organization_is_invalid
=== RUN   TestWorkspacesDelete/when_workspace_is_invalid
--- PASS: TestWorkspacesDelete (1.61s)
    --- PASS: TestWorkspacesDelete/with_valid_options (0.35s)
    --- PASS: TestWorkspacesDelete/when_organization_is_invalid (0.00s)
    --- PASS: TestWorkspacesDelete/when_workspace_is_invalid (0.00s)
=== RUN   TestWorkspacesDeleteByID
=== RUN   TestWorkspacesDeleteByID/with_valid_options
=== RUN   TestWorkspacesDeleteByID/without_a_valid_workspace_ID
--- PASS: TestWorkspacesDeleteByID (1.66s)
    --- PASS: TestWorkspacesDeleteByID/with_valid_options (0.34s)
    --- PASS: TestWorkspacesDeleteByID/without_a_valid_workspace_ID (0.00s)
PASS
ok      github.com/hashicorp/go-tfe     3.378s

❯ go test .  -run TestCanForceDeletePermission -tags=integration -v
=== RUN   TestCanForceDeletePermission
=== RUN   TestCanForceDeletePermission/workspace_permission_set_includes_can-force-delete
--- PASS: TestCanForceDeletePermission (1.59s)
    --- PASS: TestCanForceDeletePermission/workspace_permission_set_includes_can-force-delete (0.26s)
PASS
ok      github.com/hashicorp/go-tfe     1.704s
...
```

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -1599,6 +1599,12 @@ func createWorkspaceWithOptions(t *testing.T, client *Client, org *Organization,
 
 	return w, func() {
 		if err := client.Workspaces.Delete(ctx, org.Name, w.Name); err != nil {
+
+			// if the workspace has already been cleaned up, or destroyed in the testing process, no need to raise the alarm
+			if err == ErrResourceNotFound {
+				return
+			}
+
 			t.Errorf("Error destroying workspace! WARNING: Dangling resources\n"+
 				"may exist! The full error is shown below.\n\n"+
 				"Workspace: %s\nError: %s", w.Name, err)
--- a/helper_test.go
+++ b/helper_test.go
@@ -1598,16 +1598,10 @@ func createWorkspaceWithOptions(t *testing.T, client *Client, org *Organization,
 	}
 
 	return w, func() {
-		if err := client.Workspaces.Delete(ctx, org.Name, w.Name); err != nil {
-
-			// if the workspace has already been cleaned up, or destroyed in the testing process, no need to raise the alarm
-			if err == ErrResourceNotFound {
-				return
-			}
-
+		if err := client.Workspaces.DeleteByID(ctx, w.ID); err != nil {
 			t.Errorf("Error destroying workspace! WARNING: Dangling resources\n"+
 				"may exist! The full error is shown below.\n\n"+
-				"Workspace: %s\nError: %s", w.Name, err)
+				"Workspace: %s\nError: %s", w.ID, err)
 		}
 
 		if orgCleanup != nil {
--- a/organization.go
+++ b/organization.go
@@ -78,6 +78,7 @@ type Organization struct {
 	TrialExpiresAt                                    time.Time                `jsonapi:"attr,trial-expires-at,iso8601"`
 	TwoFactorConformant                               bool                     `jsonapi:"attr,two-factor-conformant"`
 	SendPassingStatusesForUntriggeredSpeculativePlans bool                     `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
+	AllowForceDeleteWorkspaces                        bool                     `jsonapi:"attr,allow-force-delete-workspaces"`
 }
 
 // Capacity represents the current run capacity of an organization.
@@ -166,6 +167,9 @@ type OrganizationCreateOptions struct {
 
 	// Optional: SendPassingStatusesForUntriggeredSpeculativePlans toggles behavior of untriggered speculative plans to send status updates to version control systems like GitHub.
 	SendPassingStatusesForUntriggeredSpeculativePlans *bool `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans,omitempty"`
+
+	// Optional: AllowForceDeleteWorkspaces toggles behavior of allowing workspace admins to delete workspaces with resources under management.
+	AllowForceDeleteWorkspaces *bool `jsonapi:"attr,allow-force-delete-workspaces,omitempty"`
 }
 
 // OrganizationUpdateOptions represents the options for updating an organization.
@@ -202,6 +206,9 @@ type OrganizationUpdateOptions struct {
 
 	// SendPassingStatusesForUntriggeredSpeculativePlans toggles behavior of untriggered speculative plans to send status updates to version control systems like GitHub.
 	SendPassingStatusesForUntriggeredSpeculativePlans *bool `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans,omitempty"`
+
+	// Optional: AllowForceDeleteWorkspaces toggles behavior of allowing workspace admins to delete workspaces with resources under management.
+	AllowForceDeleteWorkspaces *bool `jsonapi:"attr,allow-force-delete-workspaces,omitempty"`
 }
 
 // ReadRunQueueOptions represents the options for showing the queue.
--- a/organization.go
+++ b/organization.go
@@ -78,7 +78,9 @@ type Organization struct {
 	TrialExpiresAt                                    time.Time                `jsonapi:"attr,trial-expires-at,iso8601"`
 	TwoFactorConformant                               bool                     `jsonapi:"attr,two-factor-conformant"`
 	SendPassingStatusesForUntriggeredSpeculativePlans bool                     `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
-	AllowForceDeleteWorkspaces                        bool                     `jsonapi:"attr,allow-force-delete-workspaces"`
+	// Note: This will be false for TFE versions older than v202211, where the setting was introduced.
+	// On those TFE versions, safe delete does not exist, so ALL deletes will be force deletes.
+	AllowForceDeleteWorkspaces bool `jsonapi:"attr,allow-force-delete-workspaces"`
 }
 
 // Capacity represents the current run capacity of an organization.
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -1211,6 +1211,126 @@ func TestWorkspacesDeleteByID(t *testing.T) {
 	})
 }
 
+func TestCanForceDeletePermission(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, _ := createWorkspace(t, client, orgTest)
+
+	t.Run("workspace permission set includes can-force-delete", func(t *testing.T) {
+		w, err := client.Workspaces.ReadByID(ctx, wTest.ID)
+		require.NoError(t, err)
+		assert.Equal(t, wTest, w)
+		assert.Equal(t, wTest, w)
+		assert.NotNil(t, w.Permissions.CanForceDelete)
+		assert.True(t, *w.Permissions.CanForceDelete)
+	})
+}
+
+func TestWorkspacesSafeDelete(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, _ := createWorkspace(t, client, orgTest)
+
+	t.Run("with valid options", func(t *testing.T) {
+		err := client.Workspaces.SafeDelete(ctx, orgTest.Name, wTest.Name)
+		require.NoError(t, err)
+
+		// Try loading the workspace - it should fail.
+		_, err = client.Workspaces.Read(ctx, orgTest.Name, wTest.Name)
+		assert.Equal(t, ErrResourceNotFound, err)
+	})
+
+	t.Run("when organization is invalid", func(t *testing.T) {
+		err := client.Workspaces.SafeDelete(ctx, badIdentifier, wTest.Name)
+		assert.EqualError(t, err, ErrInvalidOrg.Error())
+	})
+
+	t.Run("when workspace is invalid", func(t *testing.T) {
+		err := client.Workspaces.SafeDelete(ctx, orgTest.Name, badIdentifier)
+		assert.EqualError(t, err, ErrInvalidWorkspaceValue.Error())
+	})
+
+	t.Run("when workspace is locked", func(t *testing.T) {
+		wTest, workspaceCleanup := createWorkspace(t, client, orgTest)
+		defer workspaceCleanup()
+		w, err := client.Workspaces.Lock(ctx, wTest.ID, WorkspaceLockOptions{})
+		assert.NoError(t, err)
+		assert.True(t, w.Locked)
+
+		err = client.Workspaces.SafeDelete(ctx, orgTest.Name, wTest.Name)
+		assert.Contains(t, err.Error(), "conflict")
+		assert.Contains(t, err.Error(), "currently locked")
+	})
+
+	t.Run("when workspace has resources under management", func(t *testing.T) {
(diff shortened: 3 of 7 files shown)


brandonc (approved):

organization.go
```diff
@@ -78,6 +78,7 @@ type Organization struct {
 	TrialExpiresAt                                    time.Time                `jsonapi:"attr,trial-expires-at,iso8601"`
 	TwoFactorConformant                               bool                     `jsonapi:"attr,two-factor-conformant"`
 	SendPassingStatusesForUntriggeredSpeculativePlans bool                     `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
+	AllowForceDeleteWorkspaces                        bool                     `jsonapi:"attr,allow-force-delete-workspaces"`
```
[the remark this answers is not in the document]
JarrettSpiker: That is a really good point...I am also ok with the contradiction, it seems less confusing than making this a `*bool` or something when it isn't strictly necessary and we use regular `bool`s are used for everything else in this struct. I also cant think of a less confusing name...

I will add a comment though to make it clear that this will be false for older TFE versions, even though all deletes are force deletes in that case

workspace_integration_test.go
```diff
@@ -1186,6 +1186,126 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, _ := createWorkspace(t, client, orgTest)
+
+	t.Run("workspace permission set includes can-force-delete", func(t *testing.T) {
+		w, err := client.Workspaces.ReadByID(ctx, wTest.ID)
+		require.NoError(t, err)
+		assert.Equal(t, wTest, w)
+		assert.Equal(t, wTest, w)
```
SwiftEngineer, line 1202: duplicate assertion here

SwiftEngineer (commented):
Only thing I saw was that little duplicate assertion, apart from that I'd say this PR looks great 💯

workspace_integration_test.go
JarrettSpiker: good catch, thanks!

workspace_integration_test.go
```diff
@@ -1211,6 +1211,125 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
 }
 
+func TestCanForceDeletePermission(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	wTest, _ := createWorkspace(t, client, orgTest)
```
sebasslash, line 1221: Hm, Is there a reason we don't clean this workspace up?

workspace_integration_test.go
```diff
@@ -1211,6 +1211,125 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
+		err = client.Workspaces.SafeDeleteByID(ctx, wTest.ID)
+		assert.Contains(t, err.Error(), "conflict")
+		assert.Contains(t, err.Error(), "currently locked")
+	})
+
+	t.Run("when workspace has resources under management", func(t *testing.T) {
+		wTest, workspaceCleanup := createWorkspace(t, client, orgTest)
+		defer workspaceCleanup()
+		_, svTestCleanup := createStateVersion(t, client, 0, wTest)
+		t.Cleanup(svTestCleanup)
```
sebasslash, line 1324: Non-blocker: Not a huge deal to mix the two, but the preference here is for `t.Cleanup()`

sebasslash (changes requested):
Overall looks great :+1: . There are a few minor blocking issues concerning your tests. I also wanted to mention the changelog is a bit out of order at the moment and will be resolved in #549

workspace_integration_test.go
JarrettSpiker: Do you mean that it is preferred to do `t.Cleanup(workspaceCleanup)` instead of `defer workspaceCleanup()` directly?

workspace_integration_test.go
JarrettSpiker: I believe I was copying from the [update tests in this file](https://github.com/hashicorp/go-tfe/blob/main/workspace_integration_test.go#L740), which also ignore the cleanup

Im not sure why though...I will try to update all the tests to do the cleanup

JarrettSpiker: Oh, I remember. Because the the workspace gets removed in the process of the test, and then the cleanup call throws an errors. Or, in some other tests (and also this one, I would guess) the org also gets cleaned up first, which deletes the workspace and causes a defered workspace cleanup to fail

E.g. if you try to defer the cleanup in the test I linked above you get

```
=== RUN   TestWorkspacesUpdate
=== CONT  TestWorkspacesUpdate
    helper_test.go:1602: Error destroying workspace! WARNING: Dangling resources
        may exist! The full error is shown below.
        
        Workspace: 0874350d-2197-1355-74c7-4ee437bbc010
        Error: resource not found
--- FAIL: TestWorkspacesUpdate (5.32s)
```

JarrettSpiker: That feels like it should be a solve-able problem. I will update the workspace cleanup function not to error on 404s, and then update the other tests in this file to cleanup the workspaces which they create

helper_test.go
```diff
@@ -1599,6 +1599,12 @@ func createWorkspaceWithOptions(t *testing.T, client *Client, org *Organization,
 
 	return w, func() {
 		if err := client.Workspaces.Delete(ctx, org.Name, w.Name); err != nil {
+
+			// if the workspace has already been cleaned up, or destroyed in the testing process, no need to raise the alarm
+			if err == ErrResourceNotFound {
+				return
+			}
```
sebasslash, line 1606: Ah, a recurring pain point for us. I tried introducing this in #527 but it was ultimately rejected considering it could obfuscate an authorization bug. So unfortunately will have to reject it here -- I think we can omit the cleanup for now if an "already deleted" workspace is causing this error.

workspace_integration_test.go
```diff
@@ -1211,6 +1218,134 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
+	t.Run("when workspace is invalid", func(t *testing.T) {
+		err := client.Workspaces.SafeDelete(ctx, orgTest.Name, badIdentifier)
+		assert.EqualError(t, err, ErrInvalidWorkspaceValue.Error())
+	})
+
+	t.Run("when workspace is locked", func(t *testing.T) {
+		wTest, workspaceCleanup := createWorkspace(t, client, orgTest)
+		t.Cleanup(workspaceCleanup)
+		w, err := client.Workspaces.Lock(ctx, wTest.ID, WorkspaceLockOptions{})
+		assert.NoError(t, err)
```
sebasslash, line 1277: missed one `require.NoError(t, err)`, I'd even say we should also `require.True(t, w.Locked)` since the rest of your test depends on this being true. If this fails, there's no point in continuing to run the rest of the test.

workspace_integration_test.go
```diff
@@ -1211,6 +1218,134 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
+	t.Run("without a valid workspace ID", func(t *testing.T) {
+		err := client.Workspaces.SafeDeleteByID(ctx, badIdentifier)
+		assert.EqualError(t, err, ErrInvalidWorkspaceID.Error())
+	})
+
+	t.Run("when workspace is locked", func(t *testing.T) {
+		wTest, workspaceCleanup := createWorkspace(t, client, orgTest)
+		t.Cleanup(workspaceCleanup)
+		w, err := client.Workspaces.Lock(ctx, wTest.ID, WorkspaceLockOptions{})
+		assert.NoError(t, err)
```
sebasslash, line 1328: Same here.

sebasslash (changes requested):
Almost at the finish line :runner: :arrow_down:

workspace_integration_test.go
```diff
@@ -1211,6 +1218,134 @@ func TestWorkspacesDeleteByID(t *testing.T) {
...
+
+	wTest, wCleanup := createWorkspace(t, client, orgTest)
+	t.Cleanup(wCleanup)
+
+	t.Run("workspace permission set includes can-force-delete", func(t *testing.T) {
+		w, err := client.Workspaces.ReadByID(ctx, wTest.ID)
+		require.NoError(t, err)
+		assert.Equal(t, wTest, w)
+		assert.NotNil(t, w.Permissions.CanForceDelete)
+		assert.True(t, *w.Permissions.CanForceDelete)
```
sebasslash, line 1238: Looks like your test is panicking:
```
=== FAIL: . TestCanForceDeletePermission/workspace_permission_set_includes_can-force-delete (0.13s)
    workspace_integration_test.go:1237: 
        	Error Trace:	/home/runner/work/go-tfe/go-tfe/workspace_integration_test.go:1237
        	Error:      	Expected value not to be nil.
        	Test:       	TestCanForceDeletePermission/workspace_permission_set_includes_can-force-delete
    --- FAIL: TestCanForceDeletePermission/workspace_permission_set_includes_can-force-delete (0.13s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7f0a47]

goroutine 2995 [running]:
testing.tRunner.func1.2({0x861640, 0xbf9470})
	/opt/hostedtoolcache/go/1.19.2/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.2/x64/src/testing/testing.go:1399 +0x39f
panic({0x861640, 0xbf9470})
	/opt/hostedtoolcache/go/1.19.2/x64/src/runtime/panic.go:884 +0x212
github.com/hashicorp/go-tfe.TestCanForceDeletePermission.func1(0x1?)
	/home/runner/work/go-tfe/go-tfe/workspace_integration_test.go:1238 +0xe7
testing.tRunner(0xc0004b56c0, 0xc0006b2780)
	/opt/hostedtoolcache/go/1.19.2/x64/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.2/x64/src/testing/testing.go:1493 +0x35f
exit status 2
``` 

Two things:
* This very well may be an issue with our CI instance not having the latest API changes. In which case we'd need to fix this. 
* We tend to require the existence of a relationship struct, i.e `require.NotNil(t, w.Permissions)`, before we check said relationship's fields (otherwise we risk a panic).

helper_test.go
JarrettSpiker: Interesting...I think I would lean the other direction on that, since a test creating a workspace and then being unable to "see" it when it goes for a delete seems unlikely, but I definitely see the logic

I removed the `ErrResourceNotFound` check, and switched a lot of the tests to using `t.Cleanup` instead of `defer` for the cleanup callbacks, and that does seem to have cleared up most of the cleanup errors~

I also changed this to delete the workspace by ID instead of Name, because we had one test which changes the workspace name (causing this cleanup to fail)

workspace_integration_test.go
JarrettSpiker: > This very well may be an issue with our CI instance not having the latest API changes. In which case we'd need to fix this.

I think  that this is at least partially it...where can I find the instance that the acceptance tests run against? I can check out if it has all the api changes expected

sebasslash (approved):
:rocket: Looks like there are some flakes unrelated to this PR but still approving.

PR #541, opened 2022-09-29
Balance tests across CI_NODE_TOTAL using go test -list

brandonc:


helper_test.go
```diff
@@ -1924,27 +1925,60 @@ func randomSemver(t *testing.T) string {
...
+	result := make(map[string]int)
+	index := 0
+	for _, line := range strings.Split(string(output), "\n") {
+		if strings.HasPrefix(line, "Test") {
+			result[line] = index
+			index += 1
+		}
 	}
-	return "3"
+	return result
```
sebasslash, line 1981: I just ran `go test -v ./... -list=. -tags=integration` and it is a tad bit slow. Alternatively, we can add the assignment for `testNames` to `init()` so it runs before any tests are run: 
```go
func init() {
  testNames = listTestsCI()
}
```
and remove `t.Helper()`. What do you think?

brandonc: To be clear I should only be running it once per node because it's being assigned to a package variable

brandonc: I think init is a better approach. I'll modify it.

sebasslash (approved):
FWIW, I **really** like this change since it saves us from having to rebalance tests manually. 

Good with me 🚀 . There's an odd failure going on with the codegen workflow 😅

PR #543, opened 2022-09-29
refactor StateVersionsCurrent integration test

laurenolivia:
## Description

Test flaking because the StateVersion and TerraformVersion fields were asserted to be equal. These fields are populated by the API, and are not required to match for the purpose of this test.

## Testing plan

N/A

## External links

N/A

## Output from tests

```

...
```


sebasslash (approved):
:shipit:

brandonc (approved):

PR #545, opened 2022-10-05
[CI] Fix: remove inline terraform block 

sebasslash:
Terraform expects the `cloud` block to be defined in its own line:
```
A single-line block definition can contain only a
single argument. If you meant to define argument "cloud", use an equals
sign to assign it a value. To define a nested block, place it on a line of
its own within its parent block.
```

Also remove the logic to await a run's completion since we are now using Terraform to initiate the run.

### DIFF
--- a/.github/scripts/build-tflocal/main.go
+++ b/.github/scripts/build-tflocal/main.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"os"
 	"os/signal"
-	"time"
 
 	tfe "github.com/hashicorp/go-tfe"
 )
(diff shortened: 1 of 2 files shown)


.github/scripts/build-tflocal/main.go
```diff
@@ -45,9 +44,7 @@ func main() {
 		log.Fatal(err)
 	}
 
-	if err := waitForRun(ctx, client, runID); err != nil {
-		log.Fatal(err)
-	}
+	log.Printf(runID)
```
Uk1288, line 47: is there a reason for printing just the runID here?

sebasslash: We don't necessarily have to, but it makes it easier to troubleshoot and associate a workflow to a specific run.

annawinkler (approved):
🥳

PR #546, opened 2022-10-06
Remove duplicate cloud block

sebasslash:
I completely missed that we already have the `cloud` block defined in [tflocal/terraform.tf](https://github.com/hashicorp/go-tfe/blob/2f3fd5b9fe528649bb46c53aacc236f1536f0c16/tflocal/terraform.tf#L26). 

This PR removes the duplicate `backend.tf` file created for the workflow.


annawinkler (approved):
👍

PR #550, opened 2022-10-14
Add `Query` and `Status` query param fields to `OrganizationMembershipListOptions`

sebasslash:
## Description

Adds two missing fields supported by our [List Organization Memberships API](https://www.terraform.io/cloud-docs/api-docs/organization-memberships#list-memberships-for-an-organization):
* `q`: A search query string. Organization memberships are searchable by user name and email.
* `filter[status]`:  If specified, restricts results to those with the matching status value. Valid values are `invited` and `active`

One thing to note is that we skip one of the query string tests because it depends on being able to assign a username to the user of an organization membership. A test like this requires [Account API support](https://www.terraform.io/cloud-docs/api-docs/account#update-your-account-info), which go-tfe currently does not support. Once supported, we can update the skipped test to be able to dynamically assign usernames and thus make them query-able.  

## Testing plan

`go test -v ./... -run TestOrganizationMembershipsList -tags=integration`

## External links

- [Query params supported by List Organization Memberships endpoint. ](https://www.terraform.io/cloud-docs/api-docs/organization-memberships#query-parameters)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# (Unreleased)
+
+## Enhancements
+* Adds `Query` and `Status` fields to `OrganizationMembershipListOptions` to allow filtering memberships by status or username by @sebasslash [#550](https://github.com/hashicorp/go-tfe/pull/550)
+
 # v1.10.0
 
 ## Enhancements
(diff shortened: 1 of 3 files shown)


nfagerlund (approved):
LGTM! At first I wasn't sure what you meant about not being able to set username, but I did a couple tests and now I get it — we straight-up can't create users from these tests yet. Ah well!

PR #520, opened 2022-08-30
Add method to list variable sets by workspace

tstapler:
## Description

The Variable Sets api supports listing VariableSets using a WorkspaceID. This is useful if you only want to select the VariableSets used by an API.

## Testing plan

1. Create a Workspace and add a Variable set to it.
2. Call ListForWorkspace using the workspaceID.

## External links

https://www.terraform.io/enterprise/api-docs/variable-sets#list-variable-set

PR #551, opened 2022-10-17
List Variable Sets associated with workspace

sebasslash:
## Description

These changes build off of #520, namely I'm making a slight modification on which interface this method belongs to.

In the original PR, the method to list variable sets for a workspace was `VariableSets.ListForWorkspaces()` and I've changed it to: `Workspaces.ListVariableSets()`. Given we typically have children relationships exposed as part of the parent interface, i.e `Workspaces.ListSomeChildren()`, I felt this API was more natural in describing the one-to-many relationship.

It is worth noting that this endpoint is [documented under our Variable Set API](https://www.terraform.io/cloud-docs/api-docs/variable-sets#list-variable-set), however the path is `/workspaces/:workspace_id/varsets` so I felt like this is ultimately part of the Workspaces API. 

I've also added the `applyVariableSetToWorkspace()` test helper, which was missing in the original PR. 

Besides this, I've kept the original implementation from @tstapler (Great work if you see this!) 
 
## Testing plan

```sh
go test -v ./... -run TestWorkspacesListVariableSets -tags=integration
```

## External links

- [Original PR](https://github.com/hashicorp/go-tfe/pull/520)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Enhancements
+
+* Add `ListVariableSets` method to `Workspaces` interface to enable fetching variable sets associated with a workspace. [#551](https://github.com/hashicorp/go-tfe/pull/551)
+
 # v1.10.0
 
 ## Enhancements
--- a/helper_test.go
+++ b/helper_test.go
@@ -1742,6 +1742,34 @@ func createVariableSet(t *testing.T, client *Client, org *Organization, options 
 	}
 }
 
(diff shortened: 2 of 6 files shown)


brandonc (changes requested):
I think my preference is to attach this to the VariableSets interface. Here is my rationale: Nearly everything belongs to a workspace, but the non-core interface we offer in go-tfe seems to scoped to workspace _settings_ like tags, ssh keys, and remote state consumers.

Other things associated to workspaces (config versions, state versions, notifications, variables, run tasks) don't extend the workspace interface to add support.

sebasslash:
That's a fair point. I'll close this PR and rebase again.

PR #552, opened 2022-10-17
List workspace variable sets

sebasslash:
## Description

Build off of #520 and #551, maintaining the original interface but adding missing test helper to apply a variable set to a workspace (otherwise test fails). 

This also fixes the assertion that a given variable set is contained within the list fetched from workspace. `assert.Contains()` is a tricky call because it performs deep equality. A resource may or may not have relationships defined depending on the endpoint you're calling, in this case:
```go
// when we create a Variable Set
vsTest := client.VariableSets.Create() //omit params
vsTest.Workspaces == [] // true

// after applying vsTest to a workspace
vsList := client.VariableSets.ListForWorkspace()
vsTest.Items[0].Workspaces == [] // false

// therefore
assert.Contains(t, vsList.Items, vsTest) // false
```

### DIFF
--- a/variable_set.go
+++ b/variable_set.go
@@ -17,6 +17,9 @@ type VariableSets interface {
 	// List all the variable sets within an organization.
 	List(ctx context.Context, organization string, options *VariableSetListOptions) (*VariableSetList, error)
 
+	// ListForWorkspace gets the associated variable sets for a workspace.
+	ListForWorkspace(ctx context.Context, workspaceID string, options *VariableSetListOptions) (*VariableSetList, error)
+
 	// Create is used to create a new variable set.
 	Create(ctx context.Context, organization string, options *VariableSetCreateOptions) (*VariableSet, error)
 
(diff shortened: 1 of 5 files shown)


variable_set.go
```diff
@@ -179,6 +182,32 @@ func (s *variableSets) List(ctx context.Context, organization string, options *V
 	return vl, nil
 }
 
+// ListForWorkspace gets the associated variable sets for a workspace.
+func (s *variableSets) ListForWorkspace(ctx context.Context, workspaceID string, options *VariableSetListOptions) (*VariableSetList, error) {
+	if !validStringID(&workspaceID) {
+		return nil, ErrInvalidOrg
```
Uk1288, line 188: should this be `ErrInvalidWorkspaceID`?

variable_set.go
```diff
@@ -17,6 +17,9 @@ type VariableSets interface {
 	// List all the variable sets within an organization.
 	List(ctx context.Context, organization string, options *VariableSetListOptions) (*VariableSetList, error)
 
+	// ListForWorkspace gets the associated variable sets for a workspace.
+	ListForWorkspace(ctx context.Context, workspaceId string, options *VariableSetListOptions) (*VariableSetList, error)
```
Uk1288, line 21: 💅🏾 change workspaceId to `workspaceID`?

variable_set_test.go
```diff
@@ -51,9 +51,63 @@ func TestVariableSetsList(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+		assert.Empty(t, vsl.Items)
+		assert.Equal(t, 999, vsl.CurrentPage)
+		assert.Equal(t, 2, vsl.TotalCount)
+	})
+
+	t.Run("when Workspace ID is an invalid ID", func(t *testing.T) {
+		vsl, err := client.VariableSets.ListForWorkspace(ctx, badIdentifier, nil)
+		assert.Nil(t, vsl)
```
Uk1288, line 110: test assertion `ErrInvalidWorkspaceID` instead of ErrInvalidOrg

Uk1288 (commented):
Looking good 🎉

Uk1288 (approved):
🎉

PR #563, opened 2022-10-25
Add X-TFE-Version header

sebasslash:
## Description

Since TFE Version 202208-3, we've included the TFE version string as a header for each response. This PR adds support for the client to retrieve that header during the initial metadata request and cache it locally. That version string can be fetched using the new `Client` method: `RemoteTFEVersion()`

## Testing plan

You can run the integration test:

```sh
go test -run TestClient_newClient -v ./... -tags=integration
```

Optionally, you can test against an actual TFE instance if you have one at your disposal:

```go

func main() {
  ctx := context.Background()
  cfg := &tfe.Config{
    Address: "https://your-tfe-address",
    Token: "your-token",
    RetryServerErrors: true,
  }
  
  client, err := tfe.NewClient(cfg)
  if err != nil {
    // sound the alarm!
  }
  
  if version := client.RemoteTFEVersion(); version == "" {
    panic("sound the alarm again!")
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Enhancements
 
 * Add `NotificationTriggerAssessmentCheckFailed` notification trigger type by @rexredinger [#549](https://github.com/hashicorp/go-tfe/pull/549)
+* Add `RemoteTFEVersion()` to the `Client` interface, which exposes the `X-TFE-Version` header set by a remote TFE instance by @sebasslash [#563](https://github.com/hashicorp/go-tfe/pull/563)
 
 # v1.11.0
 
--- a/tfe_integration_test.go
+++ b/tfe_integration_test.go
@@ -27,6 +27,7 @@ func TestClient_newClient(t *testing.T) {
 		w.Header().Set("Content-Type", "application/vnd.api+json")
 		w.Header().Set("X-RateLimit-Limit", "30")
 		w.Header().Set("TFP-API-Version", "34.21.9")
+		w.Header().Set("X-TFE-Version", "202205-1")
 		w.WriteHeader(204) // We query the configured ping URL which should return a 204.
 	}))
 	defer ts.Close()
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Enhancements
 
 * Add `NotificationTriggerAssessmentCheckFailed` notification trigger type by @rexredinger [#549](https://github.com/hashicorp/go-tfe/pull/549)
+* Add `RemoteTFEVersion()` to the `Client` interface, which exposes the `X-TFE-Version` header set by a remote TFE instance by @sebasslash []
```
uturunku1, line 6: * Add `RemoteTFEVersion()` to the `Client` interface, which exposes the `X-TFE-Version` header set by a remote TFE instance by @sebasslash [#563](https://github.com/hashicorp/go-tfe/pull/563)
```suggestion
* Add `RemoteTFEVersion()` to the `Client` interface, which exposes the `X-TFE-Version` header set by a remote TFE instance by @sebasslash []
```

tfe_integration_test.go
```diff
@@ -27,6 +27,7 @@ func TestClient_newClient(t *testing.T) {
 		w.Header().Set("Content-Type", "application/vnd.api+json")
 		w.Header().Set("X-RateLimit-Limit", "30")
 		w.Header().Set("TFP-API-Version", "34.21.9")
+		w.Header().Set("X-TFE-Version", "202205-1")
```
annawinkler, line 30: What do you think about testing for the case where we have TFC?

uturunku1:
@annawinkler and I got together today, and we partially tested your PR. We were able to verify the proper behavior for TFC (an empty string returned by the method) by using your code sample. However, we were not able to test the behavior for a TFE instance because we do have access to one. We requested one in the # proj-tfe-releases channel, but the request is pending.

tfe_integration_test.go
sebasslash: The issue with this is that I'd have rewrite most of the test simply to test that a field is an empty string (the test isn't really operating under a TFC vs TFE context).  

One nice guarantee we have is that the function that's used to fetch the header `http.Response.Header.Get()` returns a string and the docs mention that if a key does not exist, it will return an empty string. So by testing that we can fetch the value of that header we are also inadvertently testing that it can potentially be empty as well since we know that header can be omitted.

annawinkler (approved):
✅ Tested with latest tfe test instance and test script and got `v202211-1`

:shipit:

PR #564, opened 2022-10-25
Add CRUD support for Projects API

hs26gill:
## Description

A new `Projects` API is added to allow users to organize and centrally manage their workspaces at scale.

## Testing plan

To run the integration test:

```
ENABLE_BETA=1 go test -run TestProjectsRead  -v ./... -tags=integration

ENABLE_BETA=1 go test -run TestProjectsCreate  -v ./... -tags=integration

ENABLE_BETA=1 go test -run TestProjectsUpdate -v ./... -tags=integration

ENABLE_BETA=1 go test -run TestProjectsDelete  -v ./... -tags=integration

ENABLE_BETA=1 go test -run TestProjectsList  -v ./... -tags=integration
```

---
  - https://app.asana.com/0/0/1203185457641102
  - https://app.asana.com/0/0/1203185457641104
  - https://app.asana.com/0/0/1203185457641103


projects.go
```diff
@@ -0,0 +1,216 @@
...
+
+// ProjectCreateOptions represents the options for creating a project
+type ProjectCreateOptions struct {
+	// Type is a public field utilized by JSON:API to
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,projects"`
+
+	// Required: A name to identify the agent pool.
```
jbonhag, line 66: Some typos: `agent pool` => `project`

tfe.go
```diff
@@ -403,6 +404,7 @@ func NewClient(cfg *Config) (*Client, error) {
 	client.VariableSetVariables = &variableSetVariables{client: client}
 	client.WorkspaceRunTasks = &workspaceRunTasks{client: client}
 	client.Workspaces = &workspaces{client: client}
+	client.Projects = &projects{client: client}
```
jbonhag, line 407: Tiny nit, but it seems like this list is alphabetized.

workspace_integration_test.go
```diff
@@ -2097,3 +2097,37 @@ func TestWorkspacesRunTasksPermission(t *testing.T) {
...
+	})
+
+	t.Run("list workspace includes project ID", func(t *testing.T) {
+		workspaces, err := client.Workspaces.List(ctx, orgTest.Name, &WorkspaceListOptions{})
+		assert.NoError(t, err)
+		for idx, item := range workspaces.Items {
+			assert.NotNil(t, item.Project.ID, "No project ID set on workspace %s at idx %d", item.ID, idx)
+		}
+	})
+
```
jbonhag, line 2132: Should we add tests here for creating a workspace in a particular project? Or is that out of scope?

projects.go
```diff
@@ -0,0 +1,216 @@
+package tfe
```
jbonhag, line 1: I noticed that the standard for the resource file naming seems to be singular (`project.go`, `project_integration_test.go`) instead of plural.

jbonhag (commented):
Nicely done -- a couple of nitpick/naming-related things but it seems to be in good shape!

workspace_integration_test.go
hs26gill: Added new tests for workspace associated with projects

project.go
```diff
@@ -0,0 +1,216 @@
...
+// ProjectCreateOptions represents the options for creating a project
+type ProjectCreateOptions struct {
+	// Type is a public field utilized by JSON:API to
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,projects"`
+
+	// Required: A name to identify the project.
+	Name *string `jsonapi:"attr,name"`
```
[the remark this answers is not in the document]
sebasslash: If we set a field as a pointer, it should either:
* Be a nullable field
* Be omitted with `omitempty`

We shouldn't expose the field as a pointer and then silently validate `if option.Name == nil { return error }`.

project.go
```diff
@@ -0,0 +1,217 @@
...
+
+// ProjectUpdateOptions represents the options for updating a project
+type ProjectUpdateOptions struct {
+	// Type is a public field utilized by JSON:API to
+	// set the resource type via the field tag.
+	// It is not a user-defined value and does not need to be set.
+	// https://jsonapi.org/format/#crud-creating
+	Type string `jsonapi:"primary,projects"`
+
+	// Optional: A name to identify the agent pool.
```
brandonc, line 78:
```suggestion
	// Optional: A name to identify the project
```

brandonc (changes requested):
All this looks great. My one bit of blocking feedback is that all the new go docs need to be clearly labeled as beta, and the CHANGELOG should make it clear that the features are not generally available and subject to change.

Ex:

```
type Projects interface
    // List all projects in the given organization
    // **Note: This field is still in BETA and subject to change.**
    List(ctx...)
...
}
```

See the [Adding API Changes that are Not Generally Available section](https://github.com/hashicorp/go-tfe/blob/main/docs/CONTRIBUTING.md#adding-api-changes-that-are-not-generally-available) of CONTRIBUTING.md for details.

brandonc (approved):

PR #568, opened 2022-10-27
Fix failing config version test

JarrettSpiker:
Fixes [this test](https://github.com/hashicorp/go-tfe/blob/main/configuration_version_integration_test.go#L253) which has started to fail because of [this atlas change](https://github.com/hashicorp/atlas/pull/13288/files#diff-42de4f4e06a79c933dc219d1e0d683c042a9314373f3b730e5a0bf2e1cc9def1R309)


Uk1288 (approved):
🎉 thanks for fixing this

PR #571, opened 2022-10-31
Validate the module version as version instead of ID

annawinkler:
Without these changes a module version with prerelease or metadata will fail, while these are actually valid SemVer versions that are allowed by the module registry.

The PR includes tests that fail without the additional code changes, so everything that is needed should be in here...


sebasslash (approved):
:rocket: Looks good to me

PR #574, opened 2022-10-31
Update workflow names to be more meaningful in the GH UI

annawinkler:
## Description

Update workflow names to be more meaningful in the github UI list.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,4 +1,4 @@
-name: golangci-lint
+name: Linter
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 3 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -1,4 +1,4 @@
-name: golangci-lint
+name: Linting with golangci-lint
```
sebasslash, line 1: Could we leave this as `Linter` as we seem to be using nouns for other workflow names, i.e `CI Tests` and `TFLocal Instance`? Heck we can even rename the file to `lint.yml` as well.

sebasslash: Ah, that's a fair point 😃. Looking now, I see your pending changes in #577 would alter this workflow to run gofmt as well, making the filename partially inaccurate. What do you think?

sebasslash (approved):
:shipit: :rocket:

PR #577, opened 2022-11-01
Add check for go fmt

annawinkler:
## Description

Add `go fmt` check before tests are run.

## Testing plan

1. A build should fail if `go fmt` has a file that needs `go fmt` to be run on.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,17 @@ jobs:
           go mod download
           go mod tidy
 
+      # Check go fmt output because it does not report non-zero when there are fmt changes
+      - name: Run gofmt
+        run: |
+          go fmt ./...
+          files=$(go fmt ./...)
+            if [ -n "$files" ]; then
+              echo "The following file(s) do not conform to go fmt:"
+              echo "$files"
+              exit 1
+            fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -65,6 +65,17 @@ jobs:
           go mod download
           go mod tidy
 
+      # Check go fmt output because it does not report non-zero when there are fmt changes
+      - name: Run gofmt
+        run: |
+          go fmt ./...
```
brandonc, line 71:
```suggestion
```

Extra line?

brandonc (changes requested):
Since we run this `tests` job on a matrix of two nodes, without failfast, it may be better to run this as part of a broader linting workflow that does this plus golangci-lint

That way, we don't run go fmt on every node and potentially get earlier feedback about the fmt problem.

annawinkler:
> Since we run this `tests` job on a matrix of two nodes, without failfast, it may be better to run this as part of a broader linting workflow that does this plus golangci-lint (see `.github/workflows/golangci-lint.yml`)
> 
> That way, we don't run go fmt on every node and potentially get earlier feedback about the fmt problem.

Good point. I'll move the test to be part of linting.

PR #580, opened 2022-11-04
Split tests within action

brandonc:
I never liked putting the test selection code in every test and felt there was a better way to set up test splitting. This is my attempt. See https://github.com/brandonc/go-test-split-action


sebasslash (approved):
This is an awesome change and saves a lot of headache for future contributors -- I've already encountered friction with having to remind authors to add `skipIfNotCINode()`. I'll approve but we should probably hold off merging until another member approves?

PR #581, opened 2022-11-07
Since we only have one test suite, remove integration build flag

brandonc:
There are quite a few tests that are skipped because the author forgot to add this flag to new test files and I'm not sure what this is used for.


sebasslash (approved):
:broom: :dash: :+1:

annawinkler:
🥳 Thank you for updating the tests!

PR #596, opened 2022-11-30
fix make lint

estenssoros:
## Description

I was attempting to add a new feature and ran into a significant amount of errors running make lint. I attempted to fix them here, but wasn't sure how you wanted to structure the constants and/or if I should split this into multiple PRs. There are some changes to non-test files which might introduce breaking changes so maybe this is just a bad idea overall. Let me know what you think.

## Output from tests

```
TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./... -v -run TestFunctionsAffectedByChange
testing: warning: no tests to run
PASS
ok  	github.com/hashicorp/go-tfe	0.002s [no tests to run]
?   	github.com/hashicorp/go-tfe/mocks	[no test files]

...
```

### DIFF
--- a/admin_run_integration_test.go
+++ b/admin_run_integration_test.go
@@ -13,6 +13,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testWhenRunDoesNotExist = "when the run does not exist"
+const testQueuedAtTime = "2020-03-16T23:15:59+00:00"
+const testWithoutListOptions = "without list options"
+
 func TestAdminRuns_List(t *testing.T) {
 	skipIfCloud(t)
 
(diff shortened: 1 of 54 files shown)


brandonc:
@estenssoros Hello! Thanks so much for drawing my attention to this. Over time, the linter errors can become cluttered because our CI action only checks changed code and it is also not a required check for merge.

Your changes are really appreciated, but I found several problems that I didn't want to make you correct, so I created a new commit and added you as co-author https://github.com/hashicorp/go-tfe/pull/597

Here's what I found:
- two of our tests were incorrectly build-tagged so they weren't being considered as callers for the unparam linter. These tests were being skipped by our CI! I had no idea until I started working through the linter issues.
- We intended not to use the goconst linter for test files, so after excluding those files a lot of changes became unnecessary. I think literals as test descriptions are far nicer than variables.
- Unfortunately, we can't change "Ascii" to "ASCII" without breaking the API and having to release as 2.0 😬 

Thanks again! I'll get this merged when it passes CI and you can rebase your feature onto main after.

PR #598, opened 2022-11-30
Chore: skipIfBeta is worded as the opposite logic

brandonc:
Also prefer "skipUnless" to "skipIfNot"

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -2167,7 +2167,13 @@ func skipIfFreeOnly(t *testing.T) {
 	}
 }
 
-// skips a test if the test requires a beta feature
+// skips a test if the underlying beta feature is not available.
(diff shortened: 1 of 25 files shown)


helper_test.go
```diff
@@ -2168,14 +2168,14 @@ func skipIfFreeOnly(t *testing.T) {
 }
 
 // skips a test if the test requires a beta feature
-func skipIfBeta(t *testing.T) {
+func skipUnlessBeta(t *testing.T) {
```
Uk1288, line 2171: The comment above should probably change to `// skips a test if beta feature is not enabled`

brandonc: I added a lot of detail in the comments about the usage of this skip

Uk1288 (approved):
🎉  Looks good to me.

Issue #471, opened 2022-07-18
[Feature Request] Add List() to retreieve GPG Keys


straubt1:
GPG key functions exist, but `List()` is missing.

https://github.com/hashicorp/go-tfe/blob/main/gpg_key.go

Docs: Docs: https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys

Potential interface change:

```
type GPGKeys interface {
	... existing

	// List all GPG keys.
	List(ctx context.Context, organizationName string) (*GPGKeyList, error)
}
```

annawinkler:
Thank you for submitting this issue! It is interesting that the API does not have list. It seems like a reasonable request though! I'll ask 🥳

straubt1:
I was able to shim this with this url format if that helps.

```
	url := fmt.Sprintf(
		"https://%s/api/registry/private/v2/gpg-keys?filter[namespace]=%s",
		c.Hostname,
		c.OrganizationName,
	)
```

annawinkler:
Addressed with merged PR.

Issue #470, opened 2022-07-18
[Feature Request] Add "provider-binary-uploaded" to response


straubt1:
When dealing with Provider Platforms within the Registry, there is an attribute that is helpful to determine if a platform version has been created, but has not been uploaded with the binary.

Request to have `RegistryProviderPlatforms.List()` and `RegistryProviderPlatforms.Read()` functions return the "provider-binary-uploaded" attribute of a registry provider platform.

Public Docs: https://www.terraform.io/cloud-docs/api-docs/private-registry/provider-versions-platforms#get-all-platforms-for-a-single-version

Go struct: https://github.com/hashicorp/go-tfe/blob/main/registry_provider_platform.go#L35

Potential code change
```
ProviderBinaryUploaded bool   `json:"provider-binary-uploaded"`
```

annawinkler:
Thank you for submitting this issue! That seems like a reasonable request. ❤️

annawinkler:
Addressed in merged PR.

PR #602, opened 2022-12-06
Add `List()` to GPG Keys, `ProviderBinaryUploaded` to registry platforms

sebasslash:
## Description

These are two feature requests submitted by @straubt1 (thank you!):
* Add support for listing GPG keys (#471)
* Add `provider-binary-upload` field to response (#470)

Effectively, this PR adds `List()` to the `GPGKeys` interface and `ProviderBinaryUploaded` field to `RegistryPlatform` struct. 

**Note:** There does seem to be an issue in our tests with requesting GPG keys for multiple namespaces. The docs specify a [comma separated list](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/gpg-keys#query-parameters) -- which go-tfe will [automagically encode](https://github.com/hashicorp/go-tfe/blob/6c8979908ba9431a84e85c3cffca7093c1a3a842/tfe.go#L632-L636) any `[]string` to a comma-separated string if the param starts with `filter[`. I have skipped the test for now. 

## Testing plan

```sh
go test -v ./... -run TestGPGKeysList
go test -v ./... -run TestRegistryProviderPlatformsList
go test -v ./... -run TestRegistryProviderPlatformsRead
```

### DIFF
--- a/gpg_key.go
+++ b/gpg_key.go
@@ -15,6 +15,9 @@ var _ GPGKeys = (*gpgKeys)(nil)
 //
 // TFE API Docs: https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys
 type GPGKeys interface {
+	// Lists GPG keys in a private registry.
+	ListPrivate(ctx context.Context, options GPGKeyListOptions) (*GPGKeyList, error)
+
 	// Uploads a GPG Key to a private registry scoped with a namespace.
 	Create(ctx context.Context, registryName RegistryName, options GPGKeyCreateOptions) (*GPGKey, error)
 
@@ -31,6 +34,12 @@ type GPGKeys interface {
 // gpgKeys implements GPGKeys
 type gpgKeys struct {
 	client *Client
+}
+
+// GPGKeyList represents a list of GPG keys.
+type GPGKeyList struct {
+	*Pagination
+	Items []*GPGKey
 }
 
 // GPGKey represents a signed GPG key for a TFC/E private provider.
@@ -53,6 +62,14 @@ type GPGKeyID struct {
 	KeyID        string
 }
 
+// GPGKeyListOptions represents all the available options to list keys in a registry.
+type GPGKeyListOptions struct {
+	ListOptions
+
+	// Required: A list of one or more namespaces. Must be authorized TFC/E organization names.
+	Namespaces []string `url:"filter[namespace]"`
+}
+
 // GPGKeyCreateOptions represents all the available options used to create a GPG key.
 type GPGKeyCreateOptions struct {
 	Type       string `jsonapi:"primary,gpg-keys"`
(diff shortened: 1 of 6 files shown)


uturunku1:
Thanks for adding the changes to the failing tests in this PR. This test TestRegistryProviderPlatformsList/with_platforms is failing because the it is iterating over the same versionID https://github.com/hashicorp/go-tfe/blob/22885be660182ab4099606842d1cca398cce8a4e/registry_provider_platform_integration_test.go#L303
which is fine because I think those two tests within `TestRegistryProviderPlatformsList/with_platforms` wants to reuse the same provider version. However, the issue is the create api request for the registry platform expects unique os and arch (for the same provider). So I think you may want to overwrite (over the loop on that test) the default values for os and arch that are set in the helper_test file.

gpg_key.go
```diff
@@ -15,6 +15,9 @@ var _ GPGKeys = (*gpgKeys)(nil)
 //
 // TFE API Docs: https://www.terraform.io/cloud-docs/api-docs/private-registry/gpg-keys
 type GPGKeys interface {
+	// Lists GPG keys in a private registry.
+	List(ctx context.Context, registryName RegistryName, options *GPGKeyListOptions) (*GPGKeyList, error)
```
brandonc, line 19: If filter[namespace] is required, we should not make this a pointer in order to reinforce that, I think

gpg_key.go
```diff
@@ -66,6 +83,30 @@ type GPGKeyUpdateOptions struct {
 	Namespace string `jsonapi:"attr,namespace"`
 }
 
+func (s *gpgKeys) List(ctx context.Context, registryName RegistryName, options *GPGKeyListOptions) (*GPGKeyList, error) {
+	if registryName != PrivateRegistry {
+		return nil, ErrInvalidRegistryName
+	}
```
brandonc, line 89: I don't find this very user friendly since it must always be "private". Another alternative is to omit the parameter and allow extension in the future by naming the method "ListPrivate". WDYT?

brandonc (commented):
Not necessarily blocking feedback, but wondering what you think

gpg_key.go
sebasslash: I was iffy on this and decided on consistency (our ListOpts are almost always pointers). But yeah, I can imagine this confusing someone so I'll change.

gpg_key.go
sebasslash: Cool with removing this param and renaming to ListPrivate (on the off chance we end up supporting listing keys from the public registry).

gpg_key.go
```diff
@@ -179,6 +217,20 @@ func (o GPGKeyID) valid() error {
 	return nil
 }
 
+func (o *GPGKeyListOptions) valid() error {
+	if len(o.Namespaces) == 0 {
+		return ErrInvalidNamespace
+	}
+
+	for _, namespace := range o.Namespaces {
+		if namespace == "" || !validString(&namespace) {
```
brandonc, line 226: validString already checks for nil or empty string

PR #603, opened 2022-12-07
use available combination for os and arch platform

uturunku1:
## Output from tests

```

...
```


uturunku1:
I added this small change to another PR. Closing this one.

PR #616, opened 2022-12-13
Add comments to state_version.go about fields that get populated async

nfagerlund:
## Description

This is a common pitfall when consuming stuff in state versions that needs to be pulled out of the actual raw state data, and even our own tests have run afoul of it! So let's document it nice and clearly.

## Testing plan / tests output

- Comments-only PR.

### DIFF
--- a/state_version.go
+++ b/state_version.go
@@ -38,7 +38,9 @@ type StateVersions interface {
 	// Download retrieves the actual stored state of a state version
 	Download(ctx context.Context, url string) ([]byte, error)
 
-	// ListOutputs retrieves all the outputs of a state version by its ID.
+	// ListOutputs retrieves all the outputs of a state version by its ID. IMPORTANT: Terraform Cloud might
+	// process outputs asynchronously. When consuming outputs or other async StateVersion fields, be sure to
+	// wait for ResourcesProcessed to become `true` before assuming they are empty.
 	ListOutputs(ctx context.Context, svID string, options *StateVersionOutputsListOptions) (*StateVersionOutputsList, error)
 }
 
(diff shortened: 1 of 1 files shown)


state_version.go
```diff
@@ -55,18 +57,25 @@ type StateVersionList struct {
...
-	Providers          *StateVersionProviders   `jsonapi:"attr,providers"`
-	Resources          []*StateVersionResources `jsonapi:"attr,resources"`
+	ID           string    `jsonapi:"primary,state-versions"`
+	CreatedAt    time.Time `jsonapi:"attr,created-at,iso8601"`
+	DownloadURL  string    `jsonapi:"attr,hosted-state-download-url"`
+	Serial       int64     `jsonapi:"attr,serial"`
+	VCSCommitSHA string    `jsonapi:"attr,vcs-commit-sha"`
+	VCSCommitURL string    `jsonapi:"attr,vcs-commit-url"`
+	// Whether Terraform Cloud has finished populating any StateVersion fields that required async processing.
+	// If `false`, some fields may appear empty even if they should actually contain data; see comments on
```
uturunku1, line 67: Am I understanding correctly that if `ResourcesProcessed`  is set to `false`, then all the fields listed beneath (StateVersion, TerraformVersion, Modules, etc) will not have values populated yet?

annawinkler: Yes, that's correct. I think a nuance though is that `ResourcesProcessed` could be false, and then when it's true there could still be no values for those fields (depending on the state).

nfagerlund: Unfortunately, _some of them might while others will not._ There are _up to two_ rounds of async processing that might happen on a stateversion, and the atlas API as of today doesn't have granular progress tracking on them -- `resources-processed` gets flipped after the _final_ round of async processing completes, so it's the best signal we have that everything's done.

uturunku1: Gotcha 👍

uturunku1 (approved):

Issue #604, opened 2022-12-07
Go-slug v0.10.x breaks symlinks


alexlokshin-czi:
Hi, 

We've previously opened https://github.com/hashicorp/go-tfe/issues/507, but it got closed with no resolution. With the release of go-slug v0.10.1 the issue re-surfaced again. If your tf folder structure contains symlinks and you're trying to publish to TFE, you'll experience the following error:

```
Rel: can't make /Users/username/full/path/terraform/modules relative to ../../modules
```

sebasslash:
Hi @alexlokshin-czi  :wave: 

Thanks for reporting this issue!

Was wondering if you could shed some insight on how you are creating these symlinks? It seems there might something wonky with your symlink path but I have not yet been able to reproduce this error. That is, I've been able to successfully pack directories containing symlinks. 

Are these links resolving to a path existing outside your configuration directory? This is currently disabled when `go-slug` packs your target config path and we can potentially extend the `ConfigurationVersion` interface to allow for this.

PR #623, opened 2022-12-23
Add `UploadTarGzip` methods to `RegistryModule` and `ConfigurationVersion` interface

sebasslash:
## Description

The goal here is to add a simple interface for users who wish to upload their configurations using something other than the default method (using go-slug with default settings). There are particular use cases that either require a custom go-slug object to pack a Terraform config directory or to use something else entirely. One known issue with go-slug revolves around double symlinks as reported here:
- #604 
- [(go-slug) Pack fails to handle multiple levels of symlinks #32](https://github.com/hashicorp/go-slug/issues/32)
- [(terraform) archive/tar: write too long error when files are symlinked at least twice](https://github.com/hashicorp/terraform/issues/31895)

I have been unable to reproduce this issue, but I'm wondering if go-slug's `AllowSymlinkTarget()` configuration option may be a solution here ([It was introduced v0.10](https://github.com/hashicorp/go-slug/pull/29)). 

I have added `UploadRaw()` to the `RegistryModules` and `ConfigurationVersions` interface. Anything that implements `io.Writer` can be passed to `UploadRaw()`. 
 
## Testing plan

```sh
go test -v ./... -run TestConfigurationVersionsUploadRaw
go test -v ./... -run TestRegistryModulesUploadRaw
```

Check out the examples in `example_test.go`

### DIFF
--- a/configuration_version.go
+++ b/configuration_version.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net/url"
 	"time"
 )
@@ -34,6 +35,9 @@ type ConfigurationVersions interface {
 	// the upload URL from a configuration version and the full path to the
 	// configuration files on disk.
 	Upload(ctx context.Context, url string, path string) error
+
+	// Upload a tar gzip archive to the specified configuration version upload URL.
+	UploadTarGzip(ctx context.Context, url string, archive io.Reader) error
 
 	// Archive a configuration version. This can only be done on configuration versions that
 	// were created with the API or CLI, are in an uploaded state, and have no runs in progress.
@@ -258,7 +262,17 @@ func (s *configurationVersions) Upload(ctx context.Context, uploadURL, path stri
 		return err
 	}
 
-	req, err := s.client.NewRequest("PUT", uploadURL, body)
+	return s.UploadTarGzip(ctx, uploadURL, body)
+}
+
+// UploadTarGzip is used to upload Terraform configuration files contained a tar gzip archive.
+// Any stream implementing io.Reader can be passed into this method. This method is also
+// particularly useful for tar streams created by non-default go-slug configurations.
+//
+// **Note**: This method does not validate the content being uploaded and is therefore the caller's
+// responsibility to ensure the raw content is a valid Terraform configuration.
+func (s *configurationVersions) UploadTarGzip(ctx context.Context, uploadURL string, archive io.Reader) error {
+	req, err := s.client.NewRequest("PUT", uploadURL, archive)
 	if err != nil {
 		return err
 	}
--- a/example_test.go
+++ b/example_test.go
@@ -8,9 +8,75 @@ import (
 	slug "github.com/hashicorp/go-slug"
 )
 
+func ExampleOrganizations() {
+	config := &Config{
+		Token:             "insert-your-token-here",
+		RetryServerErrors: true,
+	}
+
+	client, err := NewClient(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Create a context
+	ctx := context.Background()
+
+	// Create a new organization
+	options := OrganizationCreateOptions{
+		Name:  String("example"),
+		Email: String("info@example.com"),
+	}
+
+	org, err := client.Organizations.Create(ctx, options)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Delete an organization
+	err = client.Organizations.Delete(ctx, org.Name)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func ExampleWorkspaces() {
+	config := &Config{
+		Token:             "insert-your-token-here",
+		RetryServerErrors: true,
+	}
+
+	client, err := NewClient(config)
+	if err != nil {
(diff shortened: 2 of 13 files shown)


brandonc (dismissed):
I'm not sure an io.Writer interface makes sense here.

In your test, bytes.Buffer implements io.Writer, so this passes... but if I were exploring this interface I would not expect this param to be _read from_ if described by this interface.

an io.Reader might be a better choice, but I'm unsure how NewRequest would handle common types of io.Readers (like File, or TeeReader) Can you try it? (With those types of underlying readers)

We need to provide an examples/ for uploading a configuration version this way and help explain why this interface exists.

brandonc:
What do you think about the name UploadTarGZip or UploadTarGZipStream ? So that it's clear exactly the kind of encoding is expected.

configuration_version.go
```diff
@@ -35,6 +36,9 @@ type ConfigurationVersions interface {
 	// configuration files on disk.
 	Upload(ctx context.Context, url string, path string) error
 
+	// Upload a tar gzip archive to the specified configuration version upload URL.
+	UploadTarGZip(ctx context.Context, url string, w io.Reader) error
```
Uk1288, line 40: update `w` to `r` to align with io.Reader

helper_test.go
```diff
@@ -2025,6 +2027,59 @@ func createProject(t *testing.T, client *Client, org *Organization) (*Project, f
...
+	}
+	defer out.Close()
+
+	gw := gzip.NewWriter(out)
+	defer gw.Close()
+
+	tw := tar.NewWriter(gw)
+	defer tw.Close()
+
+	if len(files) == 0 {
```
Uk1288, line 2043: since the createTar... is terminated if files are empty, it might be best to move this statement to lines 2031 so that the `os` doesn't bother creating the above path only to discard it in empty cases.

helper_test.go
```diff
@@ -2025,6 +2027,59 @@ func createProject(t *testing.T, client *Client, org *Organization) (*Project, f
...
+		}
+
+		_, err = io.Copy(tw, file)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	t.Cleanup(func() {
+		err := os.Remove(outputPath)
```
Uk1288, line 2076: 🎉

helper_test.go
```diff
@@ -2025,6 +2027,60 @@ func createProject(t *testing.T, client *Client, org *Organization) (*Project, f
...
+
+	tw := tar.NewWriter(gw)
+	defer tw.Close()
+
+	for _, filename := range files {
+		file, err := os.Open(filename)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer file.Close()
```
nfagerlund, line 2052: Hmm, the linter didn't like this `defer`, and I was curious why. I guess it's because it doesn't close the files until the whole function returns, even though you're done with each file at the end of its loop iteration. 

The best suggestion I found was to wrap the body of the loop in an anonymous, immediately executed function (`func() { ... }()`), so each defer will get resolved at the end of its iteration.

nfagerlund (changes requested):
I ran it and it works fine! I think Ukeme and Brandon already caught the substantive issues, but I'd like to get the names in proper shape before it ships. Also a couple other comments/questions, most notably about how we want to lay out examples.

example_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"bytes"
+	"context"
+	"log"
+
+	slug "github.com/hashicorp/go-slug"
+)
+
+func ExampleConfigurationVersionsUploadRaw() {
```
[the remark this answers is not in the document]
sebasslash: Thanks for this feedback! The `ExampleFunc()` idiom is so that examples[ show up in the GoDocs](https://go.dev/blog/examples). What I'll do is have both examples in `examples/` (if the user prefers to view them through Github) as well as in `example_test.go` if the user prefers to look at the GoDocs. I will update our existing examples as well.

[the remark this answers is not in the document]
sebasslash: No I've decided the examples won't be run as part of our test suite for the time being. Examples can be optionally executed and will only do so if the example function contains a `// Output: <some value>` comment at the end of the body.

configuration_version.go
```diff
@@ -258,7 +262,17 @@ func (s *configurationVersions) Upload(ctx context.Context, uploadURL, path stri
...
 
-	req, err := s.client.NewRequest("PUT", uploadURL, body)
+	return s.UploadTarGzip(ctx, uploadURL, body)
+}
+
+// UploadTarGzip is used to upload Terraform configuration files contained a tar gzip archive.
+// Any stream implementing io.Reader can be passed into this method. This method is also
+// particularly useful for tar streams created by non-default go-slug configurations.
+//
+// **Note**: This method does not validate the content being uploaded an2:00pmd is therefore the caller's
```
nfagerlund, line 272:
```suggestion
// **Note**: This method does not validate the content being uploaded and is therefore the caller's
```

Some kind of weird cmd-V problem 😆

sebasslash: Woah! :rofl:

nfagerlund (approved):

nfagerlund (approved):

PR #625, opened 2023-01-03
Remove installation instructions from README

### LINKED ISSUE #624: go get no longer exist with latest go
README tell to run go get command to install but this is no longer possible as this error out with 
go get -u github.com/hashicorp/go-tfe
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

looks like go get command was deprecated and has been completely removed with go 1.19

Please update README with the right command to install 
I tried unsuccesfully 
go install github.com/hashicorp/go-tfe@latest
package github.com/hashicorp/go-tfe is not a main package 

Using go 1.19.4 (latest version available via brew) on macosx 12.6

brandonc:
## Description

go-tfe is not installed (that is, there is no executable) so I'm clarifying the README. This was brought up in issue #624 in the context of `go get` being deprecated for binary installation, but I think it's best to just clarify its usage in a full example.

Closes #624

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,25 +20,20 @@ the other, if applicable (rare).
 
 Almost always, minor version changes will indicate backwards-compatible features and enhancements. Occasionally, function signature changes that reflect a bug fix may appear as a minor version change. Patch version changes will be used for bug fixes, performance improvements, and otherwise unimpactful changes.
 
-## Installation
-
-Installation can be done with a normal `go get`:
-
-```
-go get -u github.com/hashicorp/go-tfe
-```
-
-## Usage
-
-```go
-import tfe "github.com/hashicorp/go-tfe"
-```
+## Example Usage
 
 Construct a new TFE client, then use the various endpoints on the client to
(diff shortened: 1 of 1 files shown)


annawinkler (approved):
🌟

README.md
```diff
@@ -20,25 +20,20 @@ the other, if applicable (rare).
...
-import tfe "github.com/hashicorp/go-tfe"
-```
+## Example Usage
 
 Construct a new TFE client, then use the various endpoints on the client to
 access different parts of the Terraform Enterprise API. For example, to list
 all organizations:
 
 ```go
+import (
```
annawinkler, line 30: I like this update! Makes sense.

Uk1288 (approved):

PR #626, opened 2023-01-03
Update test.md to show usage of both tfe_address and tfe_hostname vars

Uk1288:
## Description

This PR updates the test.md and readme to show the use of both tfe_address and tfe_hostname env variables.

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./... 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,39 @@ Almost always, minor version changes will indicate backwards-compatible features
 ## Example Usage
 
 Construct a new TFE client, then use the various endpoints on the client to
-access different parts of the Terraform Enterprise API. For example, to list
-all organizations:
+access different parts of the Terraform Enterprise API. The following example list
+all organizations.
+
+### Using the default config with env vars
+The default configuration makes use of the `TFE_ADDRESS` and `TFE_TOKEN` environment variables.
+
+1. `TFE_ADDRESS` - URL of a Terraform Cloud or Terraform Enterprise instance. Example: `https://tfe.local`
+1. `TFE_TOKEN` - An [API token](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/api-tokens) for the Terraform Cloud or Terraform Enterprise instance.
+
+**Note:** Alternatively, you can set `TFE_HOSTNAME` which serves as a fallback for `TFE_ADDRESS`. It will only be used if `TFE_ADDRESS` is not set and will resolve the host to an `https` scheme. Example: `tfe.local` => resolves to `https://tfe.local`
+
+Set up the environment variables and use the default config like so:
+
+```go
+import (
+  "context"
+  "log"
+
+  "github.com/hashicorp/go-tfe"
+)
+
+client, err := tfe.NewClient(tfe.DefaultConfig())
+if err != nil {
+	log.Fatal(err)
+}
+
+orgs, err := client.Organizations.List(context.Background(), nil)
+if err != nil {
+	log.Fatal(err)
+}
+```
+
+### Using custom config to provide configuration details to the API client
 
 ```go
 import (
@@ -35,6 +66,7 @@ import (
 )
 
 config := &tfe.Config{
+	Address: "https://tfe.local",
 	Token: "insert-your-token-here",
 }
 
--- a/README.md
+++ b/README.md
@@ -23,39 +23,10 @@ Almost always, minor version changes will indicate backwards-compatible features
 ## Example Usage
 
 Construct a new TFE client, then use the various endpoints on the client to
-access different parts of the Terraform Enterprise API. The following example list
+access different parts of the Terraform Enterprise API. The following example lists
 all organizations.
 
-### Using the default config with env vars
-The default configuration makes use of the `TFE_ADDRESS` and `TFE_TOKEN` environment variables.
-
-1. `TFE_ADDRESS` - URL of a Terraform Cloud or Terraform Enterprise instance. Example: `https://tfe.local`
-1. `TFE_TOKEN` - An [API token](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/api-tokens) for the Terraform Cloud or Terraform Enterprise instance.
-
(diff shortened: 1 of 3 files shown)


brandonc (commented):
I think the actual issue here is that these environment variables configure the defaults of the package library itself, not just in the context of tests https://github.com/hashicorp/go-tfe/blob/main/tfe.go#L82-L99

nfagerlund (changes requested):
The tests.md improvements are solid 👍🏼 but I think we ought to have a section in the README about using env vars to configure the provider. The existing material only demonstrates providing the config directly in your Go code, so the fact that we'll reach out to the end user's environment on your behalf could be surprising if you're not expecting it.

README.md
```diff
@@ -23,8 +23,39 @@ Almost always, minor version changes will indicate backwards-compatible features
...
+
+```go
+import (
+  "context"
+  "log"
+
+  "github.com/hashicorp/go-tfe"
+)
+
+client, err := tfe.NewClient(tfe.DefaultConfig())
```
brandonc, line 47: I don't know how to express this in the context of your change, but if you pass a nil config or even empty values for Address or Token, the package will also use the env vars.

README.md
```diff
@@ -23,8 +23,39 @@ Almost always, minor version changes will indicate backwards-compatible features
 ## Example Usage
 
 Construct a new TFE client, then use the various endpoints on the client to
-access different parts of the Terraform Enterprise API. For example, to list
-all organizations:
+access different parts of the Terraform Enterprise API. The following example list
```
brandonc, line 26:
```suggestion
access different parts of the Terraform Enterprise API. The following example lists
```

README.md
```diff
@@ -23,8 +23,39 @@ Almost always, minor version changes will indicate backwards-compatible features
...
+	log.Fatal(err)
+}
+
+orgs, err := client.Organizations.List(context.Background(), nil)
+if err != nil {
+	log.Fatal(err)
+}
+```
+
+### Using custom config to provide configuration details to the API client
```
brandonc, line 58: This is definitely the recommended approach. I think we should promote this example, then mention later that if no config values are provided, the ENV is used as a fallback.

brandonc (changes requested):

nfagerlund (approved):

README.md
```diff
@@ -49,6 +52,81 @@ if err != nil {
...
+client, err := tfe.NewClient(tfe.DefaultConfig())
+if err != nil {
+	log.Fatal(err)
+}
+
+orgs, err := client.Organizations.List(context.Background(), nil)
+if err != nil {
+	log.Fatal(err)
+}
+```
```
sebasslash, line 83: I like that the example is explicit about what configuration we're using, but it may also be worth mentioning that `NewClient()` [will call](https://github.com/hashicorp/go-tfe/blob/b4072d56973028fc00921161247d66b8bd56716b/tfe.go#L271) `DefaultConfig()` so if `nil` is passed it is effectively equivalent.

docs/TESTS.md
```diff
@@ -71,14 +71,19 @@ Typically, you'll want to run specific tests. The commands below use notificatio
...
-#### Without envchain:
+#### Without envchain (Using TFE_ADDRESS):
 ```sh
-$ TFE_TOKEN=xyz TFE_ADDRESS=xyz ENABLE_TFE=1 go test -run TestNotificationConfiguration -v ./...
+$ TFE_TOKEN=xyz TFE_ADDRESS=https://tfe.local ENABLE_TFE=1 go test -run TestNotificationConfiguration -v ./...
+```
+
+#### Without envchain (Using TFE_HOSTNAME):
+```sh
+$ TFE_TOKEN=xyz TFE_HOSTNAME=tfe.local ENABLE_TFE=1 go test -run TestNotificationConfiguration -v ./...
```
sebasslash, line 81: :+1:

brandonc (approved):

README.md
```diff
@@ -35,6 +37,7 @@ import (
 )
 
 config := &tfe.Config{
+	Address: "https://tfe.local",
 	Token: "insert-your-token-here",
```
sebasslash, line 41: I think it may be worth adding `RetryServerErrors: true` to this. In our tests we configure the client this way and it perhaps should be part of best practices when using go-tfe. Although, our default configuration will set `RetryServerErrors` to false, so maybe we can ignore it if we feel it's not necessary.

sebasslash (commented):
All in all looks great! Won't approve just yet barring the conflict with the changelog. Some non-blocking items for discussion below.

README.md
Uk1288: The second example, lines#86 `When no tfe.Config is provided` captures this scenario, I am not sure where to add this additional summary in a way that is not confusing. Do you have any suggestions of where we can add this additional context?

sebasslash: I'm not sure we have to provide an extra example for passing nil, we can simply add a code comment above line 75:
`// Passing nil to this method will also use the default configuration`.

Uk1288: done 🎉

README.md
```diff
@@ -35,7 +37,85 @@ import (
...
+client, err := tfe.NewClient()
+if err != nil {
+	log.Fatal(err)
+}
+
+orgs, err := client.Organizations.List(context.Background(), nil)
+if err != nil {
+	log.Fatal(err)
+}
+```
```
sebasslash, line 105: Let's remove this section

README.md
```diff
@@ -35,7 +37,85 @@ import (
...
+The default configuration makes use of the `TFE_ADDRESS` and `TFE_TOKEN` environment variables.
+
+1. `TFE_ADDRESS` - URL of a Terraform Cloud or Terraform Enterprise instance. Example: `https://tfe.local`
+1. `TFE_TOKEN` - An [API token](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/api-tokens) for the Terraform Cloud or Terraform Enterprise instance.
+
+**Note:** Alternatively, you can set `TFE_HOSTNAME` which serves as a fallback for `TFE_ADDRESS`. It will only be used if `TFE_ADDRESS` is not set and will resolve the host to an `https` scheme. Example: `tfe.local` => resolves to `https://tfe.local`
+
+The environment variables are used as a fallback to configure TFE client if the Address or Token values are not provided as in the cases below:
+
+#### When tfe.DefaultConfig is used
```
sebasslash, line 66:
```suggestion
#### Using the default configuration
```

brandonc (approved):

PR #630, opened 2023-01-07
Update tfe tests

uturunku1:
## Background

A bunch of tests from the nightly TFE tests workflow have been failing (at least since the gh action workflow got implemented): https://github.com/hashicorp/go-tfe/actions/runs/3809715235/jobs/6481306992

These tests are failing due to different issues. That only issue that requires changes to the codebase are being addressed here. The smtp integration test was just missing the required field to update smtp with no auth.

The other issues required changes that will happen outside of this codebase. To understand more about the nature of these issues and how they will be addressed, please refer to my JIRA ticket 3488.

### DIFF
--- a/admin_run_integration_test.go
+++ b/admin_run_integration_test.go
@@ -93,7 +93,7 @@ func TestAdminRuns_List(t *testing.T) {
 			Include: []AdminRunIncludeOpt{"workpsace"},
 		})
 
-		assert.Equal(t, err, ErrInvalidIncludeValue)
+		assert.Equal(t, ErrInvalidIncludeValue, err)
 	})
 
 	t.Run("with RunStatus.pending filter", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


admin_run_integration_test.go
```diff
@@ -93,7 +93,7 @@ func TestAdminRuns_List(t *testing.T) {
 			Include: []AdminRunIncludeOpt{"workpsace"},
 		})
 
-		assert.Equal(t, err, ErrInvalidIncludeValue)
+		assert.Equal(t, ErrInvalidIncludeValue, err)
```
uturunku1, line 96: fyi, this change is completely unrelated to the issues that we are seeing for the tests. It's just that when I was debugging for this test, I noticed that the arguments for `expected` and `actual` where in the opposite places.

sebasslash (approved):
Looks good to me 👍

PR #632, opened 2023-01-11
Does not need review yet: Update main.go

uturunku1:
These tests are occasionally running with an incorrect TFE_TOKEN env var: https://github.com/hashicorp/go-tfe/blob/main/.github/workflows/nightly-tfe-ci.yml#L108

As an example, you can see that this run https://github.com/hashicorp/go-tfe/actions/runs/3878869728/jobs/6615534375 is showing that some tests are failing with "unauthorized", which is a symptom of a token that is querying for an organization or other resource that is does not have privileges to. 

This token, is an admin token, that is coming from the state outputs https://github.com/hashicorp/go-tfe/blob/main/.github/scripts/fetch_outputs/main.go#L42, that should have privileges to all the resources that are being managed on the tflocal instance.

Why would the tests sometimes run with the right token and not other times?  The value is clearly not empty otherwise we would see another error, which means that the token must switching in the middle of the process.
This issue is not reproducible locally. Every time that I grab the token and the hostname from the state outputs and place them into my environment before running the tests, I get [the regular test results](https://github.com/hashicorp/go-tfe/actions/runs/3859440900/jobs/6579019439), not the `unauthorized` errors.

My goal, for next time that we run into "unauthorized", is to check against that the last digits of the token to either match them to the one we see in our workspace's state outputs or to another workspace.

## Output from tests

```

...
```

### DIFF
--- a/.github/scripts/fetch_outputs/main.go
+++ b/.github/scripts/fetch_outputs/main.go
@@ -47,6 +47,11 @@ func newRunnerConfiguration(ctx context.Context, outputs []*tfe.StateVersionOutp
 	if _, ok := config["TFE_TOKEN"]; !ok {
 		return nil, fmt.Errorf("tfe_token output variable is not set")
 	}
+	// temporary debugging lines to troubleshoot for an issue around tfe_token
+	tk := config["TFE_TOKEN"]
+	fmt.Println(len(tk))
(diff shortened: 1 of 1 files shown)


uturunku1:
TDP created a JIRA ticket (499) to address this issue 🎉

Issue #640, opened 2023-01-20
[Feature] AdminWorkspaceListOptions - Add Sort and Filter capabilities


cyberious:
The current package needs the capabilities of Filter and Sort that are documented in the API. 

The current workaround is to grab all Workspaces with the IncludeOptions to include CurrentRuns and other objects I am attempting to filter. It would be advantageous to have this in `go-tfe` as well, as it would drastically cut down the processing time for our end users.

laurenolivia:
Thank you for flagging @cyberious. Our team will look into this.

PR #641, opened 2023-01-24
add sort and filter to admin/workspaces#list-all-workspaces

laurenolivia:
## Description

The Admin Workspace API includes a sort and filter feature. There was a [request](https://github.com/hashicorp/go-tfe/issues/640) to expose this in go-tfe.

## Testing plan

1. Create more than one workspace
2. With admin credentials, make an API call to either sort or filter
3. Sort only accepts `name` or `current-run.created-at`
4. Filter only accepts a `run status`

## External links

[API Docs](https://developer.hashicorp.com/terraform/enterprise/api-docs/admin/workspaces#query-parameters)

## Output from tests

```

...
```

### DIFF
--- a/admin_workspace_integration_test.go
+++ b/admin_workspace_integration_test.go
@@ -40,8 +40,8 @@ func TestAdminWorkspaces_ListWithFilter(t *testing.T) {
 
 		require.NoError(t, err)
 		require.NotEmpty(t, wl.Items)
-		assert.GreaterOrEqual(t, len(wl.Items), 1)
 		assert.Equal(t, wl.Items[0].CurrentRun.Status, RunApplied)
+		assert.NotContains(t, wl.Items, wTest2)
 	})
 }
 
--- a/admin_workspace_integration_test.go
+++ b/admin_workspace_integration_test.go
@@ -14,6 +14,76 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestAdminWorkspaces_ListWithFilter(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	org, orgCleanup := createOrganization(t, client)
+	defer orgCleanup()
+
+	wTest1, wTest1Cleanup := createWorkspace(t, client, org)
+	defer wTest1Cleanup()
+
+	wTest2, wTest2Cleanup := createWorkspace(t, client, org)
+	defer wTest2Cleanup()
+
+	t.Run("when filtering workspaces on a current run status", func(t *testing.T) {
+		_, appliedCleanup := createRunApply(t, client, wTest1)
+		t.Cleanup(appliedCleanup)
+
+		_, unAppliedCleanup := createRunUnapplied(t, client, wTest2)
+		t.Cleanup(unAppliedCleanup)
+
+		wl, err := client.Admin.Workspaces.List(ctx, &AdminWorkspaceListOptions{
+			Filter: string(RunApplied), Include: []AdminWorkspaceIncludeOpt{AdminWorkspaceCurrentRun},
+		})
+
+		require.NoError(t, err)
+		require.NotEmpty(t, wl.Items)
+		assert.GreaterOrEqual(t, len(wl.Items), 1)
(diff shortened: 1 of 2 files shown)


admin_workspace_integration_test.go
```diff
@@ -14,6 +14,68 @@ import (
...
+		_, appliedCleanup := createRunApply(t, client, wTest1)
+		t.Cleanup(appliedCleanup)
+
+		wl, err := client.Admin.Workspaces.List(ctx, &AdminWorkspaceListOptions{
+			Filter: "applied", Include: []AdminWorkspaceIncludeOpt{AdminWorkspaceCurrentRun},
+		})
+
+		require.NoError(t, err)
+		require.NotEmpty(t, wl.Items)
+	})
```
laurenolivia, line 37: ❓Looking for weigh-in: since this API call with an admin token will get all workspaces across all orgs, slicing at `Items[0]` to check for a run status applied won't exactly work. And even though this workspace applies the run, the `CurrentRun` attr was `null` when I printed the output of wTest1 obj.

sebasslash: For this given test: `filtering workspaces on current run status`, if you filter on status `applied` each workspace in the returned list should have a current run with status `applied`.

laurenolivia: I went ahead and pushed an assertion on the run status, but some things I am thinking about:

- when this API call returns several workspaces across different orgs, then how do we ensure the correct workspace run status is accounted for?
- `Items` is an array of objs, so this tests passes because `Items[0]` is the only object with a `CurrentRun.Status`

sebasslash: > when this API call returns several workspaces across different orgs, then how do we ensure the correct workspace run status is accounted for?

You're already doing so by:
```go
for _, wkspace := range wl.Items {
  assert.Equal(t, wkspace.CurrentRun.Status, "the-status-you-filtered-on")
}
``` 
Furthermore, I would use the type `RunStatus` for the `Filter` param instead of `string`. That way you can perform validations and instead assert:
```go
assert.Equal(t, wkspace.CurrentRun.Status, RunApplied)
```


> Items is an array of objs, so this test passes because Items[0] is the only object with a CurrentRun.Status

Correct, you may have one workspace in the returned list because the tests have only created **one** applied run on **one** workspace so far (I wouldn't be confident that this was always the case however). If you want to feel more confident that it's working properly, create several applied runs.

admin_workspace_integration_test.go
```diff
@@ -97,6 +160,8 @@ func TestAdminWorkspaces_List(t *testing.T) {
 		assert.NotEmpty(t, wl.Items[0].Organization.Name)
 	})
 
+	// This sub-test should remain last because it creates a run that does not apply
```
brandonc, line 163: 👍

brandonc (dismissed):

sebasslash (dismissed):

brandonc (approved):

admin_workspace_integration_test.go
```diff
@@ -14,6 +14,76 @@ import (
...
+		_, unAppliedCleanup := createRunUnapplied(t, client, wTest2)
+		t.Cleanup(unAppliedCleanup)
+
+		wl, err := client.Admin.Workspaces.List(ctx, &AdminWorkspaceListOptions{
+			Filter: string(RunApplied), Include: []AdminWorkspaceIncludeOpt{AdminWorkspaceCurrentRun},
+		})
+
+		require.NoError(t, err)
+		require.NotEmpty(t, wl.Items)
+		assert.Equal(t, wl.Items[0].CurrentRun.Status, RunApplied)
```
nfagerlund, line 43: Oh, now that I think this through an extra step, this is fine and doesn't need to be changed to an `assert.Contains` with our run. I got all turned around! But because the filter is limiting runs to the RunApplied status, the first run definitely has to be in status RunApplied, even if it's not the run we created. 👍🏼

nfagerlund (approved):

admin_workspace_integration_test.go
laurenolivia: Ok!

PR #643, opened 2023-01-27
Add copywrite headers and GH Action

barrettclark:
### :hammer_and_wrench: Description

* HashiCorp public repos will begin being scanned for source code copyright headers starting Feb.1, 2023. We want to make sure this repo is in adherence to the automated checks.
* Add Github action which will run a check on the repo during pull requests. This is simply a validation and will not automatically correct any source.
* Bring repo up-to-date using the `copywrite` CLI.

### :link: External Links

* [GH Action](https://github.com/marketplace/actions/setup-copywrite)
* `copywrite` [CLI](https://github.com/hashicorp/copywrite#getting-started)


.github/workflows/copywrite.yml
```diff
@@ -0,0 +1,28 @@
...
+  push:
+    tags:
+      - v*
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  copywrite:
```
brandonc, line 13: Can you add this job to the ci.yml workflow instead? What defines a workflow most of all is when it is run, and in this case we want this to run in CI events

barrettclark: Done. I'll update the other repos as well.

brandonc (approved):

PR #662, opened 2023-03-08
TF-4020 Apply and Remove Varsets from Projects

rberecka:
## Description

Adds the ability to use the new beta varsets/varset_id/relationships/project endpoint, allowing for the ability to apply and remove varsets from projects. Currently the feature flag is only enabled for internal users.

You can also list all of the varsets associated with a project at projects/project_id/varsets.
These changes align with the API additions we have made for enabling scoping varsets to projects.

We are not adding the ability to update projects or create a varset with a project relationship because we can see that the similar methods for workspaces are deprecated in the provider. See the deprecation warnings[ here in the provider](https://github.com/hashicorp/terraform-provider-tfe/blob/55a36f13ed3e0b73b83b113d1fc952664234bdc3/tfe/resource_tfe_variable_set.go#LL197).

## Testing plan

* Run tests against instance with "project-varsets" enabled.
```
ENABLE_BETA=1 TFE_TOKEN=XXXX TFE_HOSTNAME=XXXXXX TESTARGS="-run TestVariableSetsApplyToAndRemoveFromProjects" make testacc
```

## External links

- [API documentation PR](https://github.com/hashicorp/terraform-docs-common/pull/279)

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -2001,6 +2001,35 @@ func applyVariableSetToWorkspace(t *testing.T, client *Client, vsID, wsID string
 	})
 }
 
+func applyVariableSetToProject(t *testing.T, client *Client, vsID, prjID string) {
+	t.Helper()
+	if vsID == "" {
+		t.Fatal("variable set ID must not be empty")
+	}
+
+	if prjID == "" {
+		t.Fatal("project ID must not be empty")
+	}
+
+	opts := VariableSetApplyToProjectsOptions{}
+	opts.Projects = append(opts.Projects, &Project{ID: prjID})
+
+	ctx := context.Background()
+	if err := client.VariableSets.ApplyToProjects(ctx, vsID, opts); err != nil {
+		t.Fatalf("Error applying variable set %s to project %s: %v", vsID, prjID, err)
+	}
+
+	t.Cleanup(func() {
+		removeOpts := VariableSetRemoveFromProjectsOptions{}
+		removeOpts.Projects = append(removeOpts.Projects, &Project{ID: prjID})
+		if err := client.VariableSets.RemoveFromProjects(ctx, vsID, removeOpts); err != nil {
+			t.Errorf("Error removing variable set from project! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"VariableSet ID: %s\nError: %s", vsID, err)
+		}
+	})
+}
+
 func createVariableSetVariable(t *testing.T, client *Client, vs *VariableSet, options VariableSetVariableCreateOptions) (*VariableSetVariable, func()) {
 	var vsCleanup func()
 
--- a/variable_set.go
+++ b/variable_set.go
@@ -23,6 +23,9 @@ type VariableSets interface {
 	// ListForWorkspace gets the associated variable sets for a workspace.
 	ListForWorkspace(ctx context.Context, workspaceID string, options *VariableSetListOptions) (*VariableSetList, error)
 
+	// ListForProject gets the associated variable sets for a project.
+	ListForProject(ctx context.Context, projectID string, options *VariableSetListOptions) (*VariableSetList, error)
+
 	// Create is used to create a new variable set.
 	Create(ctx context.Context, organization string, options *VariableSetCreateOptions) (*VariableSet, error)
 
@@ -40,6 +43,12 @@ type VariableSets interface {
 
 	// Remove variable set from workspaces in the supplied list.
 	RemoveFromWorkspaces(ctx context.Context, variableSetID string, options *VariableSetRemoveFromWorkspacesOptions) error
+
+	// Apply variable set to projects in the supplied list.
+	ApplyToProjects(ctx context.Context, variableSetID string, options VariableSetApplyToProjectsOptions) error
+
+	// Remove variable set from projects in the supplied list.
+	RemoveFromProjects(ctx context.Context, variableSetID string, options VariableSetRemoveFromProjectsOptions) error
 
 	// Update list of workspaces to which the variable set is applied to match the supplied list.
 	UpdateWorkspaces(ctx context.Context, variableSetID string, options *VariableSetUpdateWorkspacesOptions) (*VariableSet, error)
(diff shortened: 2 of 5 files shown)


variable_set_test.go
```diff
@@ -111,6 +111,59 @@ func TestVariableSetsListForWorkspace(t *testing.T) {
...
+		for _, varset := range vsl.Items {
+			assert.Contains(t, ids, varset.ID)
+		}
+	})
+
+	t.Run("with list options", func(t *testing.T) {
+		t.Skip("paging not supported yet in API")
+		// Request a page number which is out of range. The result should
+		// be successful, but return no results if the paging options are
+		// properly passed along.
```
jbonhag, line 147: I think we can remove this `t.Skip` since the backend supports pagination now.

```suggestion
```

variable_set.go
```diff
@@ -141,6 +152,18 @@ type VariableSetRemoveFromWorkspacesOptions struct {
 	Workspaces []*Workspace
 }
 
+// VariableSetApplyToProjectsOptions represents the options for applying variable sets to projects.
+type VariableSetApplyToProjectsOptions struct {
+	// The workspaces to apply the variable set to (additive).
```
jbonhag, line 157: Tiny nit:

```suggestion
	// The projects to apply the variable set to (additive).
```

variable_set.go
```diff
@@ -141,6 +152,18 @@ type VariableSetRemoveFromWorkspacesOptions struct {
...
+	Projects []*Project
+}
+
+// VariableSetRemoveFromProjectsOptions represents the options for removing variable sets from projects.
+type VariableSetRemoveFromProjectsOptions struct {
+	// The workspaces to remove the variable set from.
```
jbonhag, line 163:
```suggestion
	// The projects to remove the variable set from.
```

jbonhag (dismissed):
Looks great! The tiniest of nits, nothing blocking.

jbonhag (dismissed):

helper_test.go
```diff
@@ -2001,6 +2001,34 @@ func applyVariableSetToWorkspace(t *testing.T, client *Client, vsID, wsID string
 	})
 }
 
+func applyVariableSetToProject(t *testing.T, client *Client, vsID, prjID string) {
+	if vsID == "" {
```
brandonc, line 2005: 💅 
```suggestion
	t.Helper()   
	if vsID == "" {
```

variable_set.go
```diff
@@ -330,6 +379,47 @@ func (s *variableSets) RemoveFromWorkspaces(ctx context.Context, variableSetID s
 	return req.Do(ctx, nil)
 }
 
+// ************************
+// Apply variable set to projects in the supplied list.
+// Note: this method will return an error if the variable set has global = true.
```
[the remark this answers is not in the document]
brandonc: Since these are not generally available, go ahead and add a note (to be cleaned up later when the feature is GA)

`// **Note: This feature is still in BETA and subject to change.**`

variable_set_test.go
```diff
@@ -349,6 +398,109 @@ func TestVariableSetsApplyToAndRemoveFromWorkspaces(t *testing.T) {
...
 }
 
+func TestVariableSetsApplyToAndRemoveFromProjects(t *testing.T) {
+	// TO DO: Remove when TFE GA
+	skipUnlessBeta(t)
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
```
brandonc, line 408: 💅 There's a mix of `defer` and `t.Cleanup` in the new code. Both work, but I was wondering if you ran into an issue with t.Cleanup, because that's what we prefer these days.

variable_set.go
```diff
@@ -330,6 +379,47 @@ func (s *variableSets) RemoveFromWorkspaces(ctx context.Context, variableSetID s
...
+func (s *variableSets) ApplyToProjects(ctx context.Context, variableSetID string, options *VariableSetApplyToProjectsOptions) error {
+	if !validStringID(&variableSetID) {
+		return ErrInvalidVariableSetID
+	}
+	if err := options.valid(); err != nil {
+		return err
+	}
+
+	u := fmt.Sprintf("varsets/%s/relationships/projects", url.QueryEscape(variableSetID))
+	req, err := s.client.NewRequest("POST", u, options.Projects)
```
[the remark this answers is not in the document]
jbonhag: Ah, thanks for pointing out that example @brandonc. I think the projects endpoint is a little different than VariableSetVariable Create because it expects an array of `data` (like [its workspaces sibling](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/variable-sets#apply-variable-set-to-workspaces)):

```
POST varsets/:varset_external_id/relationships/projects

{
  "data": [
    {
      "type": "projects",
      "id": "prj-YwfuBJZkdai4xj9w"
    }
  ]
}
```

versus VariableSetVarset Create, which wants a [single object](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/variable-sets#add-variable):

```
POST varsets/:varset_external_id/relationships/vars

{
  "data": {
    "type": "vars",
    "attributes": {
      "key": "g6e45ae7564a17e81ef62fd1c7fa86138",
      "value": "61e400d5ccffb3782f215344481e6c82",
      "description": "cheeeese",
      "sensitive": false,
      "category": "terraform",
      "hcl": false
    }
  }
}
```

I couldn't find a jsonapi decorator to turn that options struct into an array of objects under a root-level `data` key... does it make sense to change that method signature to take a slice of Projects instead, something like this?

```go
func (s *variableSets) ApplyToProjects(ctx context.Context, variableSetID string, projects []Project)
...
	u := fmt.Sprintf("varsets/%s/relationships/projects", url.QueryEscape(variableSetID))
	req, err := s.client.NewRequest("POST", u, projects)
```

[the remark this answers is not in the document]
brandonc: Ah, I didn't even realize this was an array. With a body like that, there it not a lot of room for other meaningful arguments. Here's a summary of our current conventions:

- AdminOrganizations UpdateModuleConsumers: accepts consumerOrganizationIDs []string and wraps each in a special purpose jsonapi ID/type struct
- OrganizationTags AddWorkspaces: accepts options struct with ProjectIDs []string and re-wraps them in the same type of special purpose jsonapi ID/type struct
- PolicySet AddPolicies: accepts options struct with a full []Policy member -- I think this is similar to what you are doing in this change

Having all that context, I think what is implemented is the best choice! Sorry for my incompetence.

brandonc (approved):
💯

PR #666, opened 2023-03-15
Add support for rotate / trim admin OIDC endpoints

mpminardi:
## Description

Adds support for the rotate key and trim key OIDC admin endpoints.

## Output from tests

```
go test -run TestAdminSettings_Oidc -v
=== RUN   TestAdminSettings_Oidc_RotateKey
--- PASS: TestAdminSettings_Oidc_RotateKey (1.23s)
=== RUN   TestAdminSettings_Oidc_TrimKey
--- PASS: TestAdminSettings_Oidc_TrimKey (1.54s)
PASS
ok  	github.com/hashicorp/go-tfe	3.158s
```

### DIFF
--- a/admin_setting_oidc_integration_test.go
+++ b/admin_setting_oidc_integration_test.go
@@ -1,0 +1,134 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tfe
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type Jwks struct {
+	Keys []Key `json:"keys"`
+}
+
+type Key struct {
+	Kid string `json:"kid"`
+}
+
+func TestAdminSettings_Oidc_RotateKey(t *testing.T) {
+	skipUnlessEnterprise(t)
+
+	client := testClient(t)
+	jwksClient := http.Client{
+		Timeout: time.Second * 2,
+	}
+	baseURL := client.baseURL
+	token := client.token
+
+	ctx := context.Background()
+
+	jwks, err := getJwks(jwksClient, baseURL, token)
+	require.NoError(t, err)
+
+	// Don't assume there is only 1 key to start
+	originalNumKeys := len(jwks.Keys)
+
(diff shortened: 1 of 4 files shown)


brandonc:
Our CI infrastructure is down right now but we're working on it.

brandonc:
All good for now

admin_setting_oidc_integration_test.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type Jwks struct {
+	Keys []Key `json:"keys"`
+}
+
+type Key struct {
+	Kid string `json:"kid"`
+}
+
```
brandonc, line 25: Since this is the main tfe package, I would recommend renaming these to something a little more specific (and making them private for good measure)

```suggestion
type wellKnownJwks struct {
	Keys []struct {
		Kid string `json:"kid"`
	} `json:"keys"`
}

```

admin_setting_oidc_integration_test.go
```diff
@@ -0,0 +1,134 @@
...
+	req, err := http.NewRequest(http.MethodGet, jwksEndpoint.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Accept", "application/json")
+
+	res, getErr := client.Do(req)
+	if getErr != nil {
```
brandonc, line 119: Just FYI this doesn't guarantee that the response was not an error response. I don't know if one is possible from the server, however, so you're probably good. It would only matter for test debugging anyway.

brandonc (dismissed):
Just very minor naming feedback. Looks fine!

admin_setting_oidc_integration_test.go
mpminardi: Ah good catch. Will add a check of the status code as well here for good measure in case it does help with test debugging in the future, but this shouldn't return an error response in regular operations.

admin_setting_oidc_integration_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	baseURL := client.baseURL
+	token := client.token
+
+	ctx := context.Background()
+
+	jwks, err := getJwks(jwksClient, baseURL, token)
+	require.NoError(t, err)
+
+	// Don't assume there is only 1 key to start
```
nfagerlund, line 70: Yeah good eye, testing global state is always 😬 like this.

nfagerlund (approved):
Seems good!

PR #668, opened 2023-03-23
hs26gill/TF-5101-Add-name-filter-for-listing-projects

hs26gill:
## Description

This PR adds a name option for listing projects. This will allow the user with the ability to filter projects based on a given name. This also updates the `ProjectListOptions` as optional which is the expected behaviour.

## Testing plan

Unit Tests have been added/updated. To run them, please 

```
go test -run TestTeamsUpdateManageProjects  -v ./... -tags=integration
```

### DIFF
--- a/project.go
+++ b/project.go
@@ -58,6 +58,8 @@ type ProjectListOptions struct {
 	ListOptions
 
 	// Optional: String (partial project name) used to filter the results.
+	// If multiple, comma separated values are specified, projects matching
+	// any of the names are returned.
 	Name string `url:"filter[names],omitempty"`
 }
 
--- a/project.go
+++ b/project.go
@@ -56,6 +56,9 @@ type Project struct {
 // ProjectListOptions represents the options for listing projects
 type ProjectListOptions struct {
 	ListOptions
+
+	// Optional: String (partial project name) used to filter the results.
+	Name string `url:"filter[names],omitempty"`
 }
 
 // ProjectCreateOptions represents the options for creating a project
@@ -88,10 +91,6 @@ func (s *projects) List(ctx context.Context, organization string, options *Proje
 		return nil, ErrInvalidOrg
 	}
 
-	if err := options.valid(); err != nil {
-		return nil, err
-	}
-
 	u := fmt.Sprintf("organizations/%s/projects", url.QueryEscape(organization))
 	req, err := s.client.NewRequest("GET", u, options)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


project.go
```diff
@@ -56,6 +56,9 @@ type Project struct {
 // ProjectListOptions represents the options for listing projects
 type ProjectListOptions struct {
 	ListOptions
+
+	// Optional: String (partial project name) used to filter the results.
```
brandonc, line 60: Can you mention that multiple names can be matched using comma-separated values?
```suggestion
	// Optional: String (partial project name) used to filter the results. If multiple, comma separated values are specified, projects matching any of the names are returned.
```

project.go
```diff
@@ -107,14 +106,6 @@ func (s *projects) List(ctx context.Context, organization string, options *Proje
...
 }
 
-func (o *ProjectListOptions) valid() error {
-	if o == nil || o.PageNumber == 0 || o.PageSize == 0 {
-		return ErrInvalidPagination
-	}
-
-	return nil
-}
-
```
brandonc, line 117: Curious: Why remove the validation?

project.go
hs26gill: Done!

project.go
hs26gill: The ProjectListOptions should be optional and not required. This validation checks that it is always present so removed it.

brandonc: Ah! I think I understand. In that case you can remove the nil check here. A "nil ProjectListOptions" is valid.

brandonc:
```go
func (o *ProjectListOptions) valid() error {
	if o != nil && (o.PageNumber == 0 || o.PageSize == 0) {
		return ErrInvalidPagination
	}

	return nil
}
```

brandonc (changes requested):

project.go
hs26gill: That makes sense. Thanks @brandonc!

brandonc (approved):

project.go
```diff
@@ -201,6 +197,14 @@ func (s *projects) Delete(ctx context.Context, projectID string) error {
 	return req.Do(ctx, nil)
 }
 
+func (o *ProjectListOptions) valid() error {
+	if o != nil && (o.PageNumber == 0 || o.PageSize == 0) {
```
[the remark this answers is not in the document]
hs26gill: Removed, thanks for the clarification @jbonhag

nfagerlund (approved):
Looks like everyone's comments were addressed, and it seems good to me. 👍🏼

PR #671, opened 2023-03-30
Add filtering option to list workspaces of a particular project

hs26gill:
## Description

This add the support for users to list all the workspaces associated to a particular project in an organization by providing project_id filter option. API docs: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#list-workspaces

## Testing plan
Integration tests have been added to cover this. You can run them using the below command:

```
go test -run TestWorkspacesList  -v ./... -tags=integration
```

## External links
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#list-workspaces)
- [Related PR](https://github.com/hashicorp/terraform-provider-tfe/pull/829)

### DIFF
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -217,6 +217,26 @@ func TestWorkspacesList(t *testing.T) {
 		assert.NotEmpty(t, wTest.ID)
 		assert.Equal(t, 0, wl.TotalCount)
 	})
+
+	t.Run("when filtering using project id", func(t *testing.T) {
+		p, pTestCleanup := createProject(t, client, orgTest)
+		defer pTestCleanup()
+
+		// create a workspace with project
+		w, wTestCleanup := createWorkspaceWithOptions(t, client, orgTest, WorkspaceCreateOptions{
+			Name:    String(randomString(t)),
+			Project: p,
+		})
+		defer wTestCleanup()
+		// Request a page number which is out of range. The result should
+		// be successful, but return no results if the paging options are
+		// properly passed along.
+		wl, err := client.Workspaces.List(ctx, orgTest.Name, &WorkspaceListOptions{
+			ProjectID: p.ID,
+		})
+		require.NoError(t, err)
+		assert.Contains(t, wl.Items, w)
+	})
 }
 
 func TestWorkspacesCreateTableDriven(t *testing.T) {
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -218,24 +218,36 @@ func TestWorkspacesList(t *testing.T) {
 		assert.Equal(t, 0, wl.TotalCount)
 	})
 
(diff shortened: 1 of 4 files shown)


workspace_integration_test.go
```diff
@@ -217,6 +217,26 @@ func TestWorkspacesList(t *testing.T) {
...
+
+		// create a workspace with project
+		w, wTestCleanup := createWorkspaceWithOptions(t, client, orgTest, WorkspaceCreateOptions{
+			Name:    String(randomString(t)),
+			Project: p,
+		})
+		defer wTestCleanup()
+		// Request a page number which is out of range. The result should
+		// be successful, but return no results if the paging options are
+		// properly passed along.
```
nfagerlund, line 233: This comment seems to be a copy-paste error!

workspace_integration_test.go
```diff
@@ -217,6 +217,26 @@ func TestWorkspacesList(t *testing.T) {
...
+		wl, err := client.Workspaces.List(ctx, orgTest.Name, &WorkspaceListOptions{
+			ProjectID: p.ID,
+		})
+		require.NoError(t, err)
+		assert.Contains(t, wl.Items, w)
```
nfagerlund, line 238: Hmm, to properly test the filter, I think we also need a test for _exclusion_ from the list. 

I'd be satisfied with either: 
- Create a workspace that lands in the default project, and confirm that it's excluded from the filtered list.
- Create another project and do a second search, confirming that our workspace is excluded when searching that other project.

I think we'd only need one of those, so whichever's comfier.

hs26gill: Added one where workspaces are added to the default project automatically, and the new project is empty.

workspace_integration_test.go
hs26gill: Added an appropriate comment

workspace_integration_test.go
```diff
@@ -217,6 +217,38 @@ func TestWorkspacesList(t *testing.T) {
...
+	})
+
+	t.Run("when using project id filter but project contains no workspaces", func(t *testing.T) {
+		// create a project in the orgTest
+		p, pTestCleanup := createProject(t, client, orgTest)
+		defer pTestCleanup()
```
nfagerlund, line 243: As written, that final assertion would also be true if the org just happened to have zero workspaces at the moment that we listed them. Let's make sure the test sets all the conditions for verifying:

```suggestion
		defer pTestCleanup()
		// create a workspace that's not in that project
		w, wTestCleaunp := createWorkspaceWithOptions(t, client, orgTest, WorkspaceCreateOptions{
			Name:    String(randomString(t)),
		})
		defer wTestCleanup()
```

nfagerlund (dismissed):
Thank you, sorry I got confused about the test setup 😅

barrettclark (approved):

PR #675, opened 2023-04-06
Add `IsCloud()` and `IsEnterprise()` helper methods to client

sebasslash:
## Description

These are helper methods that will enable external tools using a go-tfe client to make TFC or TFE-specific decisions. Thus far we haven't had an authoritative, simple way to determine if an instance is TFC or TFE. TFC and future TFE releases will start including a header `TFP-AppName` which will let us know if the given instance is TFC or TFE. For most users however, they will know beforehand what kind of instance their client is interacting with.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,14 @@
 # Unreleased
 
-FEATURES:
-* Add beta endpoints `ApplyToProjects`  and `RemoveFromProjects` to `VariableSets`.  Applying a variable set to a project will apply that variable set to all current and future workspaces in that project. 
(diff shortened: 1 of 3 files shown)


nfagerlund (approved):
I don't have a TFE instance around to trivially test this against, but it works against cloud right now, and the code looks fine. (Though, I can't seem to find where in the TFC code we actually add the header. 😅)

sebasslash:
This header won't be available until the next TFE release and I don't have a way of spinning up a TFE box, I'll just wait for nightly CI. You can find the header in the api/v2 base controller, search for app_name. This change was mostly for determining if an instance was TFC, since not every TFE release includes the TFE version header or the AppName header.

brandonc (approved):

PR #678, opened 2023-04-10
fix: updates VariableSetReadOptions tag to allow including `vars` and…

sgap:
… `workspaces` in the VariableSets.Read response

## Description

This change fixes an issue where VariableSets.Read does not honor the VariableSetReadOptions.Include values due to a syntax error in the struct tag of VariableSetReadOptions.

## Output from tests


brandonc (approved):
🙇 Thank you

brandonc:
This fix was released in [v1.21.0](https://github.com/hashicorp/go-tfe/releases/tag/v1.21.0)

PR #690, opened 2023-04-26
Use `make lint` instead of custom action

brandonc:
At some point, the custom action stopped leaving annotations on the PR and also doesn't offer log output in the action output.


Uk1288 (approved):
LGTM

PR #697, opened 2023-05-08
[WIP] Add support for a new variable field 'version-id'

arybolovlev:
## Description

This PR adds support for a new workspace variable field `version-id` that was recently added to the API.

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspace-variables) _search for `variable-id`_

## Output from tests

```

...
```


brandonc (approved):
😎

PR #699, opened 2023-05-09
Set tokens for admin roles in the environment

annawinkler:
## Description

We output admin tokens by role, so let's take that and set an environment variable to use in future tests.

## Testing plan

1. Nothing should change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,13 @@ jobs:
           matrix_total: ${{ matrix.total }}
           address: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_address }}
           token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_token }}
+          admin_configuration_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.configuration }}
+          admin_provision_licenses_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.provision_licenses }}
+          admin_security_maintenance_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.security_maintenance }}
+          admin_site_admin_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.site_admin }}
+          admin_subscription_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.subscription }}
+          admin_support_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.support }}
+          admin_version_maintenance_token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.version_maintenance }}
           oauth-client-github-token: ${{ secrets.OAUTH_CLIENT_GITHUB_TOKEN }}
 
   tests-combine-summaries:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - name: terraform-cloud-outputs
         id: tflocal
-        uses: hashicorp-forge/terraform-cloud-action/outputs@v1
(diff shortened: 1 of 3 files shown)


sebasslash (dismissed):

.github/workflows/ci.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           matrix_total: ${{ matrix.total }}
           address: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_address }}
           token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_token }}
+          admin_token_by_role: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.value }}
```
brandonc, line 46: I see a couple of issues: What is the `value` key? I'm pretty sure this object only has keys [named after each role](https://github.com/hashicorp/terraform-aws-tflocal-cloud/blob/main/variables.tf#L5) Also, I'm not sure what the behavior is when you force an object into an action input and furthermore into an env, since these things expect strings only...

brandonc (changes requested):

.github/workflows/ci.yml
annawinkler: Yeah I wasn't sure either to be honest.

.github/workflows/ci.yml
```diff
@@ -43,6 +43,13 @@ jobs:
           matrix_total: ${{ matrix.total }}
           address: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_address }}
           token: ${{ fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_token }}
+          admin_configuration_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.configuration }}
+          admin_provision_licenses_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.provision_licenses }}
+          admin_security_maintenance_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.security_maintenance }}
+          admin_site_admin_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.site_admin }}
+          admin_subscription_token: ${{steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.subscription }}
+          admin_support_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.support }}
+          admin_version_maintenance_token: ${{ steps.tflocal.outputs.workspace-outputs-json.tfe_admin_token_by_role.version_maintenance }}
```
brandonc, line 52: The CI itself isn't giving you the output you need to verify this, but you can see in the action log on this PR that these values aren't present, even though they are present in the workspace outputs:

Screenshot 2023-05-11 at 2 44 12 PM

The problem, I think, is that `steps.tflocal.outputs.workspace-outputs-json` comes out as a string (all GHA I/O are strings) so you have to use expressions like `fromJSON(steps.tflocal.outputs.workspace-outputs-json).tfe_admin_token_by_role.configuration`

brandonc (changes requested):

.github/workflows/ci.yml
annawinkler: Ah - I was wondering why I wasn't seeing the variables in the outputs. OK. Another update coming.

brandonc:
Closing this for now because the branch exposes test infra secrets. (Previously revealed tokens have been revoked)

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     steps:
       - name: terraform-cloud-outputs
         id: tflocal
-        uses: hashicorp-forge/terraform-cloud-action/outputs@v1
+        uses: hashicorp-forge/terraform-cloud-action/outputs@4adbe7eea886138ac10a4c09e63c5c568aaa6672
```
brandonc, line 31: It worked!

Screenshot 2023-05-12 at 10 36 06 AM

Nit: can you pin all the terraform-cloud-action/ actions to 4adbe7eea886138ac10a4c09e63c5c568aaa6672 ?

brandonc (dismissed):

.github/workflows/ci.yml
annawinkler: I noticed that not all of the tokens were listed so I fixed the reference to use hyphens for real (in my mind I always translate - to _)

brandonc (approved):

PR #701, opened 2023-05-11
Add UpdateAllowedWorkspaces method to Agent Pool

hs26gill:
## Description

This pull request is a bug fix to the Agent Pool resource that would enable users to remove all allowed workspaces from an agent pool. AgentPool `Update` was previously not able to accomplish this, so `UpdateAllowedWorkspaces` method is exclusively added to allow user to make this request. To facilitate this change, the `Allowed-Workspaces` attribute has been relocated from the `AgentPoolUpdateOptions` struct to a new struct named `AgentPoolAllowedWorkspacesUpdateOptions`. 

This modification is necessary to provide users with the ability to clear all allowed workspaces and set it to an empty value. Alternatively, users can use the existing `Update` method to modify other attributes of the agent pool while retaining the current allowed workspaces.

## Testing plan

Integration tests have been added to cover this section

## External links

- [Terraform Provider PR](https://github.com/hashicorp/terraform-provider-tfe/pull/870)

## Output from tests
```
=== RUN   TestAgentPoolsUpdateAllowedWorkspaces
--- PASS: TestAgentPoolsUpdateAllowedWorkspaces (5.60s)
=== RUN   TestAgentPoolsUpdateAllowedWorkspaces/when_updating_allowed-workspaces
    --- PASS: TestAgentPoolsUpdateAllowedWorkspaces/when_updating_allowed-workspaces (2.03s)
=== RUN   TestAgentPoolsUpdateAllowedWorkspaces/when_removing_all_the_allowed-workspaces
    --- PASS: TestAgentPoolsUpdateAllowedWorkspaces/when_removing_all_the_allowed-workspaces (1.43s)
PASS

...
```


jbonhag (dismissed):
Looks good to me! I think we could combine the changelog entries into a single bugfix entry but this implementation feels good.

jbonhag (dismissed):
Thanks!!

agent_pool.go
```diff
@@ -189,13 +192,22 @@ type AgentPoolUpdateOptions struct {
...
 
 	// A new name to identify the agent pool.
-	Name *string `jsonapi:"attr,name"`
+	Name *string `jsonapi:"attr,name,omitempty"`
 
 	// True if the agent pool is organization scoped, false otherwise.
 	OrganizationScoped *bool `jsonapi:"attr,organization-scoped,omitempty"`
+}
+
+// AgentPoolUpdateAllowedWorkspacesOptions represents the options for updating allowed workspace on an agent pool
```
annawinkler, line 201: 💅 

```suggestion
// AgentPoolUpdateAllowedWorkspacesOptions represents the options for updating the allowed workspace on an agent pool
```

uturunku1 (changes requested):
Can you add a test for when `UpdateAllowedWorkspaces` is called with `AgentPoolAllowedWorkspacesUpdateOptions{}`? 🙏

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ to the original message using the error `ErrWorkspaceLockedCannotDelete` instead
 * Adds `ExpiredAt` field to `OrganizationToken`, `TeamToken`, and `UserToken`. This enhancement will be available in TFE release, v202305-1. @JuliannaTetreault [#672](https://github.com/hashicorp/go-tfe/pull/672)
 * Adds `ContextWithResponseHeaderHook` context for use with the ClientRequest Do method that allows callers to define a callback which receives raw http Response headers.  @apparentlymart [#689](https://github.com/hashicorp/go-tfe/pull/689)
 
+## Bug Fixes
```
Uk1288, line 26: This should go under the `Unreleased` section since this change has not been released yet. Add the section to the top of the file if it does not exist:
```suggestion
# UNRELEASED

## Bug Fixes
```

Uk1288 (approved):
LGTM

PR #709, opened 2023-05-23
Fix accidental error incompatibility for safe-delete + ws locked

brandonc:
In the last release, I unintentionally overwrote the specific error message returned when safe deleting a locked workspace.


annawinkler (dismissed):
Good catch - thank you.

annawinkler (approved):
:shipit:

PR #710, opened 2023-05-30
add new API attributes to support import and config generation

radditude:
## Description

Add new API attributes to the `Run`, `Plan`, and `Apply` structs to support import and config generation.

I'm not sure what the protocol is on tests when adding new API attributes these days; I'd prefer not to add tests for this logic just yet, since all this is still behind a feature flag in Terraform Cloud so tests will fail.

## External links

- [Terraform CLI PR depending on these new attributes](https://github.com/hashicorp/terraform/pull/33278)


brandonc (approved):
This is fine and can be released as soon as merged. If these platform features are behind a feature flag, we should notate in the CHANGELOG and release notes that these enhancements are not generally available and subject to change. Don't worry about adding to the CHANGELOG in this branch, we have to modify it before applying the release tag anyway, which I can do for you.

PR #718, opened 2023-06-09
Re-add allowed-workspaces field to AgentPoolUpdateOptions

joekarl:
## Description

https://github.com/hashicorp/go-tfe/pull/701 made a breaking change to the AgentPoolUpdateOptions struct. This fixes that and re-adds a related test.

## Testing plan

- Run integration tests


brandonc (approved):
Thank you!

PR #717, opened 2023-06-08
Upload pending StateVersion

brandonc:
## Description

Adds `pending` StateVersion support to the model as well as a new method Upload which creates and uploads and re-fetches a state version.

## Refactor Notes

This change alters the behavior of requests that are made directly to `IPRanges` endpoint and the object store for `ConfigurationVersions`, `RegistryModels`, `PolicySetVersions`, (and now `StateVersions`). Previously, we were reusing the normal request methods for these, which unnecessarily sent an Authorization header, and additionally attempted to use jsonapi decoding (with the exception of `IPRanges`, which seems to have fundamentally broken error handling)

I added a new Client method, `doForeignPUTRequest`, which does an object-store-style application/octet-stream request with optional JSON response handling. The "foreign" part of the name reflects the fact that we aren't making API requests that have standard API features like Authorization with tokens and rate limiting.

## External links

- [Related terraform PR](https://github.com/hashicorp/terraform/pull/33336)

## Output from tests
Including output from tests may require access to a TFE instance. Ignore this section if you have no environment to test against.

```
$ ENABLE_BETA=1 go test ./... -run TestStateVersions -v
=== RUN   TestStateVersionsList
=== RUN   TestStateVersionsList/without_StateVersionListOptions
=== RUN   TestStateVersionsList/without_list_options
=== RUN   TestStateVersionsList/with_list_options
=== RUN   TestStateVersionsList/without_an_organization
=== RUN   TestStateVersionsList/without_a_workspace
--- PASS: TestStateVersionsList (21.07s)
    --- PASS: TestStateVersionsList/without_StateVersionListOptions (0.00s)
    --- PASS: TestStateVersionsList/without_list_options (1.43s)
    --- PASS: TestStateVersionsList/with_list_options (1.51s)
    --- PASS: TestStateVersionsList/without_an_organization (0.00s)
    --- PASS: TestStateVersionsList/without_a_workspace (0.00s)
=== RUN   TestStateVersionsUpload
=== RUN   TestStateVersionsUpload/can_create_finalized_state_versions
=== RUN   TestStateVersionsUpload/cannot_provide_base64_state_parameter_when_uploading
=== RUN   TestStateVersionsUpload/RawState_parameter_is_required_when_uploading
--- PASS: TestStateVersionsUpload (16.26s)
    --- PASS: TestStateVersionsUpload/can_create_finalized_state_versions (6.47s)
[...]

PR #722, opened 2023-06-16
Update error handling in DoJSON

brandonc:
This was part of a new method in #717 -- when an error occurs, the resp.Body is nil so we need to wait until after checking `err != nil`


sebasslash (dismissed):

brandonc:
I'm going to add some tests.

PR #737, opened 2023-07-13
Update policy_set to support project relationship on CREATE

hs26gill:
## Description

This PR provide users with the ability to apply projects when creating a new policy set. This feature is currently behind a feature flag in production

## Testing plan

Integration test is added to complete the testing plan

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/policy-sets#create-a-policy-set)

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./... -v -run policy_set_integration_test.go

=== RUN   TestPolicySetsCreate
--- PASS: TestPolicySetsCreate (2.65s)
=== RUN   TestPolicySetsCreate/with_policies_and_projects_provided
    --- PASS: TestPolicySetsCreate/with_policies_and_projects_provided (1.63s)
PASS

```

### DIFF
--- a/policy_set.go
+++ b/policy_set.go
@@ -73,18 +73,20 @@ type PolicySetList struct {
 
 // PolicySet represents a Terraform Enterprise policy set.
 type PolicySet struct {
-	ID             string     `jsonapi:"primary,policy-sets"`
-	Name           string     `jsonapi:"attr,name"`
-	Description    string     `jsonapi:"attr,description"`
-	Kind           PolicyKind `jsonapi:"attr,kind"`
-	Overridable    *bool      `jsonapi:"attr,overridable"`
-	Global         bool       `jsonapi:"attr,global"`
-	PoliciesPath   string     `jsonapi:"attr,policies-path"`
-	PolicyCount    int        `jsonapi:"attr,policy-count"`
-	VCSRepo        *VCSRepo   `jsonapi:"attr,vcs-repo"`
-	WorkspaceCount int        `jsonapi:"attr,workspace-count"`
-	CreatedAt      time.Time  `jsonapi:"attr,created-at,iso8601"`
-	UpdatedAt      time.Time  `jsonapi:"attr,updated-at,iso8601"`
+	ID           string     `jsonapi:"primary,policy-sets"`
+	Name         string     `jsonapi:"attr,name"`
+	Description  string     `jsonapi:"attr,description"`
+	Kind         PolicyKind `jsonapi:"attr,kind"`
+	Overridable  *bool      `jsonapi:"attr,overridable"`
+	Global       bool       `jsonapi:"attr,global"`
+	PoliciesPath string     `jsonapi:"attr,policies-path"`
+	// **Note: This field is still in BETA and subject to change.**
+	PolicyCount    int       `jsonapi:"attr,policy-count"`
+	VCSRepo        *VCSRepo  `jsonapi:"attr,vcs-repo"`
+	WorkspaceCount int       `jsonapi:"attr,workspace-count"`
+	ProjectCount   int       `jsonapi:"attr,project-count"`
+	CreatedAt      time.Time `jsonapi:"attr,created-at,iso8601"`
+	UpdatedAt      time.Time `jsonapi:"attr,updated-at,iso8601"`
 
 	// Relations
 	// The organization to which the policy set belongs to.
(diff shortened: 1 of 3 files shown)


Netra2104 (approved):

policy_set_integration_test.go
```diff
@@ -154,6 +154,34 @@ func TestPolicySetsCreate(t *testing.T) {
...
+		}
+
+		ps, err := client.PolicySets.Create(ctx, orgTest.Name, options)
+		require.NoError(t, err)
+
+		assert.Equal(t, ps.Name, *options.Name)
+		assert.Equal(t, ps.PolicyCount, 1)
+		assert.Equal(t, ps.Policies[0].ID, pTest.ID)
+		assert.Equal(t, ps.WorkspaceCount, 1)
+		assert.Equal(t, ps.Workspaces[0].ID, wTest.ID)
```
Netra2104, line 180: The 2nd argument is the expected and the 3rd argument is the actual value.

policy_set_integration_test.go
```diff
@@ -154,6 +154,34 @@ func TestPolicySetsCreate(t *testing.T) {
...
+		skipUnlessBeta(t)
+		pTest, pTestCleanup := createPolicy(t, client, orgTest)
+		defer pTestCleanup()
+		wTest, wTestCleanup := createWorkspace(t, client, orgTest)
+		defer wTestCleanup()
+		prjTest, prjTestCleanup := createProject(t, client, orgTest)
+		defer prjTestCleanup()
+
+		options := PolicySetCreateOptions{
+			Name:       String("populated-policy-set"),
```
Netra2104, line 167: Doesn't affect the test but maybe we should rename this to something else since we are already using this name for the test above.

policy_set_integration_test.go
```diff
@@ -154,6 +154,34 @@ func TestPolicySetsCreate(t *testing.T) {
...
+		ps, err := client.PolicySets.Create(ctx, orgTest.Name, options)
+		require.NoError(t, err)
+
+		assert.Equal(t, ps.Name, *options.Name)
+		assert.Equal(t, ps.PolicyCount, 1)
+		assert.Equal(t, ps.Policies[0].ID, pTest.ID)
+		assert.Equal(t, ps.WorkspaceCount, 1)
+		assert.Equal(t, ps.Workspaces[0].ID, wTest.ID)
+		assert.Equal(t, ps.ProjectCount, 1)
+		assert.Equal(t, ps.Projects[0].ID, prjTest.ID)
```
Netra2104, line 182: Doesn't affect the test but, for .Equal the 2nd argument is expected and 3rd argument is the actual value.
eg. assert.Equal(t, 1, ps.ProjectCount)
But, now that I look at this, this is not followed in the entire file lol, so I don't know we want to change it now.

policy_set_integration_test.go
```diff
@@ -154,6 +154,34 @@ func TestPolicySetsCreate(t *testing.T) {
 		assert.Equal(t, ps.Workspaces[0].ID, wTest.ID)
 	})
 
+	t.Run("with policies and projects provided", func(t *testing.T) {
```
Netra2104, line 157: We are adding workspaces as well so maybe "with policies, workspaces, and projects provided"

Side note: Is there a reason why you are doing both createWorkspace and createProject in this test instead of just createProject?

policy_set.go
```diff
@@ -99,6 +100,9 @@ type PolicySet struct {
 	NewestVersion *PolicySetVersion `jsonapi:"relation,newest-version"`
 	// The most recent successful policy set version.
 	CurrentVersion *PolicySetVersion `jsonapi:"relation,current-version"`
+	// **Note: This field is still in BETA and subject to change.**
+	// The projects to which the policy set applies.
+	Projects []*Project `jsonapi:"relation,projects"`
```
Netra2104, line 105: 👍

Netra2104 (dismissed):
Just left a few minor comments and questions. Looks great overall 💯

policy_set_integration_test.go
hs26gill: I've kept it for consistency. It is important for both the values to match which would be asserted this way as well.

policy_set_integration_test.go
hs26gill: Done!

policy_set_integration_test.go
hs26gill: adding workspaces allows to test that both project and workspaces work together as well.

brandonc (approved):

PR #732, opened 2023-07-10
Fix `StateVersionResources` field types

derekmckinnon:
## Description

When calling the `StateVersions.ReadCurrent` function, I noticed that the return `StateVersion` struct's `Resources` slice was always empty, even though `ResourcesProcessed` was `true` and my TFE state version definitely has resources. I used `curl` against the same endpoint and received the resources as expected, which led me to believe there was an unmarshalling error going on.

After debugging for a while, I found that the underlying `jsonapi` dependency was silently discarding the `StateVersionResources` with an error message of: `"The struct field was not of a known number type"`. I went to the code and realized that the `StateVersionResources` `Count` and `Type` had their data types swapped.

Changing `Count` to `int` and `Type` to `string` resolved the issue. I was able to testing using my forked version and swapping that in to my app. I had a look at the existing test suites in this project and I hesitate to change them because none of them seem to check for `StateVersionResources` at this time.

## Testing plan

1. Use the SDK to call any of the `StateVersions.Read*` functions on a workspace that has resources
2. Output the values of `ResourcesProcessed` and `StateVersionResources`

You will see that `ResourcesProcessed == true` and `len(StateVersionResources) == 0` .

PR #739, opened 2023-07-14
PR-732/run-CI-tests

laurenolivia:
This was previously causing jsonapi to fail silently and discarded all the resources, resulting in an empty slice.

## Description

This PR is carrying over the commit(s) of PR #732 so CI can run against an internal token.

## Testing plan

Tests should pass.

## Output from tests

```

...
```


sebasslash (approved):
👍 Thanks for fixing this @derekmckinnon

PR #743, opened 2023-07-25
Set component for the Jira Issue Sync Workflow

barrettclark:


brandonc:
Would you mind testing it by creating a bug in this repo?

PR #745, opened 2023-07-25
TF-5568 add support for project custom permissions

rberecka:
## Description

Add ability to choose custom access for team projects. Once selected, users can then choose to set a variety of customizable permissions to different levels of access. For example, if I add a team project and set the `TeamProjectAccessAddOptions` to include `access: "custom"` I can then also configure different permissions that will apply to either the project itself (`project-access`) or all workspace in the project (`workspace-access`).

Authored with @jbonhag 
## Testing plan

Pull down locally
Run tests below and comment out the `skipUnlessBeta(t)` flag. Once we have roll out the feature to TFE and remove the feature flag, we'll remove these skips as well.

## Output from tests
If you run atlas locally you can run the tests with the `ENABLE_BETA=1` environment variable.

```
$ ENABLE_BETA=1 TFE_ADDRESS=http://localhost:21080/ TFE_TOKEN=<API TOKEN> go test -run TestTeamProjectAccessesAdd
$ ENABLE_BETA=1 TFE_ADDRESS=http://localhost:21080/ TFE_TOKEN=<API TOKEN> go test -run TestTeamProjectAccessesUpdate
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ## Enhancements
 * Added BETA support for including `projects` relationship and `projects-count` attribute to policy_set on create by @hs26gill [#737](https://github.com/hashicorp/go-tfe/pull/737)
 * Added BETA method `AddProjects` and `RemoveProjects` for attaching/detaching policy set to projects by @Netra2104 [#735](https://github.com/hashicorp/go-tfe/pull/735)
+* Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
+the project level to the project itself (`ProjectAccessOptions`) and all of the workspaces in a project (`WorkspaceAccessOptions`).[#745](https://github.com/hashicorp/go-tfe/pull/745)
 
 # v1.30.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Added BETA support for including `projects` relationship and `projects-count` attribute to policy_set on create by @hs26gill [#737](https://github.com/hashicorp/go-tfe/pull/737)
 * Added BETA method `AddProjects` and `RemoveProjects` for attaching/detaching policy set to projects by @Netra2104 [#735](https://github.com/hashicorp/go-tfe/pull/735)
 * Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
-the project level to the project itself (`ProjectAccessOptions`) and all of the workspaces in a project (`WorkspaceAccessOptions`).[#745](https://github.com/hashicorp/go-tfe/pull/745)
+the project level to the project itself (`TeamProjectAccessProjectPermissionsOptions`) and all of the workspaces in a project (`TeamProjectAccessWorkspacePermissionsOptions`) by @rberecka [#745](https://github.com/hashicorp/go-tfe/pull/745)
 
 # v1.30.0
 
--- a/team_project_access.go
+++ b/team_project_access.go
@@ -74,7 +74,7 @@ type TeamProjectAccessProjectPermissions struct {
 	ProjectTeamsPermission    ProjectTeamsPermissionType    `jsonapi:"attr,teams"`
 }
 
-// Workspacepermissions represents the team's permission on all workspaces in its project
+// WorkspacePermissions represents the team's permission on all workspaces in its project
 type TeamProjectAccessWorkspacePermissions struct {
 	WorkspaceRunsPermission          WorkspaceRunsPermissionType          `jsonapi:"attr,runs"`
 	WorkspaceSentinelMocksPermission WorkspaceSentinelMocksPermissionType `jsonapi:"attr,sentinel-mocks"`
--- a/team_project_access.go
+++ b/team_project_access.go
@@ -47,6 +47,7 @@ const (
 	TeamProjectAccessMaintain TeamProjectAccessType = "maintain"
 	TeamProjectAccessWrite    TeamProjectAccessType = "write"
 	TeamProjectAccessRead     TeamProjectAccessType = "read"
+	TeamProjectAccessCustom   TeamProjectAccessType = "custom"
 )
 
 // TeamProjectAccessList represents a list of team project accesses
@@ -57,12 +58,103 @@ type TeamProjectAccessList struct {
 
 // TeamProjectAccess represents a project access for a team
 type TeamProjectAccess struct {
-	ID     string                `jsonapi:"primary,team-projects"`
-	Access TeamProjectAccessType `jsonapi:"attr,access"`
+	ID              string                                 `jsonapi:"primary,team-projects"`
+	Access          TeamProjectAccessType                  `jsonapi:"attr,access"`
+	ProjectAccess   *TeamProjectAccessProjectPermissions   `jsonapi:"attr,project-access"`
+	WorkspaceAccess *TeamProjectAccessWorkspacePermissions `jsonapi:"attr,workspace-access"`
 
 	// Relations
 	Team    *Team    `jsonapi:"relation,team"`
 	Project *Project `jsonapi:"relation,project"`
+}
+
+// ProjectPermissions represents the team's permissions on its project
+type TeamProjectAccessProjectPermissions struct {
+	ProjectSettingsPermission ProjectSettingsPermissionType `jsonapi:"attr,settings"`
+	ProjectTeamsPermission    ProjectTeamsPermissionType    `jsonapi:"attr,teams"`
+}
+
+type TeamProjectAccessWorkspacePermissions struct {
+	WorkspaceRunsPermission          WorkspaceRunsPermissionType          `jsonapi:"attr,runs"`
+	WorkspaceSentinelMocksPermission WorkspaceSentinelMocksPermissionType `jsonapi:"attr,sentinel-mocks"`
+	WorkspaceStateVersionsPermission WorkspaceStateVersionsPermissionType `jsonapi:"attr,state-versions"`
+	WorkspaceVariablesPermission     WorkspaceVariablesPermissionType     `jsonapi:"attr,variables"`
+	WorkspaceCreatePermission        bool                                 `jsonapi:"attr,create"`
+	WorkspaceLockingPermission       bool                                 `jsonapi:"attr,locking"`
+	WorkspaceMovePermission          bool                                 `jsonapi:"attr,move"`
+	WorkspaceDeletePermission        bool                                 `jsonapi:"attr,delete"`
+	WorkspaceRunTasksPermission      bool                                 `jsonapi:"attr,run-tasks"`
+}
+
+// ProjectSettingsPermissionType represents the permissiontype to a project's settings
+type ProjectSettingsPermissionType string
+
+const (
+	ProjectSettingsPermissionRead   ProjectSettingsPermissionType = "read"
+	ProjectSettingsPermissionUpdate ProjectSettingsPermissionType = "update"
+	ProjectSettingsPermissionDelete ProjectSettingsPermissionType = "delete"
+)
+
+// ProjectTeamsPermissionType represents the permissiontype to a project's teams
+type ProjectTeamsPermissionType string
+
+const (
+	ProjectTeamsPermissionNone   ProjectTeamsPermissionType = "none"
+	ProjectTeamsPermissionRead   ProjectTeamsPermissionType = "read"
+	ProjectTeamsPermissionManage ProjectTeamsPermissionType = "manage"
+)
+
+// WorkspaceRunsPermissionType represents the permissiontype to project workspaces' runs
+type WorkspaceRunsPermissionType string
+
+const (
+	WorkspaceRunsPermissionRead  WorkspaceRunsPermissionType = "read"
+	WorkspaceRunsPermissionPlan  WorkspaceRunsPermissionType = "plan"
+	WorkspaceRunsPermissionApply WorkspaceRunsPermissionType = "apply"
+)
+
+// WorkspaceSentinelMocksPermissionType represents the permissiontype to project workspaces' sentinel-mocks
+type WorkspaceSentinelMocksPermissionType string
+
+const (
+	WorkspaceSentinelMocksPermissionNone WorkspaceSentinelMocksPermissionType = "none"
+	WorkspaceSentinelMocksPermissionRead WorkspaceSentinelMocksPermissionType = "read"
+)
+
+// WorkspaceStateVersionsPermissionType represents the permissiontype to project workspaces' state-versions
+type WorkspaceStateVersionsPermissionType string
+
+const (
+	WorkspaceStateVersionsPermissionNone        WorkspaceStateVersionsPermissionType = "none"
+	WorkspaceStateVersionsPermissionReadOutputs WorkspaceStateVersionsPermissionType = "read-outputs"
+	WorkspaceStateVersionsPermissionRead        WorkspaceStateVersionsPermissionType = "read"
+	WorkspaceStateVersionsPermissionWrite       WorkspaceStateVersionsPermissionType = "write"
+)
+
+// WorkspaceVariablesPermissionType represents the permissiontype to project workspaces' variables
+type WorkspaceVariablesPermissionType string
+
+const (
+	WorkspaceVariablesPermissionNone  WorkspaceVariablesPermissionType = "none"
+	WorkspaceVariablesPermissionRead  WorkspaceVariablesPermissionType = "read"
+	WorkspaceVariablesPermissionWrite WorkspaceVariablesPermissionType = "write"
+)
+
+type TeamProjectAccessProjectPermissionsOptions struct {
+	Settings *ProjectSettingsPermissionType `json:"settings,omitempty"`
+	Teams    *ProjectTeamsPermissionType    `json:"teams,omitempty"`
+}
+
+type TeamProjectAccessWorkspacePermissionsOptions struct {
+	Runs          *WorkspaceRunsPermissionType          `json:"runs,omitempty"`
+	SentinelMocks *WorkspaceSentinelMocksPermissionType `json:"sentinel-mocks,omitempty"`
+	StateVersions *WorkspaceStateVersionsPermissionType `json:"state-versions,omitempty"`
+	Variables     *WorkspaceVariablesPermissionType     `json:"variables,omitempty"`
+	Create        *bool                                 `json:"create,omitempty"`
+	Locking       *bool                                 `json:"locking,omitempty"`
+	Move          *bool                                 `json:"move,omitempty"`
+	Delete        *bool                                 `json:"delete,omitempty"`
+	RunTasks      *bool                                 `json:"run-tasks,omitempty"`
 }
 
 // TeamProjectAccessListOptions represents the options for listing team project accesses
@@ -80,6 +172,9 @@ type TeamProjectAccessAddOptions struct {
 	Type string `jsonapi:"primary,team-projects"`
 	// The type of access to grant.
 	Access TeamProjectAccessType `jsonapi:"attr,access"`
+	// The levels that project and workspace permissions grant
+	ProjectAccess   *TeamProjectAccessProjectPermissionsOptions   `jsonapi:"attr,project-access,omitempty"`
+	WorkspaceAccess *TeamProjectAccessWorkspacePermissionsOptions `jsonapi:"attr,workspace-access,omitempty"`
 
 	// The team to add to the project
 	Team *Team `jsonapi:"relation,team"`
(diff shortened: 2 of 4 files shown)


team_project_access.go
```diff
@@ -57,14 +58,105 @@ type TeamProjectAccessList struct {
 
 // TeamProjectAccess represents a project access for a team
 type TeamProjectAccess struct {
-	ID     string                `jsonapi:"primary,team-projects"`
-	Access TeamProjectAccessType `jsonapi:"attr,access"`
+	ID              string                `jsonapi:"primary,team-projects"`
+	Access          TeamProjectAccessType `jsonapi:"attr,access"`
+	ProjectAccess   *ProjectAccess2       `jsonapi:"attr,project-access"`
```
rberecka, line 63: ProjectAccess is already taken in type_helpers.go. Because it's not something that gets called directly in a client, and in order to not have to do breaking changes without the rest of the semver process, we opted to go with this.

team_project_access.go
```diff
@@ -57,14 +58,105 @@ type TeamProjectAccessList struct {
...
+// WorkspaceVariablesPermissionType represents the permissiontype to project workspaces' variables
+type WorkspaceVariablesPermissionType string
+
+const (
+	WorkspaceVariablesPermissionNone  WorkspaceVariablesPermissionType = "none"
+	WorkspaceVariablesPermissionRead  WorkspaceVariablesPermissionType = "read"
+	WorkspaceVariablesPermissionWrite WorkspaceVariablesPermissionType = "write"
+)
+
+type ProjectAccessOptions struct {
```
jbonhag, line 143: Whatever we settle on for the base `project-access`, let's make sure to sync up the naming here.

team_project_access_integration_test.go
```diff
@@ -164,6 +165,102 @@ func TestTeamProjectAccessesAdd(t *testing.T) {
 		}
 	})
 
+	t.Run("with valid options for all custom TeamProject permissions", func(t *testing.T) {
```
jbonhag, line 168: Nice exhaustive test!

jbonhag (commented):
Amazing @rberecka!

team_project_access.go
```diff
@@ -57,14 +58,105 @@ type TeamProjectAccessList struct {
...
+	Access          TeamProjectAccessType       `jsonapi:"attr,access"`
+	ProjectAccess   *ProjectAccessPermissions   `jsonapi:"attr,project-access"`
+	WorkspaceAccess *WorkspaceAccessPermissions `jsonapi:"attr,workspace-access"`
 
 	// Relations
 	Team    *Team    `jsonapi:"relation,team"`
 	Project *Project `jsonapi:"relation,project"`
 }
 
+// ProjectPermissions represents the team's permissions on its project
```
jbonhag, line 71:
```suggestion
// ProjectAccessPermissions represents the team's specific permissions related to the project itself
```

team_project_access.go
```diff
@@ -57,14 +58,105 @@ type TeamProjectAccessList struct {
...
+type ProjectAccessPermissions struct {
+	ProjectSettingsPermission ProjectSettingsPermissionType `jsonapi:"attr,settings"`
+	ProjectTeamsPermission    ProjectTeamsPermissionType    `jsonapi:"attr,teams"`
+}
+
+type WorkspaceAccessPermissions struct {
```
jbonhag, line 77:
```suggestion
// WorkspaceAccessPermissions represents the team's specific permissions related to workspaces in the project
type WorkspaceAccessPermissions struct {
```

team_project_access_integration_test.go
```diff
@@ -242,6 +353,99 @@ func TestTeamProjectAccessesUpdate(t *testing.T) {
...
+		tm2Test, tm2TestCleanup := createTeam(t, client, orgTest)
+		defer tm2TestCleanup()
+
+		TpaOptions := TeamProjectAccessAddOptions{
+			Access:  *ProjectAccess(TeamProjectAccessCustom),
+			Team:    tm2Test,
+			Project: pTest,
+		}
+
+		tpaCustomTest, err := client.TeamProjectAccess.Add(ctx, TpaOptions)
```
jbonhag, line 407: Slight capitalization nit:

```suggestion
		tpaOptions := TeamProjectAccessAddOptions{
			Access:  *ProjectAccess(TeamProjectAccessCustom),
			Team:    tm2Test,
			Project: pTest,
		}

		tpaCustomTest, err := client.TeamProjectAccess.Add(ctx, tpaOptions)
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ## Enhancements
 * Added BETA support for including `projects` relationship and `projects-count` attribute to policy_set on create by @hs26gill [#737](https://github.com/hashicorp/go-tfe/pull/737)
 * Added BETA method `AddProjects` and `RemoveProjects` for attaching/detaching policy set to projects by @Netra2104 [#735](https://github.com/hashicorp/go-tfe/pull/735)
+* Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
+the project level to the project itself (`ProjectAccessOptions`) and all of the workspaces in a project (`WorkspaceAccessOptions`).[#745](https://github.com/hashicorp/go-tfe/pull/745)
```
jbonhag, line 9:
```suggestion
the project level to the project itself (`ProjectAccessPermissionsOptions`) and all of the workspaces in a project (`WorkspaceAccessPermissionsOptions`).[#745](https://github.com/hashicorp/go-tfe/pull/745)
```

jbonhag (commented):
Just a couple of observations related to the name changes, but this looks really good.

team_project_access_integration_test.go
```diff
@@ -242,6 +353,99 @@ func TestTeamProjectAccessesUpdate(t *testing.T) {
...
+		assert.Equal(t, *options.WorkspaceAccess.Variables, tpa.WorkspaceAccess.WorkspaceVariablesPermission)
+		assert.Equal(t, false, tpa.WorkspaceAccess.WorkspaceCreatePermission)
+		assert.Equal(t, false, tpa.WorkspaceAccess.WorkspaceLockingPermission)
+		assert.Equal(t, false, tpa.WorkspaceAccess.WorkspaceMovePermission)
+		assert.Equal(t, true, tpa.WorkspaceAccess.WorkspaceDeletePermission)
+		assert.Equal(t, true, tpa.WorkspaceAccess.WorkspaceRunTasksPermission)
+	})
+
+	t.Run("with valid custom permissions attributes for some permissions", func(t *testing.T) {
+		// create tpaCustomTest to verify unupdated attributes stay the same for custom permissions
```
jbonhag, line 396:
```suggestion
		skipUnlessBeta(t)
		// create tpaCustomTest to verify unupdated attributes stay the same for custom permissions
```

jbonhag (dismissed):
🎉

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ## Enhancements
 * Added BETA support for including `projects` relationship and `projects-count` attribute to policy_set on create by @hs26gill [#737](https://github.com/hashicorp/go-tfe/pull/737)
 * Added BETA method `AddProjects` and `RemoveProjects` for attaching/detaching policy set to projects by @Netra2104 [#735](https://github.com/hashicorp/go-tfe/pull/735)
+* Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
+the project level to the project itself (`ProjectAccessPermissionsOptions`) and all of the workspaces in a project (`WorkspaceAccessPermissionsOptions`).[#745](https://github.com/hashicorp/go-tfe/pull/745)
```
laurenolivia, line 9: 💅 can you add your github handle before the PR link? @rberecka

rberecka, line 9:
```suggestion
the project level to the project itself (`ProjectAccessPermissionsOptions`) and all of the workspaces in a project (`WorkspaceAccessPermissionsOptions`) by @rberecka [#745](https://github.com/hashicorp/go-tfe/pull/745)
```

jbonhag (approved):

laurenolivia:
@rberecka Great work on this PR ✨ 

I noticed there is a [TeamProjectAccess](https://github.com/hashicorp/go-tfe/blob/main/team_project_access.go#L59) struct, and a [TeamProjectAccesses](https://github.com/hashicorp/go-tfe/blob/main/team_project_access.go#L20) interface. The naming is almost identical and I'm wondering if it's too late to propose something like a `TeamProjectAccessMethods` (interface) instead? Especially considering the commented line above the interface reads: 
```
// TeamProjectAccesses describes all the team project access related methods that the Terraform
// Enterprise API supports
```
Not to mention this is customer-facing. What do you think? How can I support?

sebasslash:
Unfortunately given go-tfe's v1.0 nature and this interface being exported, we can't rename `TeamProjectAccesses` to something slightly less confusing. Thankfully from a client perspective you call `client.TeamProjectAccess` where the naming makes more sense.

laurenolivia:
> Unfortunately given go-tfe's v1.0 nature and this interface being exported, we can't rename `TeamProjectAccesses` to something slightly less confusing. Thankfully from a client perspective you call `client.TeamProjectAccess` where the naming makes more sense.

This would be a breaking change for v1.0 but what are your thoughts for v2.0? Or is it not worth it at that point? @sebasslash

team_project_access.go
```diff
@@ -57,14 +58,106 @@ type TeamProjectAccessList struct {
...
+	WorkspaceRunsPermission          WorkspaceRunsPermissionType          `jsonapi:"attr,runs"`
+	WorkspaceSentinelMocksPermission WorkspaceSentinelMocksPermissionType `jsonapi:"attr,sentinel-mocks"`
+	WorkspaceStateVersionsPermission WorkspaceStateVersionsPermissionType `jsonapi:"attr,state-versions"`
+	WorkspaceVariablesPermission     WorkspaceVariablesPermissionType     `jsonapi:"attr,variables"`
+	WorkspaceCreatePermission        bool                                 `jsonapi:"attr,create"`
+	WorkspaceLockingPermission       bool                                 `jsonapi:"attr,locking"`
+	WorkspaceMovePermission          bool                                 `jsonapi:"attr,move"`
+	WorkspaceDeletePermission        bool                                 `jsonapi:"attr,delete"`
+	WorkspaceRunTasksPermission      bool                                 `jsonapi:"attr,run-tasks"`
+}
```
sebasslash, line 88: I think to find this struct under `team_project_access.go` would be slightly odd. Seems like something that should be defined under `workspace.go` or if only pertinent to team projects be prefixed with `TeamProjects` -- which would make the name absurdly long so to keep things. I prefer the former to circumvent a long name.

team_project_access.go
```diff
@@ -57,14 +58,106 @@ type TeamProjectAccessList struct {
...
+
+// ProjectSettingsPermissionType represents the permissiontype to a project's settings
+type ProjectSettingsPermissionType string
+
+const (
+	ProjectSettingsPermissionRead   ProjectSettingsPermissionType = "read"
+	ProjectSettingsPermissionUpdate ProjectSettingsPermissionType = "update"
+	ProjectSettingsPermissionDelete ProjectSettingsPermissionType = "delete"
+)
```
sebasslash, line 97: Ditto for the rest of the types defined here -- should be organized under their respective resources even if only used by the `TeamProjectAccess` struct fields.

team_project_access.go
```diff
@@ -57,14 +58,106 @@ type TeamProjectAccessList struct {
...
+type WorkspaceAccessPermissionsOptions struct {
+	Runs          *WorkspaceRunsPermissionType          `json:"runs,omitempty"`
+	SentinelMocks *WorkspaceSentinelMocksPermissionType `json:"sentinel-mocks,omitempty"`
+	StateVersions *WorkspaceStateVersionsPermissionType `json:"state-versions,omitempty"`
+	Variables     *WorkspaceVariablesPermissionType     `json:"variables,omitempty"`
+	Create        bool                                  `json:"create,omitempty"`
+	Locking       bool                                  `json:"locking,omitempty"`
+	Move          bool                                  `json:"move,omitempty"`
+	Delete        bool                                  `json:"delete,omitempty"`
+	RunTasks      bool                                  `json:"run-tasks,omitempty"`
```
sebasslash, line 158: If you intend to _not_ send these over the wire, the type should be `*bool`. `omitempty` is only applied on a zeroed value, of which `bool` does not have. 

I think however `omitempty` for booleans is meaningless if the default is not `nil`, so you can remove that here 👍 😃

laurenolivia:
> I think over time if we can identify multiple quirks with our naming conventions as well as other potentially breaking changes that would improve the DX, we can build a good case for a v2.0.
> 
> For the time being, v2.0 in go-tfe is in the _way_ distant future.

Fair! Thanks for weighing-in @sebasslash 
@rberecka nvm! 👍

team_project_access_integration_test.go
```diff
@@ -164,6 +165,102 @@ func TestTeamProjectAccessesAdd(t *testing.T) {
...
+			assert.Equal(t, *options.ProjectAccess.Teams, item.ProjectAccess.ProjectTeamsPermission)
+			assert.Equal(t, *options.WorkspaceAccess.Runs, item.WorkspaceAccess.WorkspaceRunsPermission)
+			assert.Equal(t, *options.WorkspaceAccess.SentinelMocks, item.WorkspaceAccess.WorkspaceSentinelMocksPermission)
+			assert.Equal(t, *options.WorkspaceAccess.StateVersions, item.WorkspaceAccess.WorkspaceStateVersionsPermission)
+			assert.Equal(t, *options.WorkspaceAccess.Variables, item.WorkspaceAccess.WorkspaceVariablesPermission)
+			assert.Equal(t, item.WorkspaceAccess.WorkspaceCreatePermission, true)
+			assert.Equal(t, item.WorkspaceAccess.WorkspaceLockingPermission, true)
+			assert.Equal(t, item.WorkspaceAccess.WorkspaceMovePermission, true)
+			assert.Equal(t, item.WorkspaceAccess.WorkspaceDeletePermission, false)
+			assert.Equal(t, item.WorkspaceAccess.WorkspaceRunTasksPermission, false)
```
sebasslash, line 221: It's important here that you `require` each struct to not be `nil` before attempting to access a struct's field. Otherwise the tests will panic if CI flakes. 

Example:
```go
// this will exit the test early if it fails
require.NotNil(t, options.ProjectAccess)
assert.Equal(t, *options.ProjectAccess.Settings, item.ProjectAccess.ProjectSettingsPermission)
assert.Equal(t, *options.ProjectAccess.Teams, item.ProjectAccess.ProjectTeamsPermission)
...
require.NotNil(t, options.WorkspaceAccess)
assert.Equal(t, *options.WorkspaceAccess.Runs, item.WorkspaceAccess.WorkspaceRunsPermission)
```

If you want to test for each permissions struct independent of each other (in the above example, if one `require` fails the rest of the test won't execute) you should organize them into subtests:
```
t.Run("validate ProjectAccess", func(t *testing.T) { // ur project access assertions here })
t.Run("validate WorkspaceAccess", func (t *testing.T) { // ur workspace access assertions here })
```
that way a `require.NotNil()` failing won't prevent other structs from being asserted.

team_project_access_integration_test.go
```diff
@@ -164,6 +165,102 @@ func TestTeamProjectAccessesAdd(t *testing.T) {
...
+			},
+		}
+
+		tpa, err := client.TeamProjectAccess.Add(ctx, options)
+		defer func() {
+			err := client.TeamProjectAccess.Remove(ctx, tpa.ID)
+			if err != nil {
+				t.Logf("error removing team access (%s): %s", tpa.ID, err)
+			}
+		}()
```
sebasslash, line 245:
```suggestion
	 t.Cleanup(func() {
			err := client.TeamProjectAccess.Remove(ctx, tpa.ID)
			if err != nil {
				t.Logf("error removing team access (%s): %s", tpa.ID, err)
			}
		})
```

type_helpers.go
```diff
@@ -13,6 +13,36 @@ func ProjectAccess(v TeamProjectAccessType) *TeamProjectAccessType {
...
+
+// WorkspaceStateVersionsPermission returns a pointer to the given team access project type.
+func WorkspaceStateVersionsPermission(v WorkspaceStateVersionsPermissionType) *WorkspaceStateVersionsPermissionType {
+	return &v
+}
+
+// WorkspaceStateVersionsPermission returns a pointer to the given team access project type.
+func WorkspaceVariablesPermission(v WorkspaceVariablesPermissionType) *WorkspaceVariablesPermissionType {
+	return &v
+}
```
sebasslash, line 44: I still don't know how I felt about these pointer helpers in general. For me it was always easier/cleaner to use the address operator `&someStr` than `tfe.String(someStr)`. You can leave this if you feel it's helpful :)

sebasslash (changes requested):
Awesome work 🔥 ! My comments are mostly regarding code organization and testing

team_project_access.go
rberecka: This strucutre aligns with the API design of this endpoint. These perms also functionally cannot be part of workspace.go, since those are only for individual workspaces.

On the TeamProject level, there are two groups of permissions: one that only affects the project itself and those that are applied to all of the workspaces in the project. We thought that TeamProjectAccessWorkspaceAccessPermissions would be too long, but based on the nested struct of TeamProjectAccess would be clear enough.

type_helpers.go
rberecka: Jeff is out, so I can't remember what the issue quite was since my go is so new, but we def did also try the &someString to no avail and since it seemed an already established pattern we decided to use it.

team_project_access.go
rberecka: This would break with what is already done on the TeamWorkspace level. While this file is called team access which is a little confusing, here there is the similar pattern of the workspace level permissions inside the access class itself: https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_team_access.go#L62-L123
Moving this would also break with the API design since you set permission on the TeamProject and not the Project itself.

team_project_access.go
joekarl: @sebasslash Thanks!
These really are scoped to team project access only (this current implementation is very close to the situation in the TeamAccess which doesn't prefix or hoist the struct defs into the workspace file).

Would it be better in this case to go with the longer names so that we can keep locality the same as it is with TeamAccess?

sebasslash: Absolutely. I also think we can drop the second `Access`: `TeamProjectAccessWorkspacePermissions{}` and not lose any meaning here. Again these names can turn out to be quite long and difficult to remember, but the general rule would be to scope the name of the struct to the resource it belongs to. 

But yeah, `WorkspaceAccessPermissions` leads me to believe this is a struct that encapsulates all workspace permissions regardless of what resource I'm enabling those permissions for. So it'd might be best to avoid the confusion if these permissions are unique to team project access.

sebasslash (commented):
This PR has my approval pending the naming changes to scope these permission structs to the parent resource `TeamProjectAccess`. Awesome work @rberecka !!

I ran the tests locally ✅

team_project_access.go
rberecka: Thanks @sebasslash! I will work on getting it all renamed by EOD today

team_project_access.go
```diff
@@ -57,14 +58,105 @@ type TeamProjectAccessList struct {
...
 	Team    *Team    `jsonapi:"relation,team"`
 	Project *Project `jsonapi:"relation,project"`
 }
 
+// ProjectPermissions represents the team's permissions on its project
+type TeamProjectAccessProjectPermissions struct {
+	ProjectSettingsPermission ProjectSettingsPermissionType `jsonapi:"attr,settings"`
+	ProjectTeamsPermission    ProjectTeamsPermissionType    `jsonapi:"attr,teams"`
+}
+
```
rberecka, line 76:
```suggestion
// Workspacepermissions represents the team's permission on all workspaces in its project
```

CHANGELOG.md
laurenolivia, line 9: I think when you rebase for some reason this changelog removes your github handle @rberecka

team_project_access.go
laurenolivia: @rberecka great work!

nit: can we update this comment to read `WorkspacePermissions`?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ## Enhancements
 * Added BETA support for including `projects` relationship and `projects-count` attribute to policy_set on create by @hs26gill [#737](https://github.com/hashicorp/go-tfe/pull/737)
 * Added BETA method `AddProjects` and `RemoveProjects` for attaching/detaching policy set to projects by @Netra2104 [#735](https://github.com/hashicorp/go-tfe/pull/735)
+* Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
```
sebasslash, line 8:
```suggestion
* Added BETA support for adding and updating custom permissions to `TeamProjectAccess`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
```

sebasslash (dismissed):
Approved 🚀

CHANGELOG.md
sebasslash: also yeah don't forget your credit! :)

team_project_access.go
```diff
@@ -57,14 +58,106 @@ type TeamProjectAccessList struct {
...
+// Workspacepermissions represents the team's permission on all workspaces in its project
```
rberecka, line 77:
```suggestion
// WorkspacePermissions represents the team's permission on all workspaces in its project
```

laurenolivia (changes requested):
nit: can we capitalize [here](https://github.com/hashicorp/go-tfe/pull/745/files#r1280984474)

laurenolivia (approved):

PR #748, opened 2023-08-07
v1.32.0 Changelog

uturunku1:
## Output from tests

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,18 @@
 # UNRELEASED
 
+# v1.32.0
+
+## Enhancements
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,19 @@
 # UNRELEASED
```
laurenolivia, line 2: Maybe we can add a commented line here that reads something like 
```
Add new entries that have not yet been released in this section
```
So it's clearer to those who contribute, and are not on our team. What do you think?

sebasslash (approved):

PR #756, opened 2023-08-16
TF-7865 Remove beta flags for  custom access for TeamProjectAccess tests

rberecka:
## Description

Removes skipping of Team Project Access tests using custom access now that the feature is fully GA.
Change only includes test code.

## Testing plan

CI should pass

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./...  -v -run TestTeamProjectAccesses  

> PASS
> ok      github.com/hashicorp/go-tfe     12.199s
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # UNRELEASED
 <!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
-
+* Removed beta tags for TeamProjectAccess by @rberecka [#756](https://github.com/hashicorp/go-tfe/pull/756)
 
 # v1.32.0
 
--- a/team_project_access.go
+++ b/team_project_access.go
@@ -15,8 +15,7 @@ var _ TeamProjectAccesses = (*teamProjectAccesses)(nil)
 // TeamProjectAccesses describes all the team project access related methods that the Terraform
 // Enterprise API supports
 //
-// TFE API docs: Documentation will be linked once this feature is available
-// **Note: This functionality is still in BETA and subject to change.**
+// TFE API docs: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/project-team-access
 type TeamProjectAccesses interface {
 	// List all project accesses for a given project.
 	List(ctx context.Context, options TeamProjectAccessListOptions) (*TeamProjectAccessList, error)
--- a/team_project_access_integration_test.go
+++ b/team_project_access_integration_test.go
@@ -166,7 +166,6 @@ func TestTeamProjectAccessesAdd(t *testing.T) {
 	})
 
 	t.Run("with valid options for all custom TeamProject permissions", func(t *testing.T) {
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # UNRELEASED
-<!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
-
+* Added full support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
+the project level to the project itself (`TeamProjectAccessProjectPermissionsOptions`) and all of the workspaces in a project (`TeamProjectAccessWorkspacePermissionsOptions`) by @rberecka [#745](https://github.com/hashicorp/go-tfe/pull/745)<!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
```
Netra2104, line 3: --> is this line supposed to above?

Netra2104: I am referring to:

Netra2104 (dismissed):
Left 1 comment, but looks good!

CHANGELOG.md
rberecka: I guess that it gets removed when there is something added to the section -- but that's maybe also a good question for the CLI team.

sebasslash (changes requested):
👋 We don't need a changelog entry for flag removal. Please ensure that [references](https://github.com/hashicorp/go-tfe/blob/9073f23cd40f852bb2e61e888b992abe3996fa10/team_project_access.go#L19) to "This feature is still in BETA" are removed in the code docs (which are used to generated the godocs).

team_project_access.go
```diff
@@ -16,7 +16,6 @@ var _ TeamProjectAccesses = (*teamProjectAccesses)(nil)
 // Enterprise API supports
 //
 // TFE API docs: Documentation will be linked once this feature is available
```
sebasslash, line 18: Can you add the docs URL here?

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 # v1.32.0
 
 ## Enhancements
-* Added BETA support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
+* Added support for adding and updating custom permissions to `TeamProjectAccesses`. A `TeamProjectAccessType` of `"custom"` can set various permissions applied at
```
sebasslash, line 9: I don't think we should modify previous changelog entries here.

sebasslash (changes requested):
Almost there 🤞. I did give incorrect feedback in my previous review (sorry 😅 ). We should add a changelog entry: `Removed beta tags for TeamProjectAccess by @rberecka` since we _are_ making a documentation change.

nfagerlund (approved):

PR #757, opened 2023-08-16
Update policy_set to support excluded workspaces relation on CREATE

Netra2104:
## Description

This PR provide users with the ability to **not** apply workspaces when creating a new policy set. This feature is currently behind a feature flag in production.

## Testing plan

Integration test is added to complete the testing plan

## External links

[API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/policy-sets#create-a-policy-set)
## Output from tests

```
/Users/netra.mali/go/go1.19.11/bin/go tool test2json -t /Users/netra.mali/Library/Caches/JetBrains/GoLand2023.2/tmp/GoLand/___TestPolicySetsCreate_with_policies_and_excluded_workspaces_provided_in_github_com_hashicorp_go_tfe.test -test.v -test.paniconexit0 -test.run ^\QTestPolicySetsCreate\E$/^\Qwith_policies_and_excluded_workspaces_provided\E$
=== RUN   TestPolicySetsCreate
--- PASS: TestPolicySetsCreate (2.35s)
=== RUN   TestPolicySetsCreate/with_policies_and_excluded_workspaces_provided
    --- PASS: TestPolicySetsCreate/with_policies_and_excluded_workspaces_provided (1.42s)
PASS
```

PR #761, opened 2023-09-01
Add support for attaching/detaching workspace-exclusions to policy sets

hs26gill:
## Description

User should be able to attach/detach WorkspaceExclusions using `POST/DELETE policy_sets/:policy_set_id/relationships/workspace-exclusions` in the policy_set resource similar to AddWorkspaces and RemoveWorkspaces.

## Testing plan
 
Integration Tests are added 

## External links

- [Related PR](https://github.com/hashicorp/go-tfe/pull/757)

## Output from tests

```

...
```

### DIFF
--- a/policy_set.go
+++ b/policy_set.go
@@ -56,6 +56,14 @@ type PolicySets interface {
 	// Remove workspaces from a policy set.
 	RemoveWorkspaces(ctx context.Context, policySetID string, options PolicySetRemoveWorkspacesOptions) error
 
+	// Add workspace exclusions to a policy set.
+	// **Note: This function is still in BETA and subject to change.**
+	AddWorkspaceExclusions(ctx context.Context, policySetID string, options PolicySetAddWorkspaceExclusionsOptions) error
+
+	// Remove workspace exclusions from a policy set.
+	// **Note: This function is still in BETA and subject to change.**
+	RemoveWorkspaceExclusions(ctx context.Context, policySetID string, options PolicySetRemoveWorkspaceExclusionsOptions) error
+
 	// Add projects to a policy set.
 	AddProjects(ctx context.Context, policySetID string, options PolicySetAddProjectsOptions) error
 
@@ -269,6 +277,18 @@ type PolicySetRemoveWorkspacesOptions struct {
 	Workspaces []*Workspace
 }
 
+// PolicySetAddWorkspaceExclusionsOptions represents the options for adding workspace exclusions to a policy set.
+type PolicySetAddWorkspaceExclusionsOptions struct {
+	// The workspaces to add to the policy set exclusion list.
+	WorkspaceExclusions []*Workspace
+}
+
+// PolicySetRemoveWorkspaceExclusionsOptions represents the options for removing workspace exclusions from a policy set.
+type PolicySetRemoveWorkspaceExclusionsOptions struct {
+	// The workspaces to remove from the policy set exclusion list.
+	WorkspaceExclusions []*Workspace
+}
+
 // PolicySetAddProjectsOptions represents the options for adding projects
 // to a policy set.
 type PolicySetAddProjectsOptions struct {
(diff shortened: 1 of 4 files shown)


policy_set_integration_test.go
```diff
@@ -754,6 +754,138 @@ func TestPolicySetsRemoveWorkspaces(t *testing.T) {
 	})
 }
 
+func TestPolicySetsAddWorkspaceExclusions(t *testing.T) {
+	client := testClient(t)
```
Netra2104, line 758: skipUnlessBeta?

policy_set.go
```diff
@@ -524,6 +582,16 @@ func (o PolicySetRemoveWorkspacesOptions) valid() error {
 	return nil
 }
 
+func (o PolicySetRemoveWorkspaceExclusionsOptions) valid() error {
+	if o.WorkspaceExclusions == nil {
+		return ErrWorkspacesRequired
+	}
+	if len(o.WorkspaceExclusions) == 0 {
+		return ErrWorkspaceMinLimit
+	}
```
brandonc, line 591: 💅 Just pointing out this is slightly redundant since len(nil) == 0 but I know the pattern is probably elsewhere.

```suggestion
	if len(o.WorkspaceExclusions) == 0 {
		return ErrWorkspacesRequired
	}
```

policy_set.go
```diff
@@ -453,6 +475,42 @@ func (s *policySets) RemoveWorkspaces(ctx context.Context, policySetID string, o
...
+func (s *policySets) AddWorkspaceExclusions(ctx context.Context, policySetID string, options PolicySetAddWorkspaceExclusionsOptions) error {
+	if !validStringID(&policySetID) {
+		return ErrInvalidPolicySetID
+	}
+	if err := options.valid(); err != nil {
+		return err
+	}
+
+	u := fmt.Sprintf("policy-sets/%s/relationships/workspace-exclusions", url.QueryEscape(policySetID))
+	req, err := s.client.NewRequest("POST", u, options.WorkspaceExclusions)
```
brandonc, line 488: Slightly confused by this usage-- wouldn't you use "options" as the body and decorate it with json annotations to create a request body? This appears as thought the request body is expected to be an array... I don't think that is a typical pattern (see allowed-workspaces on the agent-pools endpoint)

The awkward thing with this implementation is that you would not be able to meaningfully add fields to PolicySetAdd/RemoveWorkspaceExclusionsOptions and have them appear in the request body.

brandonc (changes requested):
See ❓ below

policy_set.go
hs26gill: I saw this redundancy but only kept it due to repeated pattern. I can clean this up.

policy_set.go
hs26gill: Thanks! I will update this

policy_set.go
hs26gill: @brandonc We do not need the test for https://github.com/hashicorp/go-tfe/blob/hs26gill/TF-7895-Add-delete-policy-sets-relationships-exclusions/policy_set_integration_test.go#L870 as well then right?

policy_set.go
hs26gill: Seems like it fails the tests and returns 500. I am trying to pass an array of `workspaces` under `workspace-exclusions` similar to the `AddWorkspacesOptions` and `AddProjectOptions`. Please see: https://github.com/hashicorp/go-tfe/blob/hs26gill/TF-7895-Add-delete-policy-sets-relationships-exclusions/policy_set.go#L254-L278

brandonc: I was the reviewer for that change and didn't notice the problem then 😭

brandonc: This issue feels very familiar, though. I need to audit our relationships controllers to understand this better. Let me get back to you

policy_set.go
brandonc: I think in the future the "required" error is sufficient when you provide either 0 items or nil, but I don't mind keeping it consistent!

policy_set.go
brandonc: It appears I was wrong and this is the dominant pattern! And I can't think of a better way of implementing it, either.

brandonc (dismissed):

policy_set.go
hs26gill: Sounds good, thanks!

brandonc (approved):

PR #762, opened 2023-09-01
support organization default execution mode

SwiftEngineer:
## Description

Adds support for organization-level default execution modes and agent pools. 

## Output from tests

```
=== RUN   TestOrganizationsUpdate/with_default_execution_mode_and_without_agent_pool
    --- PASS: TestOrganizationsUpdate/with_default_execution_mode_and_without_agent_pool (4.83s)
=== RUN   TestOrganizationsUpdate/when_only_updating_a_subset_of_fields
    --- PASS: TestOrganizationsUpdate/when_only_updating_a_subset_of_fields (3.44s)
=== RUN   TestOrganizationsUpdate/with_default_execution_mode_of_'agent'
    --- PASS: TestOrganizationsUpdate/with_default_execution_mode_of_'agent' (6.23s)
=== RUN   TestOrganizationsUpdate/with_default_execution_mode_of_'remote'
    --- PASS: TestOrganizationsUpdate/with_default_execution_mode_of_'remote' (6.68s)
=== RUN   TestWorkspacesCreate/when_no_execution_mode_is_not_specified_in_an_organization_with_local_as_default_execution_mode
--- PASS: TestWorkspacesCreate (7.93s)
    --- PASS: TestWorkspacesCreate/when_no_execution_mode_is_not_specified_in_an_organization_with_local_as_default_execution_mode (3.99s)
=== RUN   TestWorkspacesCreate
--- PASS: TestWorkspacesCreate (12.17s)
=== RUN   TestWorkspacesCreate/when_organization_has_a_default_execution_mode
    --- PASS: TestWorkspacesCreate/when_organization_has_a_default_execution_mode (8.70s)
=== RUN   TestWorkspacesCreate/when_organization_has_a_default_execution_mode/with_setting_overwrites_set_to_false,_workspace_inherits_the_default_execution_mode
        --- PASS: TestWorkspacesCreate/when_organization_has_a_default_execution_mode/with_setting_overwrites_set_to_false,_workspace_inherits_the_default_execution_mode (1.02s)
=== RUN   TestWorkspacesCreate/when_organization_has_a_default_execution_mode/with_setting_overwrites_set_to_true,_workspace_ignores_the_default_execution_mode
        --- PASS: TestWorkspacesCreate/when_organization_has_a_default_execution_mode/with_setting_overwrites_set_to_true,_workspace_ignores_the_default_execution_mode (0.92s)
=== RUN   TestWorkspacesCreate/when_organization_has_a_default_execution_mode/when_explicitly_setting_execution_mode,_workspace_ignores_the_default_execution_mode
        --- PASS: TestWorkspacesCreate/when_organization_has_a_default_execution_mode/when_explicitly_setting_execution_mode,_workspace_ignores_the_default_execution_mode (0.92s)
=== RUN   TestWorkspacesRead
--- PASS: TestWorkspacesRead (14.50s)
=== RUN   TestWorkspacesRead/when_workspace_is_inheriting_the_default_execution_mode
    --- PASS: TestWorkspacesRead/when_workspace_is_inheriting_the_default_execution_mode (8.91s)
=== RUN   TestWorkspacesRead/when_workspace_is_inheriting_the_default_execution_mode/and_workspace_execution_mode_is_default
        --- PASS: TestWorkspacesRead/when_workspace_is_inheriting_the_default_execution_mode/and_workspace_execution_mode_is_default (0.83s)
=== RUN   TestWorkspacesUpdateWithDefaultExecutionMode
--- PASS: TestWorkspacesUpdateWithDefaultExecutionMode (10.60s)
=== RUN   TestWorkspacesUpdateWithDefaultExecutionMode/when_explicitly_setting_execution_mode,_workspace_ignores_the_default_execution_mode
    --- PASS: TestWorkspacesUpdateWithDefaultExecutionMode/when_explicitly_setting_execution_mode,_workspace_ignores_the_default_execution_mode (0.79s)
=== RUN   TestWorkspacesUpdateWithDefaultExecutionMode/with_setting_overwrites_set_to_true,_workspace_ignores_the_default_execution_mode
    --- PASS: TestWorkspacesUpdateWithDefaultExecutionMode/with_setting_overwrites_set_to_true,_workspace_ignores_the_default_execution_mode (0.83s)
=== RUN   TestWorkspacesUpdateWithDefaultExecutionMode/with_setting_overwrites_set_to_false,_workspace_inherits_the_default_execution_mode
    --- PASS: TestWorkspacesUpdateWithDefaultExecutionMode/with_setting_overwrites_set_to_false,_workspace_inherits_the_default_execution_mode (0.91s)
PASS
```


organization.go
```diff
@@ -197,6 +199,9 @@ type OrganizationCreateOptions struct {
 
 	// Optional: AllowForceDeleteWorkspaces toggles behavior of allowing workspace admins to delete workspaces with resources under management.
 	AllowForceDeleteWorkspaces *bool `jsonapi:"attr,allow-force-delete-workspaces,omitempty"`
+
+	// Optional: DefaultExecutionMode the default execution mode for workspaces
+	DefaultExecutionMode *string `jsonapi:"attr,default-execution-mode,omitempty"`
```
JarrettSpiker, line 204: It looks like we are missing DefaultAgentPool in the create options

workspace.go
```diff
@@ -379,6 +385,15 @@ type WorkspaceCreateOptions struct {
 	// exist, it is created and added to the workspace.
 	Tags []*Tag `jsonapi:"relation,tags,omitempty"`
 
+	// Struct containing boolean values that indicate whether the resolved value of
+	// a setting should be decided by something other than the workspace (i.e. a related
+	// organization or project).
+	//
+	// For example: When setting the execution-mode of a workspace to  "default", the
+	// `setting-overwrites.execution-mode` field should be set to true, because the execution-mode
```
JarrettSpiker, line 393: This is backwards. If the value is `true`, that means that the workspace is overwriting the value.

workspace.go
```diff
@@ -395,6 +410,12 @@ type VCSRepoOptions struct {
 	GHAInstallationID *string `json:"github-app-installation-id,omitempty"`
 }
 
+type WorkspaceSettingOverwritesOptions struct {
+	// using `json` rather than `json-api` so that nil values are truly omitted. in `jsonapi` the json key will still be present
```
JarrettSpiker, line 414: Wow, that is weird...then what does omitempty even do for json-api?

workspace.go
```diff
@@ -487,6 +508,15 @@ type WorkspaceUpdateOptions struct {
 	// repository.
 	WorkingDirectory *string `jsonapi:"attr,working-directory,omitempty"`
 
+	// Struct containing boolean values that indicate whether the resolved value of
+	// a setting should be decided by something other than the workspace (i.e. a related
+	// organization or project).
+	//
+	// For example: When setting the execution-mode of a workspace to  "default", the
+	// `setting-overwrites.execution-mode` field should be set to true, because the execution-mode
```
JarrettSpiker, line 516: Similar to previous comment, this is backwards. False means inheriting, true means overwriting

JarrettSpiker (changes requested):

organization.go
SwiftEngineer: Yeah, that's actually by design. You need an organization before you can get an agent pool.

workspace.go
SwiftEngineer: ... woops. Good catch!

workspace.go
SwiftEngineer: I think this is more due to my choice of datatype in the schema. I can double check but I saw other parts of go-tfe where we use the same idea.

workspace.go
SwiftEngineer: I think you might be missing the `to "default"` part. Either that, or all the testing I've done for the API is completely wrong

workspace.go
SwiftEngineer: Same response as last one.

emlanctot (dismissed):
looks great 💯

organization_integration_test.go
```diff
@@ -263,6 +281,20 @@ func TestOrganizationsUpdate(t *testing.T) {
 		assert.EqualError(t, err, ErrInvalidOrg.Error())
 	})
 
+	t.Run("with default execution mode and without agent pool", func(t *testing.T) {
```
Uk1288, line 284: test case summary could be clearer, it seems to be doing something in the line of
```suggestion
	t.Run("when specifying remote execution mode with default agent pool", func(t *testing.T) {
```

organization_integration_test.go
```diff
@@ -272,6 +304,29 @@ func TestOrganizationsUpdate(t *testing.T) {
...
+
+	t.Run("with default execution mode of 'agent'", func(t *testing.T) {
+		// this helper creates an organization and then updates it to use a default agent pool
+		org, orgAgentTestCleanup := createOrganizationWithDefaultAgentPool(t, client)
+		assert.Equal(t, "agent", org.DefaultExecutionMode)
+		assert.NotNil(t, org.DefaultAgentPool)
+		t.Cleanup(orgAgentTestCleanup)
+	})
+
+	t.Run("with default execution mode of 'remote'", func(t *testing.T) {
```
Uk1288, line 316: this 2 test cases seem to be doing the same thing, i.e test case
`t.Run("with default execution mode of 'agent'", func(t *testing.T) {...`
and
`t.Run("with default execution mode of 'remote'", func(t...`.

Should they be merged as one?

workspace_integration_test.go
```diff
@@ -621,6 +621,32 @@ func TestWorkspacesCreate(t *testing.T) {
 		assert.Equal(t, err, ErrRequiredAgentPoolID)
 	})
 
+	t.Run("when no execution mode is not specified in an organization with local as default execution mode", func(t *testing.T) {
```
Uk1288, line 624: the double negation made the test summary unclear. 

- [ ]  TODO to update test summary

workspace.go
Uk1288: `+1` on double checking this.

SwiftEngineer: 🖖 Looks like on the majority of the nested objects we also avoid using json-api annotations, like [organization access options](https://github.com/hashicorp/go-tfe/blob/0ef3e908c2106bf5fc194c918da3985c10ffb9b6/team.go#L146-L158).

Seems this bug specifically applies to booleans in nested objects.

organization_integration_test.go
SwiftEngineer: Good catch!

helper_test.go
```diff
@@ -965,6 +965,43 @@ func createOrganizationWithOptions(t *testing.T, client *Client, options Organiz
...
+		DefaultAgentPool:     agentPool,
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return org, func() {
+		// unset execution mode of organization
+		_, err := client.Organizations.Update(ctx, org.Name, OrganizationUpdateOptions{
```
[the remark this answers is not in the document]
SwiftEngineer: I was a little scared to do this, specifically because I know relying on atlas' cleanup logic has caused test flakes in the past, but you're probably right that I was being overprotective in this case. Thanks for the catch!

workspace.go
```diff
@@ -203,6 +204,11 @@ type VCSRepo struct {
 	WebhookURL        string `jsonapi:"attr,webhook-url"`
 }
 
+type WorkspaceSettingOverwrites struct {
+	ExecutionMode *bool `jsonapi:"attr,execution-mode,omitempty"`
+	AgentPool     *bool `jsonapi:"attr,agent-pool,omitempty"`
```
nfagerlund, line 209: Right, so:

- `WorkspaceSettingOverwrites` represents the values in a Workspace struct that we received _from_ the API. 
- `WorkspaceSettingOverwritesOptions` represents the values we may or may not be sending _to_ the API in a create or update options payload. 

Thus, this struct (the value one) should use plain `bool`s, and should not be using `omitempty`.

...I think! The theory here is that when you get a value back from the API, it should ALWAYS have either a true or false value for each of these, so we reflect that in the type system by not allowing a nil pointer as a third option. But, push back and correct me if I'm wrong in that — for example, I haven't yet tried running against an old TFE that doesn't provide this field (or doesn't provide a specific field in it, which is something we'll run into as the struct grows).

For prior art, see the nested VCSRepo / VCSRepoOptions structs, which are also doing this exact thing.

workspace.go
```diff
@@ -380,6 +386,15 @@ type WorkspaceCreateOptions struct {
 	// exist, it is created and added to the workspace.
 	Tags []*Tag `jsonapi:"relation,tags,omitempty"`
 
+	// Struct containing boolean values that indicate whether the resolved value of
+	// a setting should be decided by something other than the workspace (i.e. a related
+	// organization or project).
+	//
+	// For example: When setting the execution-mode of a workspace to "default", the
+	// `setting-overwrites.execution-mode` field should be set to false, because the execution-mode
+	// will be overwritten by either the organization or project.
```
nfagerlund, line 395: There are a few problems with this doc string: 

- There's no such value as "default", but it makes it look like there might be. 
- The "whether" is backwards! When describing a bool, the "whether" condition should align with the `true` value and the "or not" condition with the `false`. 

How about:

```suggestion
	// Optional: Struct of booleans, which indicate whether the workspace
	// specifies its own values for various settings. If you mark a setting as
	// `false` in this struct, it will clear the workspace's existing value for
	// that setting and defer to the default value that its project or
	// organization provides.
	//
	// In general, it's not necessary to mark a setting as `true` in this
	// struct; if you provide a literal value for a setting, Terraform Cloud will
	// automatically update its overwrites field to `true`. If you do choose to
	// manually mark a setting as overwritten, you must provide a value for that
	// setting at the same time. 
```

workspace.go
```diff
@@ -488,6 +509,15 @@ type WorkspaceUpdateOptions struct {
 	// repository.
 	WorkingDirectory *string `jsonapi:"attr,working-directory,omitempty"`
 
+	// Struct containing boolean values that indicate whether the resolved value of
+	// a setting should be decided by something other than the workspace (i.e. a related
+	// organization or project).
+	//
+	// For example: When setting the execution-mode of a workspace to "default", the
+	// `setting-overwrites.execution-mode` field should be set to false, because the execution-mode
+	// will be overwritten by either the organization or project.
```
nfagerlund, line 518: Same suggestion for this instance as for the occurrence above.

workspace.go
```diff
@@ -396,6 +411,12 @@ type VCSRepoOptions struct {
 	GHAInstallationID *string `json:"github-app-installation-id,omitempty"`
 }
 
+type WorkspaceSettingOverwritesOptions struct {
+	// using `json` rather than `json-api` so that nil values are truly omitted. in `jsonapi` the json key will still be present
+	ExecutionMode *bool `json:"execution-mode,omitempty"`
+	AgentPool     *bool `json:"agent-pool,omitempty"`
```
nfagerlund, line 417: This whole scheme is extremely indirect, so let's start things off right by establishing some conventions! In the create/update options version of these overwrites struct, let's document each field with the exact identity of the parent setting that gets used if the workspace doesn't set its own value. (We can be lazy in the value version of the struct, because the whole point of the scheme is so that people can just usually just read the top-level settings values and go home instead of doing inheritance math.)

```suggestion

	// If false, the workspace will defer to its organization's DefaultExecutionMode value.
	ExecutionMode *bool `json:"execution-mode,omitempty"`
	// If false, the workspace will defer to its organization's DefaultAgentPool value.
	AgentPool     *bool `json:"agent-pool,omitempty"`
```

(Extra blank line at the top is so nothing interprets your comment about jsonapi field tags as being part of the documentation for the first field.)

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 # UNRELEASED
+<!-- Add CHANGELOG entry to this section for any PR awaiting the next release -->
+<!-- Please also include if this is a Bug Fix, Enhancement, or Feature -->
+
+## Features
+* Added support for assigning agent pools at the organization level via a `default` agent pool by @SwiftEngineer [#762](https://github.com/hashicorp/go-tfe/pull/762)
```
nfagerlund, line 6: This is a little incomplete! It should probably be multiple bullet points. 

- Org-level default agent pool
- Org-level default execution mode
- New `WorkspaceSettingOverwritesOptions` field for allowing workspaces to defer some settings to a default from their org or project.

nfagerlund (changes requested):
Ok, this is in fairly solid shape! I have one code change to request (value version of overwrites doesn't need pointers), and several more documentation requests on account of how challenging this new pattern is going to be to our users. 

I ran the new code interactively in my test project, and it works as advertised!

workspace.go
SwiftEngineer: I think that's true, EXCEPT if we end up adding new values to this object in future updates. Presumably, folks then modify the old values to be pointers as well for the sake of consistency. I'm assuming that's why in `VCSRepoOptions` today, the values are all pointers with `omitempty`.

I'm happy to modify the `ExecutionMode` and `AgentPool` values to be included today, since I think you are TECHNICALLY correct, but, unless I'm mistaken, it seems like that will lead to inconsistencies when someone adds the next "setting-overwrites". OR worse yet, they'll think they don't need the `omitempty` in which case, their code will break on older versions of TFE.

workspace.go
SwiftEngineer: Lovely!

workspace.go
SwiftEngineer: Love this! Only caveat I might add is that currently we are working on allowing the project to overwrite this value too 😓 So I'll just slightly modify what you've suggested here!

CHANGELOG.md
SwiftEngineer: Good callout on adding `WorkspaceSettingOverwritesOptions` on there and the extra backticks!

I made one small change in that I merged the top two suggestions ("Org-level default agent pool" and "Org-level default execution mode") into one. That's because I don't want to give users the impression that they can set a default agent pool WITHOUT setting a default execution mode. In other words, the default agent pool depends on the default execution mode being set to `agent`.

workspace.go
```diff
@@ -203,6 +204,11 @@ type VCSRepo struct {
 	WebhookURL        string `jsonapi:"attr,webhook-url"`
 }
 
+type WorkspaceSettingOverwrites struct {
```
nfagerlund, line 207:
```suggestion
// Note: the fields of this struct are bool pointers instead of bool values, in order to simplify support for
// future TFE versions that support *some but not all* of the inherited defaults that go-tfe knows about.
type WorkspaceSettingOverwrites struct {
```

workspace.go
SwiftEngineer: For the lurkers out there: Nick and I chatted about this in the #team-tf-cli slack channel. The end result is we'd like to use pointers SOLELY in the interest of making supporting backwards-compatibility easier. By allowing these fields to be pointers, users can check to see if the attributes are `nil` or `false`, which can give go-tfe insight into what version of TFE it is interacting with.

nfagerlund (approved):
🚀

PR #765, opened 2023-09-07
Include parameter validation is a maintenance problem for new valid include options

brandonc:
## Description

We have forgotten to add new ones to the validation, the server will validate them for us, and the string is already typed. Note that the API behavior is identical with the same error being returned but only after a round trip to the server.

Example server error response (400)

```
{
  "errors": [
    {
      "status": "400",
      "title": "bad request",
      "detail": "Invalid include parameter"
    }
  ]
}
```

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -887,6 +887,16 @@ func checkResponseCode(r *http.Response) error {
 	var err error
 
 	switch r.StatusCode {
+	case 400:
+		errs, err = decodeErrorPayload(r)
+		if err != nil {
+			return err
+		}
+
+		if errorPayloadContains(errs, "Invalid include parameter") {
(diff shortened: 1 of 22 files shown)


tfe.go
```diff
@@ -887,6 +887,16 @@ func checkResponseCode(r *http.Response) error {
 	var err error
 
 	switch r.StatusCode {
+	case 400:
```
laurenolivia, line 890: Nice.

laurenolivia (approved):

brandonc:
Updated CHANGELOG https://github.com/hashicorp/go-tfe/commit/b89bf096a58a1ad405e397b6ef6eda3a3028243f

PR #770, opened 2023-09-12
TF-9274 Remove beta flags for PolicySetProjects and PolicySetWorkspaceExclusions

Netra2104:
## Description

Removes skipping of Projects and Workspace Exclusions associated with a policy set now that the feature is fully GA. 

## Testing plan

CI should pass

## Output from tests

```

...
```


brandonc (approved):
The speediest rollout process ever! Nice

PR #775, opened 2023-09-15
Add PR Templates

barrettclark:


laurenolivia (changes requested):
Sweet. I think we should still keep the `pull_request_template.md` because is creates a clean description section for every new PR that is created. Similar to the [provider](https://github.com/hashicorp/terraform-provider-tfe/blob/main/.github/pull_request_template.md).

barrettclark:
@laurenolivia I put the general template back in place. Thanks for noting that.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,48 @@
...
+your issue relates to the Terraform Cloud/Enterprise platform itself, please contact
+tf-cloud@hashicorp.support. For general usage questions, please post to our community forum:
+https://discuss.hashicorp.com.
+-->
+
+#### go-tfe version
+<!---
+What version of go-tfe are you using?
+-->
+
```
laurenolivia, line 21: What do you think about adding a plaintext section here and for `Description`, `Testing plan`?
Like:
```plaintext
...
```

barrettclark: I'm adding it for the go-tfe version and testing plan. We don't seem to have a consistent way or reason that we include plaintext fields in the templates.

laurenolivia (approved):

PR #777, opened 2023-09-20
Update OAuthClient creation params

roleesinhaHC:
## Output from tests

```

...
```

### DIFF
--- a/oauth_client.go
+++ b/oauth_client.go
@@ -80,6 +80,7 @@ type OAuthClient struct {
 	Secret              string              `jsonapi:"attr,secret"`
 	ServiceProvider     ServiceProviderType `jsonapi:"attr,service-provider"`
 	ServiceProviderName string              `jsonapi:"attr,service-provider-display-name"`
+	AgentPoolID         string              `jsonapi:"attr,agent-pool-id"`
 
 	// Relations
 	Organization *Organization `jsonapi:"relation,organization"`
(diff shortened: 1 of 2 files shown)


brandonc (commented):
I do believe this field is a relation on the API, meaning it should be defined as something like... 

```
AgentPool *AgentPool `jsonapi:"relation,agent-pool"`
```

Is this something different? Typically we see these relationships not expressed as attributes in our API.

oauth_client.go
```diff
@@ -134,6 +135,9 @@ type OAuthClientCreateOptions struct {
 
 	// Required: The VCS provider being connected with.
 	ServiceProvider *ServiceProviderType `jsonapi:"attr,service-provider"`
+
+	// Optional: AgentPoolID to associate the VCS Provider with, for PrivateVCS support
+	AgentPoolID *string `jsonapi:"attr,agent-pool-id,omitempty"`
```
brandonc, line 140: This should also be a relation. What does the API expect? agent_pool_id attribute? Is it too late to change it to be more conventional?

brandonc (changes requested):

oauth_client.go
roleesinhaHC: Yes, the API expects agent pools external_id as an attribute. Thats how the API is called from the frontend.

brandonc: That does not seem to be documented already, so is this new API? Can it be moved to the relationships key? This would bring it into convention with other APIs, like [creating an agent pool with allowed workspaces](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/agents#request-body) and [creating a run that references a configuration version and workspace](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run#create-a-run) and [updating a workspace to be in a different project](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#update-a-workspace)

roleesinhaHC: Yes, thats a new API for Private VCS project, which has only the UI changes in scope. We built just one API to support the integration tests. 
Thanks for sharing the other APIs, I'll look into those and follow the convention.

PR #778, opened 2023-09-21
TF-9286 Add the enforced attribute to variable_set

Netra2104:
## Description

The variable sets has a new attribute called enforced to disable overwrites and hence go-tfe needs to be updated to accommodate the new attribute as well for POST and PATCH.

## Testing plan

Integration tests.

## External links

- [RFC](https://hermes.hashicorp.services/document/1l13bdkwiPAqMLMUYYjPTPwD1ipcVTMhIMkDT-KDQwhk?draft=true)
- [PRD](https://docs.google.com/document/d/1GU-o1QiNxCzCqVAvyfj1FUzPrU1YWfoUIA9QKHHMFZw/edit)
- [Jira](https://hashicorp.atlassian.net/browse/TF-9286)

## Output from tests

```
=== RUN   TestVariableSetsCreate
--- PASS: TestVariableSetsCreate (1.45s)
=== RUN   TestVariableSetsCreate/with_valid_options
    --- PASS: TestVariableSetsCreate/with_valid_options (0.36s)
PASS


=== RUN   TestVariableSetsUpdate
--- PASS: TestVariableSetsUpdate (1.61s)
=== RUN   TestVariableSetsUpdate/when_updating_a_subset_of_values
    --- PASS: TestVariableSetsUpdate/when_updating_a_subset_of_values (0.18s)
=== RUN   TestVariableSetsUpdate/when_options_has_an_invalid_variable_set_ID
    --- PASS: TestVariableSetsUpdate/when_options_has_an_invalid_variable_set_ID (0.00s)
PASS
```

### DIFF
--- a/variable_set_test.go
+++ b/variable_set_test.go
@@ -171,6 +171,7 @@ func TestVariableSetsCreate(t *testing.T) {
 			Name:        String("varset"),
 			Description: String("a variable set"),
 			Global:      Bool(false),
+			Enforced:    Bool(false),
 		}
 
 		vs, err := client.VariableSets.Create(ctx, orgTest.Name, &options)
@@ -188,6 +189,7 @@ func TestVariableSetsCreate(t *testing.T) {
 			assert.Equal(t, *options.Name, item.Name)
 			assert.Equal(t, *options.Description, item.Description)
 			assert.Equal(t, *options.Global, item.Global)
+			assert.Equal(t, *options.Enforced, item.Enforced)
 		}
 	})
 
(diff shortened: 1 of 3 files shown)


variable_set_test.go
```diff
@@ -232,6 +235,7 @@ func TestVariableSetsRead(t *testing.T) {
 }
 
 func TestVariableSetsUpdate(t *testing.T) {
+	skipUnlessBeta(t)
```
brandonc, line 238: I have a slight worry that when we add a skip to an existing test it can become skipped for too long and my preference is that we create a second test for "enforced" or leave out the existing test until the beta label is removed. On the other hand, you've been really good about un-beta-ing features in the past. I'll leave it up to you based on what you know about this new feature.

brandonc (approved):

variable_set_test.go
Netra2104: This feature will be out within 3 to 4 weeks, but since it is a major feature I agree with you.

brandonc (approved):
Perfecto

PR #786, opened 2023-10-09
[TF-9914] Add auto destroy attribute to workspaces

notchairmk:
## Description

Updates `Workspaces` to include the `AutoDestroyAt` attribute. This attribute is settable on create and update.

Adds a `Time` receiver to the existing type helpers.
Adds an `Nullable` generic type with custom marshaling to allow updates to attributes which use significant `null` values, but which also require `omitempty` behavior.

## Testing plan

Updated workspace integration tests

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#create-a-workspace)

## Output from tests

```
PASS TestWorkspacesUpdate/with_valid_options (0.28s)
      --- PASS: TestWorkspacesUpdate/with_valid_options (0.28s)
PASS TestWorkspacesCreate/with_valid_options (0.34s)
      --- PASS: TestWorkspacesCreate/with_valid_options (0.34s)
```

### DIFF
--- a/.github/actions/lint-go-tfe/action.yml
+++ b/.github/actions/lint-go-tfe/action.yml
@@ -30,15 +30,7 @@ runs:
 
     - name: Install mockgen
       shell: bash
-      run: |
-        set -eux -o pipefail
-        MOCKGEN_VERSION=$(curl -Ls -o /dev/null -w %{url_effective}   https://github.com/golang/mock/releases/latest | awk -F'/v' '{printf$2}')
-        MOCKGEN_ZIP=mock_${MOCKGEN_VERSION}_linux_amd64.tar.gz
-        curl -OL https://github.com/golang/mock/releases/download/v$MOCKGEN_VERSION/$MOCKGEN_ZIP
-        sudo tar xvzf $MOCKGEN_ZIP --strip-components 1 -C /usr/local
-        sudo chmod +x /usr/local/mockgen
-        rm -f $MOCKGEN_ZIP
-        echo /usr/local/ >> $GITHUB_PATH
+      run: go install github.com/golang/mock/mockgen@v1.6.0
 
     - name: Get dependencies
       shell: bash
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Enhancements
+* Updates `Workspaces` to include an `AutoDestroyAt` attribute on create and update by @notchairmk [#786](https://github.com/hashicorp/go-tfe/pull/786)
+
 # v1.43.0
 
 ## Features
--- a/type_helpers.go
+++ b/type_helpers.go
@@ -4,7 +4,9 @@
 package tfe
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
 	"reflect"
 	"time"
 )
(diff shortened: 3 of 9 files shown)


nfagerlund (commented):
I have questions about this one before I can approve it! (And if there's good answers already, I'll probably push to get them into docs for this attribute.)

From a glance at the [workspace api docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#update-a-workspace), it seems this is a field where a string value means auto-destroy is turned on (with the value of the string being the deadline), but a `null` value means auto-destroy is turned OFF. 

I presume this is why the field on the Workspace struct is a `*time.Time` pointer instead of a value like the other times!

But: how do I _turn auto-destroy off for a workspace?_ 

In the CreateOptions / UpdateOptions structs, you've set the field to `omitempty` for the serializer, so would a nil pointer do nothing? It seems like it might be impossible to turn it off once it's turned on. 

I haven't tested this yet! If I'm wrong, please add a new integration test demonstrating turning auto-destroy on and off again.

notchairmk:
Removed `omitempty` from create and update options (probably don't need to remove it from the create options, but I personally don't have strong feelings one way or the other). I'm not sure if the current implementation works against the grain for patterns elsewhere in the repo, so definitely looking for feedback.

notchairmk:
it seems like there are maybe two ways that I can think of to do this (also open to suggestions):

1. add to existing update options without `omitempty`. this is the first implementation, but has the side effect of potentially allowing the user to unintentionally unset `AutoDestroyAt`
1. create a distinct remove action, allow `omitempty` on create and update options. this feels like the preferred path.

let me know if there's another route we should take.

to fix linting I'd probably break the Sprintf calls for `"organizations/%s/workspaces/%s"` into their own func. but will hold off until I hear further

workspace.go
```diff
@@ -846,6 +855,35 @@ func (s *workspaces) SafeDeleteByID(ctx context.Context, workspaceID string) err
...
+		return nil, err
+	}
+
+	w := &Workspace{}
+	err = req.Do(ctx, w)
+	if err != nil {
+		return nil, err
+	}
+
+	return w, nil
```
sebasslash, line 884: I personally like this approach much more, given the significance of a `null` value for this field.

.github/actions/lint-go-tfe/action.yml
```diff
@@ -30,15 +30,7 @@ runs:
...
-      run: |
-        set -eux -o pipefail
-        MOCKGEN_VERSION=$(curl -Ls -o /dev/null -w %{url_effective}   https://github.com/golang/mock/releases/latest | awk -F'/v' '{printf$2}')
-        MOCKGEN_ZIP=mock_${MOCKGEN_VERSION}_linux_amd64.tar.gz
-        curl -OL https://github.com/golang/mock/releases/download/v$MOCKGEN_VERSION/$MOCKGEN_ZIP
-        sudo tar xvzf $MOCKGEN_ZIP --strip-components 1 -C /usr/local
-        sudo chmod +x /usr/local/mockgen
-        rm -f $MOCKGEN_ZIP
-        echo /usr/local/ >> $GITHUB_PATH
+      run: go install github.com/golang/mock/mockgen@v1.6.0
```
ctrombley, line 33: This was needed to help `mockgen` parse generics. It seems that installing it manually caused it to reference an older version of go?

workspace.go
```diff
@@ -481,6 +485,9 @@ type WorkspaceUpdateOptions struct {
 	// from another workspace.
 	AutoApplyRunTrigger *bool `jsonapi:"attr,auto-apply-run-trigger,omitempty"`
 
+	// Optional: The time after which an automatic destroy run will be queued
+	AutoDestroyAt *Unsettable[time.Time] `jsonapi:"attr,auto_destroy_at,omitempty"`
```
notchairmk, line 489: this not matching the shape of `AutoDestroyAt` values in the other Workspace structs doesn't feel great 😕

type_helpers.go
```diff
@@ -117,3 +128,70 @@ func SMTPAuthValue(v SMTPAuthType) *SMTPAuthType {
...
+// This is generally useful for PATCH requests, where attributes with zero
+// values are intentionally not marshaled into the request payload.
+//
+// Helper functions are provided to help avoid pointer wrangling.
+//
+// NOTE: This type is only for use with outbound requests. It will cause errors
+// if used to unmarshal API responses, since the jsonapi module does not yet
+// support custom unmarshaling.
+
+type Unsettable[T any] struct {
```
notchairmk, line 144: my preference would probably be to only export what we want users to directly interact with. if we're using the below helper methods, then I don't think we'd need to export this struct.

type_helpers.go
```diff
@@ -117,3 +128,70 @@ func SMTPAuthValue(v SMTPAuthType) *SMTPAuthType {
...
+	if t == nil || t.Value == nil {
+		return json.RawMessage(`null`), nil
+	}
+
+	var b []byte
+	var err error
+
+	val := reflect.ValueOf(t.Value)
+	if val.Type().Kind() == reflect.Ptr && val.Elem().Type() == reflect.TypeOf(time.Time{}) {
+		return json.Marshal(val.Elem().Interface().(time.Time).Format(iso8601TimeFormat))
```
notchairmk, line 192: should this be this instead?
```suggestion
		return json.Marshal(val.Elem().Interface().(time.Time).Format(time.RFC3339))
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Unreleased
 
+
+# v1.43.0
+
+## Enhancements
+* Updates `Workspaces` to include an `AutoDestroyAt` attribute on create and update by @notchairmk [#786](https://github.com/hashicorp/go-tfe/pull/786)
+
```
sebasslash, line 8: 💅 This entry should go under `# Unreleased`

type_helpers.go
```diff
@@ -117,3 +130,138 @@ func SMTPAuthValue(v SMTPAuthType) *SMTPAuthType {
...
+//
+// Internal implementation details:
+//
+// - map[true]T means a value was provided
+// - map[false]T means an explicit null was provided
+// - nil or zero map means the field was not provided
+//
+// If the field is expected to be optional, add the `omitempty` JSON tags. Do NOT use `*Nullable`!
+//
+// Adapted from https://www.jvt.me/posts/2024/01/09/go-json-nullable/
```
sebasslash, line 153: Fantastic description 👍

type_helpers.go
```diff
@@ -117,3 +130,138 @@ func SMTPAuthValue(v SMTPAuthType) *SMTPAuthType {
...
+}
+
+// NewNullNullable is a convenience helper to allow constructing a Nullable with
+// an explicit `null`, for instance to construct a field inside a struct
+// without introducing an intermediate variable
+func NewNullNullable[T any]() Nullable[T] {
+	var n Nullable[T]
+	n.SetNull()
+	return n
+}
```
sebasslash, line 173: All these methods/types in `type_helpers` are exported publicly. Is this something we want?

workspace.go
```diff
@@ -481,6 +485,9 @@ type WorkspaceUpdateOptions struct {
 	// from another workspace.
 	AutoApplyRunTrigger *bool `jsonapi:"attr,auto-apply-run-trigger,omitempty"`
 
+	// Optional: The time after which an automatic destroy run will be queued
+	AutoDestroyAt Nullable[time.Time] `jsonapi:"attr,auto_destroy_at,omitempty"`
```
sebasslash, line 489: This is so, so nice.

workspace_integration_test.go
```diff
@@ -2644,3 +2648,47 @@ func TestWorkspace_DataRetentionPolicy(t *testing.T) {
...
+		AutoDestroyAt: nil,
+	})
+
+	require.NoError(t, err)
+	assert.NotNil(t, w.AutoDestroyAt)
+
+	// explicitly update the value of auto_destroy_at
+	w, err = client.Workspaces.Update(ctx, orgTest.Name, wTest.Name, WorkspaceUpdateOptions{
+		AutoDestroyAt: NullableTime(time.Date(2025, 1, 2, 0, 0, 0, 0, time.UTC)),
+	})
```
sebasslash, line 2681: 💅 Generally, I like to use `require` over `assert` when it doesn't make sense to continue the test if the assertion fails. `assert` will allow test execution to continue.

sebasslash (changes requested):
This pretty much has my approval pending updating the changelog entry. Comments with 💅 are entirely optional. 

Really nice work 🔥

workspace.go
ctrombley: this is being addressed upstream in [our fork of jsonapi](https://github.com/hashicorp/jsonapi/pull/21).

sebasslash (approved):

PR #796, opened 2023-10-25
TF-9291 Removed beta flag for priority varsets

Netra2104:
## Description

Removed the beta flag since priority variable sets is ready to be released and all development is done.

In the code you will see that I have removed the test. This is done because I had originally created a duplicate of the test leaving everything else the same but with the additional priority attribute. So here I am just removing the duplicate and adding the attribute to the original test.

## Testing plan

CI tests

### DIFF
--- a/variable_set_test.go
+++ b/variable_set_test.go
@@ -171,56 +171,6 @@ func TestVariableSetsCreate(t *testing.T) {
 			Name:        String("varset"),
 			Description: String("a variable set"),
 			Global:      Bool(false),
-		}
-
-		vs, err := client.VariableSets.Create(ctx, orgTest.Name, &options)
-		require.NoError(t, err)
-
-		// Get refreshed view from the API
-		refreshed, err := client.VariableSets.Read(ctx, vs.ID, nil)
-		require.NoError(t, err)
-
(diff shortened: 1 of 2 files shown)


variable_set_test.go
```diff
@@ -286,80 +236,33 @@ func TestVariableSetsRead(t *testing.T) {
...
-func TestVariableSetsUpdateWithPriority(t *testing.T) {
-	skipUnlessBeta(t)
-	client := testClient(t)
-	ctx := context.Background()
-
 	orgTest, orgTestCleanup := createOrganization(t, client)
 	t.Cleanup(orgTestCleanup)
-
 	vsTest, _ := createVariableSet(t, client, orgTest, VariableSetCreateOptions{
 		Name:        String("OriginalName"),
```
laurenolivia, line 242: 👋 Hi @Netra2104 - should the string `OriginalName` have a space similar to `line 243` or should `line 243` be updated to read like `line 242`?

Netra2104: I think it is fine the way that it is. I have set this test up the same way as other tests. I think descriptions can have spaces hence why it may have been setup this way.

laurenolivia (approved):

PR #798, opened 2023-10-31
Add AutoApplyRunTrigger attribute to workspaces

nfagerlund:
## Description

Support the new auto-apply-run-trigger workspace attribute, which is exactly like auto-apply except it has to be a new attribute to avoid surprising anyone accustomed to the prior behavior of hard-denying auto-apply for run triggers. 

This feature is shipped to TFC prod, but not yet available in TFE. 

## Testing plan

- Set the new attribute on a workspace.
- It worked, yey. 

## External links

- [tfe provider PR](https://github.com/hashicorp/terraform-provider-tfe/pull/1123)

## Output from tests

I'm internal, check CI.

### DIFF
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -506,6 +506,34 @@ func TestWorkspacesCreate(t *testing.T) {
 		}
 	})
 
+	t.Run("with valid auto-apply-run-trigger option", func(t *testing.T) {
+		skipIfEnterprise(t)
+		// FEATURE FLAG: auto-apply-run-trigger
+		// Once un-flagged, delete this test and add an AutoApplyRunTrigger field
+		// to the basic "with valid options" test below.
+
+		options := WorkspaceCreateOptions{
+			Name:                String("foo"),
+			AutoApplyRunTrigger: Bool(true),
+		}
+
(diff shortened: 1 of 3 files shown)


brandonc (dismissed):
Just needs BETA labelling since it's behind a FF and maybe a CHANGELOG

nfagerlund:
For posterity: we discussed in slack, and this is not in fact beta — there's a feature flag, but it's rolled out 100% to prod in TFC.

workspace_integration_test.go
```diff
@@ -506,6 +506,34 @@ func TestWorkspacesCreate(t *testing.T) {
 		}
 	})
 
+	t.Run("with valid auto-apply-run-trigger option", func(t *testing.T) {
+		skipUnlessBeta(t)
```
sebasslash, line 510: If this is no longer beta, but only available in TFC we should use the `skipIfEnterprise()` helper

nfagerlund: omg. yeah, thank you!

sebasslash (dismissed):

laurenolivia (approved):

PR #806, opened 2023-11-03
TF-9268 Add support for POST/DELETE oauth_client/:id/relationships/projects

Netra2104:
## Description

Users should be able to add/remove projects using POST/DELETE oauth_clients/:id/relationships/projects in the oauth_client resource.

## Testing plan

Integration test is added to complete the testing plan. The tests are currently set to beta but can be tested locally by removing the beta flag and running the tests.

## External links

- [Related PR](https://hashicorp.atlassian.net/browse/TF-9268)

## Output from tests

```

...
```


oauth_client_integration_test.go
```diff
@@ -384,6 +427,140 @@ func TestOAuthClientsCreateOptionsValid(t *testing.T) {
 	})
 }
 
+func TestOAuthClientsAddProjects(t *testing.T) {
+	skipUnlessBeta(t)
```
brandonc, line 431: I think it should be possible to remove skipUnlessBeta calls from your new tests because these endpoints are not behind a feature flag and the tests can be run.

brandonc (dismissed):
Minor suggestion but it's nonblocking because I'm not sure when your atlas changes were merged. I ran the tests locally!

oauth_client_integration_test.go
brandonc: @Netra2104 It seems like my suggestion was a bad one. It usually takes a day for the test infrastructure to catch up with the platform nightly release. I take it that the 'Remove' endpoint was added today?

Netra2104: Yup it was added today. Also the feature is feature flagged: https://github.com/hashicorp/atlas/pull/17752/files#diff-c72e00c021aa8f6d41e04f01728c117228541e94deda71313ed203e936eb5d17R7

brandonc: Ohhhhh sorry about that; I didn't notice.

brandonc (approved):

PR #815, opened 2023-11-10
Remove unused agent-pool-id field from Workspace model

brandonc:
This is not a response attribute in the API and should never be populated or used.


laurenolivia (dismissed):

nfagerlund (dismissed):
Oh what?? Hilarious! That's from 2020, wonder if we just missed it or if the API changed.

brandonc:
@nfagerlund @laurenolivia I pushed a CHANGELOG commit and that dismissed your approvals.

laurenolivia (approved):

PR #816, opened 2023-11-13
Adds polymorphic relationships for Workspaces, Tokens, and RunTriggers

brandonc:
This PR adds support for polymorphic fields that already exist in the API to Workspaces (locked-by), AuthenticationTokens (created-by), and RunTriggers (sourceable)

The purpose is to illustrate how the underlying jsonapi feature is used in go-tfe and also to run the upstream PR against a more elaborate test suite.

~It uses [this dependency PR](https://github.com/hashicorp/jsonapi/pull/15) so that should be merged and released before this PR is merged.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # UNRELEASED
 
+## Deprecations
+* The `Sourceable` field has been deprecated on `RunTrigger`. Instead, use `SourceableChoice` to locate the non-empty field representing the actual sourceable value by @brandonc [#816](https://github.com/hashicorp/go-tfe/pull/816)
+
+## Features
+* Adds `LockedBy` relationship field to `Workspace` by @brandonc [#816](https://github.com/hashicorp/go-tfe/pull/816)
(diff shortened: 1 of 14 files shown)


nfagerlund (dismissed):
This seems great. I know it's "do not merge," but I'll register my approval for future reference. 

If you're really feeling fired up, the one polymorphic relation that I REALLY got bit by lacking in the past was `Workspace.LockedBy`. I wanted to use it to give a more reliable error description when building out the saved cloud plans apply action in terraform core, but had to give up and give a "usually probably right" error instead. 

I think the options for locked-by are user/team/run -- org token can't lock a workspace.

brandonc:
@nfagerlund What a stroke of luck! Adding `LockedBy` revealed a bug in the jsonapi decoder when the field was omitted from the payload, which is the case when the workspace is not locked. I also added CHANGLOG and will merge this once we have a jsonapi release. Thanks for the suggestion!

nfagerlund:
🙌🏼!! (I had a hunch aiming it at something I tried to use in anger would turn up something interesting 😅)

sebasslash (approved):
🚀 -- Awesome work!

PR #829, opened 2023-12-11
[TF-10829] add new organization setting option

mjyocca:
## Description

Adds new Organization field to expose enablement/disablement of the new VCS status check setting. 

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/organizations)
- [Organization Setting Documentation](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/organizations#aggregated-status-checks)

## Output from tests

```

...
```

### DIFF
--- a/organization.go
+++ b/organization.go
@@ -99,7 +99,7 @@ type Organization struct {
 	SendPassingStatusesForUntriggeredSpeculativePlans bool                     `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
 	RemainingTestableCount                            int                      `jsonapi:"attr,remaining-testable-count"`
 	// Optional: If enabled, SendPassingStatusesForUntriggeredSpeculativePlans needs to be false.
-	AggregatedCommitStatusEnabled bool `jsonapi:"attr,aggregated-commit-status-enabled,omitempty"`
+	AggregatedCommitStatusEnabled bool `jsonapi:"attr,aggregated-commit-status-enabled"`
 	// Note: This will be false for TFE versions older than v202211, where the setting was introduced.
 	// On those TFE versions, safe delete does not exist, so ALL deletes will be force deletes.
 	AllowForceDeleteWorkspaces bool `jsonapi:"attr,allow-force-delete-workspaces"`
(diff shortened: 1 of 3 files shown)


organization.go
```diff
@@ -98,6 +98,8 @@ type Organization struct {
 	TwoFactorConformant                               bool                     `jsonapi:"attr,two-factor-conformant"`
 	SendPassingStatusesForUntriggeredSpeculativePlans bool                     `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans"`
 	RemainingTestableCount                            int                      `jsonapi:"attr,remaining-testable-count"`
+	// Optional: If enabled, SendPassingStatusesForUntriggeredSpeculativePlans needs to be false.
+	AggregatedCommitStatusEnabled bool `jsonapi:"attr,aggregated-commit-status-enabled,omitempty"`
```
brandonc, line 102: No need for omitempty here, since `false` would be given either way.

brandonc (approved):

organization.go
```diff
@@ -213,6 +215,9 @@ type OrganizationCreateOptions struct {
 	// Optional: SendPassingStatusesForUntriggeredSpeculativePlans toggles behavior of untriggered speculative plans to send status updates to version control systems like GitHub.
 	SendPassingStatusesForUntriggeredSpeculativePlans *bool `jsonapi:"attr,send-passing-statuses-for-untriggered-speculative-plans,omitempty"`
 
+	// Optional: If enabled, SendPassingStatusesForUntriggeredSpeculativePlans needs to be false.
+	AggregatedCommitStatusEnabled *bool `jsonapi:"attr,aggregated-commit-status-enabled"`
```
brandonc, line 219: Oh, these Update/Create options should continue to be omitempty, because in this case the empty pointer omits the attribute!

brandonc (changes requested):

brandonc (approved):

lilincmu (dismissed):
Looks good to me!

lilincmu (approved):

brandonc (approved):

PR #836, opened 2024-01-16
Adds AgentsCeilingOverride and RunsCeilingOverride fields in AdminOrganization and AdminOrganizationUpdateOptions structs

alex-ikse:
## Description

Add undocumented `agents-ceiling-override` and `runs-ceiling-override` attributes to Organization admin.

## Testing plan

Used in a modified version of the [tfe provider](https://github.com/alex-ikse/terraform-provider-tfe).

## External links

- [API documentation](https://developer.hashicorp.com/terraform/enterprise/api-docs/admin/organizations#update-an-organization)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # v1.43.0
 
 ## Features
+* Adds `AgentsCeilingOverride` and `RunsCeilingOverride` fields in `AdminOrganization` and `AdminOrganizationUpdateOptions` structs
 * Adds `AggregatedCommitStatusEnabled` field to `Organization` by @mjyocca [#829](https://github.com/hashicorp/go-tfe/pull/829)
 
 ## Enhancements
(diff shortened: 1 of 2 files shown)


brandonc:
I'm sort of confused about this addition so maybe you can help me understand! My understanding of these attributes in particular is that they are not documented nor relevant to Terraform Enterprise and using them requires internal (HashiCorp) access to read and write them in Terraform Cloud. How could it be effective for you if these appear in the SDK?

alex-ikse:
> I'm sort of confused about this addition so maybe you can help me understand! My understanding of these attributes in particular is that they are not documented nor relevant to Terraform Enterprise and using them requires internal (HashiCorp) access to read and write them in Terraform Cloud. How could it be effective for you if these appear in the SDK?

Theses settings are available in TFE admin organization page.
Capture d’écran 2024-01-17 à 23 20 25

alex-ikse:
> `GET /api/v2/organizations/<org name>/subscription`

@brandonc : This endpoint doesn't exists on Terraform Enterprise.

alex-ikse:
> > `GET /api/v2/organizations/<org name>/subscription`
> 
> 
> 
> @brandonc : This endpoint doesn't exists on Terraform Enterprise.

any news @brandonc ?

brandonc:
@alex-ikse Thanks for checking that for me. If you got a 404 from that endpoint, it likely means that you don't have a subscription associated with that organization-- this is what would be expected but also further deepens the mystery about why you are seeing the UI. I'm going to advise the appropriate team inside HashiCorp to use a cloud condition around that UI because it is not relevant to Terraform Enterprise and as such should also not be an option in this SDK.

PR #837, opened 2024-01-16
fix: add missing AdminOrganization field (GlobalProviderSharing)

alex-ikse:
## Description

We need to set the global-provider-sharing attribute with the `tfe` provider. Sadly, this attributes is not managed by go-tfe right now.

## External links

- [API documentation](https://developer.hashicorp.com/terraform/enterprise/api-docs/admin/organizations#update-an-organization)


brandonc (approved):

brandonc:
Thank you for adding this and for the reminder about it!

PR #839, opened 2024-01-18
set expiration time in the future

uturunku1:
## Description

These tests started failing today because they had hardcoded the date for when a token should expire. The expire date was set to today! 
I adjusted the date to always be in the future (a day later than the current time).

## Output from tests

```

...
```


brandonc (approved):
Yay

PR #847, opened 2024-02-05
[IPL-4359] refactor registry module delete API calls

laurenolivia:
## Description

The API calls to the registry module DELETE endpoint have been deprecated and replaced. See API [docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#delete-a-module). This PR adds new DELETE endpoints.

## Output from tests

```

...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -5,6 +5,7 @@ package tfe
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -45,12 +46,16 @@ type RegistryModules interface {
 	ReadVersion(ctx context.Context, moduleID RegistryModuleID, version string) (*RegistryModuleVersion, error)
 
 	// Delete a registry module
+	// Warning: This method is deprecated and will be removed from a future version of go-tfe. Use DeleteByName instead.
 	Delete(ctx context.Context, organization string, name string) error
 
-	// Delete a specific registry module provider
+	// Delete a registry module by name
+	DeleteByName(ctx context.Context, module RegistryModuleID) error
+
+	// Delete a specified provider for the given module along with all its versions
 	DeleteProvider(ctx context.Context, moduleID RegistryModuleID) error
 
-	// Delete a specific registry module version
+	// Delete a specified version for the given provider of the module
 	DeleteVersion(ctx context.Context, moduleID RegistryModuleID, version string) error
 
 	// Update properties of a registry module
@@ -103,6 +108,7 @@ const (
 )
 
 // RegistryModuleID represents the set of IDs that identify a RegistryModule
+// Use NewPublicRegistryModuleID or NewPrivateRegistryModuleID to build one
 type RegistryModuleID struct {
 	// The organization the module belongs to, see RegistryModule.Organization.Name
 	Organization string
@@ -582,6 +588,8 @@ func (r *registryModules) ReadVersion(ctx context.Context, moduleID RegistryModu
 }
 
 // Delete is used to delete the entire registry module
+// Warning: This method is deprecated and will be removed from a future version of go-tfe. Use DeleteByName instead.
+// See API Docs: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#delete-a-module
 func (r *registryModules) Delete(ctx context.Context, organization, name string) error {
 	if !validStringID(&organization) {
 		return ErrInvalidOrg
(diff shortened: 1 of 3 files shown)


registry_module.go
```diff
@@ -45,14 +45,26 @@ type RegistryModules interface {
...
 	DeleteVersion(ctx context.Context, moduleID RegistryModuleID, version string) error
 
+	// Delete a registry module by name
+	DeleteByName(ctx context.Context, module RegistryModuleDeleteByNameOrProvider) error
+
+	// Delete a specific registry module by provider
+	DeleteByProvider(ctx context.Context, module RegistryModuleDeleteByNameOrProvider, provider string) error
+
+	// Delete a specific registry module by version
+	DeleteByVersion(ctx context.Context, moduleID RegistryModuleID, version string) error
```
brandonc, line 66: I think this one might be misnamed because you're not deleting a module by version, you're just deleting a module version [by version].

Also, the method signature is identical, so it's probably OK to just change the implementation of the existing method if that is necessary.

registry_module.go
```diff
@@ -45,14 +45,26 @@ type RegistryModules interface {
...
 
 	// Delete a specific registry module version
+	// Deprecated: Use DeleteByVersion
 	DeleteVersion(ctx context.Context, moduleID RegistryModuleID, version string) error
 
+	// Delete a registry module by name
+	DeleteByName(ctx context.Context, module RegistryModuleDeleteByNameOrProvider) error
+
+	// Delete a specific registry module by provider
+	DeleteByProvider(ctx context.Context, module RegistryModuleDeleteByNameOrProvider, provider string) error
```
brandonc, line 63: DeleteProvider and DeleteVersion already use RegistryModuleID as a parameter, so it doesn't appear that we are adding any information to the API by adding the new methods. Can we just replace the implementation to use the new endpoints and use the data found in RegistryModuleID to request them?

brandonc (changes requested):
Hope my feedback makes sense. We can meet to discuss if you want.

registry_module.go
laurenolivia: Fair - since versions belong to providers, I figured that was implied. But, yes, technically the specified version for the given provider of the module is deleted. So something like `DeleteProviderByVersion` would be much clearer.

laurenolivia: The method signatures for deleting providers by version are identical, but the original `deprecated` API call is a `POST` request and the new call is a `DELETE` request, so wouldn't that be a breaking change?

registry_module.go
laurenolivia: Validation on `RegistryModuleID` is not the same as validation on `RegistryModuleDeleteByNameOrProvider`. I had to build a new struct to accommodate the API calls requiring parameters that weren't originally required. Maybe there's a simpler implementation?

registry_module.go
brandonc: It depends on the details of the API behavior and the deprecation. Is the new method always preferred over the deprecated method? When was the new method introduced? If we need to support the old one in TFE, would it be a good idea to attempt the new DELETE method and then fall back to the old POST method?

registry_module.go
brandonc: You could have two different validation methods on RegistryModuleID --but let's resolve the other conversation first.

registry_module.go
```diff
@@ -690,6 +799,37 @@ func (o RegistryModuleID) valid() error {
...
+		return ErrInvalidName
+	}
+
+	switch o.RegistryName {
+	case PublicRegistry:
+		if !validString(&o.Namespace) {
+			return ErrRequiredNamespace
+		}
+	case PrivateRegistry:
+		if &o.Namespace != &o.Organization {
```
Uk1288, line 821:
```suggestion
		if o.Namespace != o.Organization {
```

registry_module.go
```diff
@@ -690,6 +799,37 @@ func (o RegistryModuleID) valid() error {
...
+			return ErrOrgRequiredPrivateReqistry
+		}
+	case "":
+		// RegistryName is not required when deleting by name
```
Uk1288, line 825: Should `RegistryName` be required? the query doesn't seem to work without the RegistryName

registry_module_integration_test.go
```diff
@@ -1566,6 +1566,261 @@ func TestRegistryModulesDeleteVersion(t *testing.T) {
...
+		err := client.RegistryModules.DeleteByName(ctx, RegistryModuleDeleteByNameOrProvider{
+			Organization: orgTest.Name,
+			RegistryName: badIdentifier,
+			Namespace:    "terraform-aws-modules",
+			Name:         registryModuleTest.Name,
+		})
+		assert.Error(t, err)
+	})
+
+	t.Run("with namespace and when registry name is private", func(t *testing.T) {
```
Uk1288, line 1622: update the description to capture the scenario that is tested, did you mean: `when namespace is invalid and...`?
Additionally, the assertion can be updated to depict the expected error: `assert.Equal(t, err, ErrInvalid...)`

registry_module_integration_test.go
```diff
@@ -1566,6 +1566,261 @@ func TestRegistryModulesDeleteVersion(t *testing.T) {
...
```
Uk1288, line 1686: same here

registry_module.go
```diff
@@ -117,6 +129,19 @@ type RegistryModuleID struct {
...
 }
 
+// RegistryModuleDeleteByNameOrProvider represents the values requires to delete the entire registry module
+type RegistryModuleDeleteByNameOrProvider struct {
+	// The organization the module belongs to, see RegistryModule.Organization.Name
+	Organization string
+	// The name of the module, see RegistryModule.Name
+	Name string
+	// The namespace of the module. For private modules this is the name of the organization that owns the module
+	// Namespace is required for public modules
```
Uk1288, line 139: It seems like both `Namespace` and `RegistryName` is always required for the deleteByNameorProvider to succeed. Can you delete the private registry by name or provider without these params?

laurenolivia: Correct, I will update the comment.

registry_module_integration_test.go
laurenolivia: @Uk1288 can you restate what you are flagging here?

registry_module.go
laurenolivia: Yep. Corrected!

registry_module_integration_test.go
laurenolivia: nvm

registry_module.go
```diff
@@ -607,26 +634,46 @@ func (r *registryModules) Delete(ctx context.Context, organization, name string)
...
 
+	var Namespace string
+
+	if moduleID.RegistryName == PrivateRegistry {
+		Namespace = moduleID.Organization
+	}
+
+	if moduleID.RegistryName == PublicRegistry {
+		Namespace = moduleID.Namespace
+	}
```
brandonc, line 653: Should It be incumbent upon the caller to provide this field within the moduleID struct? There is already a Namespace field there that we are ignoring.

To simplify all of this what if we added two methods for building a moduleID?

`func NewPublicRegistryModuleID(namespace, name, provider string) RegistryModuleID`

`func NewPrivateRegistryModuleID(organization, name, provider string) RegistryModuleID`

Perhaps that would simplify DeleteByName to allow it to use this struct as well?

This could apply to all three methods

brandonc: Also, local variables should always be lowercased although I think we should get rid of this one

registry_module.go
```diff
@@ -638,14 +685,58 @@ func (r *registryModules) DeleteVersion(ctx context.Context, moduleID RegistryMo
...
+
+	u := fmt.Sprintf(
+		"organizations/%s/registry-modules/%s/%s/%s",
+		url.QueryEscape(module.Organization),
+		url.QueryEscape(string(module.RegistryName)),
+		url.QueryEscape(Namespace),
+		url.QueryEscape(module.Name),
+	)
+
+	req, err := r.client.NewRequest("DELETE", u, nil)
```
brandonc, line 739: We should call the original r.Delete as a fallback for 404s here (until it is removed)

brandonc (commented):
It looks like it should work. Just a little API and godoc feedback 👇

brandonc (approved):
Don't forget to add a CHANGELOG entry

PR #854, opened 2024-02-14
Feat: Update the Plans API to be able to fetch & return resource changes

AhmedRaafat14:
## Description

This adds a new function to the `Plans` API (`ReadResourceChanges`) that takes a PlanID and return a list of all resource changes done by the target plan.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -6,6 +6,7 @@ package tfe
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/url"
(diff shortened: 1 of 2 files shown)


brandonc:
Hey, @AhmedRaafat14, thanks for the contribution. Unfortunately, we've made a conscious decision not to include this endpoint in our [public API](https://developer.hashicorp.com/terraform/cloud-docs/api-docs) in order to reserve the right to change it or break it in the future. Including it here would link it to the stability guarantees implied by the go-tfe version. Furthermore, the profuse interface{} types limit the utility of the public types anyway.

I don't discourage you from using this endpoint! I would just try to isolate it in your code as "internal HashiCorp API that isn't covered by their [stability policy](https://developer.hashicorp.com/terraform/enterprise/api-docs/stability-policy)"

AhmedRaafat14:
Hi @brandonc thanks for your reply and these information.

PR #864, opened 2024-03-04
Update an error msg to avoid potential confusion

uturunku1:
## Description

This go-tfe error `workspace is still being processed to discover resources` in combination with this provider-tfe error `To delete this workspace without destroying the managed resources, add force_delete = true to the resource config `created confusion in a customer. 

He thought that the latter error msg was saying that his workspace had resources, when in reality he had no resources. And then he thought that the former error msg was confirming that he had resources associated with that workspace, and that they were being processed.

The customer had the intuition that perhaps these were errors related to having a state version that is in middle of processing changes, but because the errors were saying something different, he opened a ticket with Support.

The change that I am making here is to clarify that, yes, what is "still being processed" is the state version. I was able to verify this by looking at the context where [the atlas error is created](https://github.com/hashicorp/atlas/blob/a7cae87a5bac7894bb9b187bcc50f457ae863b0f/app/controllers/concerns/shared_workspace_actions.rb#L434).

And then I am making this change to provider-tfe to mention that force-delete option can be used for when a state version is being process or when a workspace has resources: https://github.com/hashicorp/terraform-provider-tfe/pull/1274

## Output from tests

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Bug fixes
+* Change the error message for `ErrWorkspaceStillProcessing` to be the same error message returned by the API by @uturunku1 [#864](https://github.com/hashicorp/go-tfe/pull/864)
+
 # v1.47.0
 
 ## Enhancements


errors.go
```diff
@@ -66,7 +66,7 @@ var (
 	// to determine if it is safe to delete. "conflict" followed by newline is used to
 	// preserve go-tfe version compatibility with the error constructed at runtime before it was
 	// defined here.
-	ErrWorkspaceStillProcessing = errors.New("conflict\nworkspace is still being processed to discover resources")
+	ErrWorkspaceStillProcessing = errors.New("conflict\nlatest workspace state is still being processed, and attempting to discover resources")
```
brandonc, line 69: I like the addition of "latest workspace state" but don't like the addition of ", and attempting" because that is the purpose of processing and not a separate operation. What do you think about fully duplicating the API error?

```suggestion
	ErrWorkspaceStillProcessing = errors.New("conflict\nLatest workspace state is being processed to discover resources, please try again later")
```

brandonc (approved):
Can you please mention in the CHANGELOG that we changed the message within ErrWorkspaceStillProcessing? This might be a string that folks depend on.

tfe.go
```diff
@@ -942,7 +942,7 @@ func checkResponseCode(r *http.Response) error {
 				return ErrWorkspaceLockedCannotDelete
 			}
 			if errorPayloadContains(errs, "being processed") {
-				return ErrWorkspaceStillProcessing
+				return ErrWorkspaceStillProcessing // as of 2024-03-04, the error message coming from atlas is "Latest workspace state is being processed to discover resources, please try again later"
```
sebasslash, line 945: This comment would be useful to include in the godocs; you can move this comment above the definition for `ErrWorkspaceStillProcessing`. I would also change `atlas` to `API` since that might lead to some confusion.

brandonc (approved):

PR #871, opened 2024-03-14
[TF-13396] Build new decoded error

laurenolivia:
## Description

When an HTTP response returns a 404, the details of the error are not logged. Currently, the error returns a string indicating the `resource is not found`. This PR builds an error that logs more info to help investigate what may be causing 404s.

## Output from tests

```

...
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -5,6 +5,9 @@ package tfe
 
 import (
 	"errors"
+	"log"
+	"net/http"
+	"strings"
 )
 
 // Generic errors applicable to all resources.
@@ -373,3 +376,16 @@ var (
 
 	ErrStateVersionUploadNotSupported = errors.New("upload not supported by this version of Terraform Enterprise")
 )
+
+type DecodeHTTPError struct {
+	Msg string
+	r   *http.Response
+}
+
+func (d *DecodeHTTPError) Error() string {
+	info, err := decodeErrorPayload(d.r)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return strings.Join(info, "\n")
+}
--- a/errors.go
+++ b/errors.go
@@ -5,9 +5,6 @@ package tfe
 
 import (
 	"errors"
-	"log"
-	"net/http"
-	"strings"
 )
 
 // Generic errors applicable to all resources.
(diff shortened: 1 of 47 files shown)


errors.go
```diff
@@ -373,3 +376,24 @@ var (
 
 	ErrStateVersionUploadNotSupported = errors.New("upload not supported by this version of Terraform Enterprise")
 )
+
+type DecodeHTTPError struct {
```
brandonc, line 380: This looks like a useful tool, but it doesn't help for types that are already response errors that can be decoded, like the ones returned by checkResponseCode.

You could add a function that can decode the contents of a variety of errors, starting with ErrResourceNotFound, that can be built with a response body and work in tandem with checkResponseCode.

It's important that the existing message does not change for ErrResourceNotFound. The additional decoded details should be available from a different field and not from the Error() method.

```
type ErrResourceNotFound {
  Details []string
}

func (ErrResourceNotFound) Error() string {
  return "resource not found" // <-- this is the important compatibility part
}

func NewErrorFromResponseCode(response *http.Response) error {
  decodedDetails, err := decodeErrorPayload(response)
  if err != nil {
    // There aren't any error details so this error isn't a 'jsonapi error'
    return fmt.Errorf("could not jsonapi error from this response: %w", err)
  }
  
  switch response.StatusCode {
  case 404:
    return ErrResourceNotFound{Details: decodedDetails}
  // etc
  }
  
  return nil
}
```

brandonc (changes requested):

errors.go
brandonc: Also don't forget unit tests

laurenolivia: Unit tests will need to be updated because the error will be wrapped with this solution.
Likely tests that check for equality:
```
assert.Equal(t, ErrResourceNotFound, err)
```

laurenolivia:
It’s been determined that this fix is a breaking change in go-tfe. At this time, we don’t have any plans to release V2. The original intent was to force more context in the output of a 404 due to a recurring flaky test in the provider. We’ve determined that any solution breaks go-tfe. Closing this PR for now.

PR #884, opened 2024-04-11
Make Organization Scoped a pointer

netramali:
## Description

Making organization scoped a pointer so that in provider when the api is not available, the value is shown as nil. Else, the true value would be confusing, if its is false or nil.

## Testing plan

Smoke test

## Output from tests

```
--- PASS: TestOAuthClientsUpdate (2.90s)
=== RUN   TestOAuthClientsUpdate/updates_organization_scoped
    --- PASS: TestOAuthClientsUpdate/updates_organization_scoped (0.71s)
PASS

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Enhancements
 * Adds Bitbucket Data Center as a new `ServiceProviderType` and ensures similar validation as Bitbucket Server by @zainq11 [#879](https://github.com/hashicorp/go-tfe/pull/879)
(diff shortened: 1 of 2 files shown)


brandonc (commented):
This is fine. Let's just add a note about the change in the CHANGELOG just in case. Note that we're changing a beta field type as a Bug Fix.

netramali:
> This is fine. Let's just add a note about the change in the CHANGELOG just in case. Note that we're changing a beta field type as a Bug Fix.

Added. Would it be possible for you to merge it since the tests on main branch are failing?

PR #893, opened 2024-04-23
Remove beta flags for description attribute for projects

netramali:
## Description

Since the project is now GA, and the provider changes have been merged, the beta flags can be removed.

## Testing plan

- ensure the integration tests pass

## Output from tests

```

...
```


roncodingenthusiast (approved):
✅

PR #902, opened 2024-05-21
feat: add workspace auto destroy duration field

notchairmk:
## Description

Adds an `auto_destroy_activity_duration` field to Workspaces (see [API docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#create-a-workspace:~:text=auto-destroy-activity-duration)). Makes use of `jsonapi.NullableAttr` because this is an optional string field, but we will want to be able to send an explicit null to unset.

## Testing plan

1. Run auto destroy tests, confirm that they're valid `envchain go-tfe go test -run "TestWorkspacesAutoDestroy.*"  -v ./...`

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./... -v -run "TestWorkspacesAutoDestroy.*"

?   	github.com/hashicorp/go-tfe/examples/backing_data	[no test files]
?   	github.com/hashicorp/go-tfe/examples/configuration_versions	[no test files]
?   	github.com/hashicorp/go-tfe/examples/registry_modules	[no test files]
?   	github.com/hashicorp/go-tfe/examples/run_errors	[no test files]
?   	github.com/hashicorp/go-tfe/examples/state_versions	[no test files]
?   	github.com/hashicorp/go-tfe/examples/users	[no test files]
?   	github.com/hashicorp/go-tfe/mocks	[no test files]
=== RUN   TestWorkspacesAutoDestroy
--- PASS: TestWorkspacesAutoDestroy (4.72s)
=== RUN   TestWorkspacesAutoDestroyDuration
--- PASS: TestWorkspacesAutoDestroyDuration (3.85s)
PASS
ok  	github.com/hashicorp/go-tfe	9.099s
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Unreleased
+## Enhancements
+* Adds the `AutoDestroyActivityDuration` field to `Workspace` by @notchairmk [#902](https://github.com/hashicorp/go-tfe/pull/902)
+
 ## Deprecations
 * The `IsSiteAdmin` field on User has been deprecated. Use the `IsAdmin` field instead [#900](https://github.com/hashicorp/go-tfe/pull/900)
-
 
 # v1.53.0
 
--- a/workspace.go
+++ b/workspace.go
@@ -155,44 +155,45 @@ type LockedByChoice struct {
 
 // Workspace represents a Terraform Enterprise workspace.
 type Workspace struct {
-	ID                         string                          `jsonapi:"primary,workspaces"`
-	Actions                    *WorkspaceActions               `jsonapi:"attr,actions"`
-	AllowDestroyPlan           bool                            `jsonapi:"attr,allow-destroy-plan"`
-	AssessmentsEnabled         bool                            `jsonapi:"attr,assessments-enabled"`
-	AutoApply                  bool                            `jsonapi:"attr,auto-apply"`
-	AutoApplyRunTrigger        bool                            `jsonapi:"attr,auto-apply-run-trigger"`
-	AutoDestroyAt              jsonapi.NullableAttr[time.Time] `jsonapi:"attr,auto-destroy-at,iso8601,omitempty"`
-	CanQueueDestroyPlan        bool                            `jsonapi:"attr,can-queue-destroy-plan"`
-	CreatedAt                  time.Time                       `jsonapi:"attr,created-at,iso8601"`
-	Description                string                          `jsonapi:"attr,description"`
-	Environment                string                          `jsonapi:"attr,environment"`
-	ExecutionMode              string                          `jsonapi:"attr,execution-mode"`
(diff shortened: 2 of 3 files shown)


workspace_integration_test.go
```diff
@@ -2747,7 +2748,7 @@ func TestWorkspacesAutoDestroy(t *testing.T) {
 	})
 	t.Cleanup(wCleanup)
 
-	require.Equal(t, wTest.AutoDestroyAt, autoDestroyAt)
+	require.Equal(t, autoDestroyAt, wTest.AutoDestroyAt)
```
notchairmk, line 2751: swapped the order on these since require wants `require.Equal(t, <expected>, <actual>)`

workspace.go
```diff
@@ -513,6 +517,9 @@ type WorkspaceUpdateOptions struct {
 	// Optional: The time after which an automatic destroy run will be queued
 	AutoDestroyAt jsonapi.NullableAttr[time.Time] `jsonapi:"attr,auto-destroy-at,iso8601,omitempty"`
 
+	// Optional: The period of time to wait after workspace activity to trigger a destroy run.
```
brandonc, line 520: Can we describe the format of this field? Is it a go-style duration string using the time unit suffix?

brandonc:
Is it the intention that a tfe_workspace without a auto_destroy_at config should unset any drifted autodestroy setting?

Usually this kind of drift correction is good but in the case of auto_destroy, it seems like terraform wouldn't be the best source of that config.

Methodology:
1. create workspace using basic config with no auto destroy
2. set up auto destroy in the UI
3. terraform apply

Outcome:
terraform plans to unset the auto destroy setting

Screenshot 2024-05-21 at 2 14 26 PM

notchairmk:
> Is it the intention that a tfe_workspace without a auto_destroy_at config should unset any drifted autodestroy setting?
> 
> Usually this kind of drift correction is good but in the case of auto_destroy, it seems like terraform wouldn't be the best source of that config.
> 
> Methodology:
> 
> 1. create workspace using basic config with no auto destroy
> 2. set up auto destroy in the UI
> 3. terraform apply
> 
> Outcome: terraform plans to unset the auto destroy setting
> 
> Screenshot 2024-05-21 at 2 14 26 PM

I think this is maybe not the right PR for this comment. Isn't this exactly the same behavior if you're setting another attribute on the workspace like `description`?

brandonc:
I totally put this in the wrong PR. Let me try again.

brandonc (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+## Enhancements
+* Adds the `AutoDestroyActivityDuration` field to `Workspace` by @notchairmk [#900](https://github.com/hashicorp/go-tfe/pull/900)
```
emailnitram, line 3: shouldn't this be 902?

PR #915, opened 2024-06-13
Expose is_unified field on unified resources

roncodingenthusiast:
## Description

- Expose is-unified field for unified resources between HCP and TFC 

## Testing plan (WIP)

- Integration tests

### DIFF
--- a/project.go
+++ b/project.go
@@ -46,8 +46,9 @@ type ProjectList struct {
 
 // Project represents a Terraform Enterprise project
 type Project struct {
-	ID   string `jsonapi:"primary,projects"`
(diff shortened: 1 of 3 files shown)


project.go
```diff
@@ -51,6 +51,9 @@ type Project struct {
 
 	Description string `jsonapi:"attr,description"`
 
+	// Optional: Indicates that the project is synced from HCP
```
brandonc, line 54: It's NBD, but none of these are an input attribute so no need to document as "Optional"

brandonc (dismissed):

ctrombley (approved):
👍

PR #922, opened 2024-06-25
Add AllowMemberTokenManagement to Team

juliannatetreault:
## Description

This PR adds a new team permission - `AllowMemberTokenManagement` - to the `Team` block. `AllowMemberTokenManagement` is  `true` by default and can only be managed by Owners or users with "Manage Teams" permissions. 

Additionally, this PR adds tests for the new permission.

## Testing plan

1. `AllowMemberTokenManagement` is  `true` by default when Teams are created.
2. `AllowMemberTokenManagement` is read for Teams.
3. `AllowMemberTokenManagement` is updatable for Teams.

## External links
- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/xxxx)

## Output from tests

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # UNRELEASED
+* Adds `AllowMemberTokenManagement` permission to `Team` by @juliannatetreault []()
 
 ## Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # UNRELEASED
-* Adds `AllowMemberTokenManagement` permission to `Team` by @juliannatetreault []()
+* Adds `AllowMemberTokenManagement` permission to `Team` by @juliannatetreault [#922](https://github.com/hashicorp/go-tfe/pull/922)
 
 ## Enhancements
 
--- a/team.go
+++ b/team.go
@@ -54,7 +54,8 @@ type Team struct {
 	Permissions        *TeamPermissions    `jsonapi:"attr,permissions"`
 	UserCount          int                 `jsonapi:"attr,users-count"`
 	SSOTeamID          string              `jsonapi:"attr,sso-team-id"`
-	AllowMemberTokenManagement bool        `jsonapi:"attr,allow-member-token-management"`
+	// AllowMemberTokenManagement is false for TFE versions older than v202408
+	AllowMemberTokenManagement bool `jsonapi:"attr,allow-member-token-management"`
 
 	// Relations
 	Users                   []*User                   `jsonapi:"relation,users"`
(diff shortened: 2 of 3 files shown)


team.go
```diff
@@ -54,6 +54,7 @@ type Team struct {
 	Permissions        *TeamPermissions    `jsonapi:"attr,permissions"`
 	UserCount          int                 `jsonapi:"attr,users-count"`
 	SSOTeamID          string              `jsonapi:"attr,sso-team-id"`
+	AllowMemberTokenManagement bool        `jsonapi:"attr,allow-member-token-management"`
```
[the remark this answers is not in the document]
juliannatetreault: This is a great idea; I'll update this! Thank you!

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # UNRELEASED
+* Adds `AllowMemberTokenManagement` permission to `Team` by @juliannatetreault []()
```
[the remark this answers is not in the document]
juliannatetreault: lol fair 😄

team.go
```diff
@@ -54,6 +54,8 @@ type Team struct {
 	Permissions        *TeamPermissions    `jsonapi:"attr,permissions"`
 	UserCount          int                 `jsonapi:"attr,users-count"`
 	SSOTeamID          string              `jsonapi:"attr,sso-team-id"`
+	// AllowMemberTokenManagement is false for TFE versions older than v202408
```
emlanctot, line 57: why is this the case?

juliannatetreault: This PR will go out before the release does, so we want to make it clear when this is available!

JarrettSpiker: When the API response is deserialized, for versions of TFE which dont know about the new feature `allow-member-token-management` will be absent. So when we initialize this object, `AllowMemberTokenManagement` wont be set, and `bool` s default to false

The comment is worthwhile because this is misleading! Checks for `team. AllowMemberTokenManagement == true` will fail, even though team members _can_ manage tokens.

We need to help users understand that this field is not useful pre TFE v202408

emlanctot: gotchyyaaa, i was confused because before this release all members should be able to create tokens and is true by default. but this makes sense for TFE though, i was just tripping myself up

team.go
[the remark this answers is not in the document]
emlanctot: wow missed this comment for my earlier question

JarrettSpiker (approved):

emlanctot (approved):

brandonc (approved):

emlanctot (approved):
💯 🛹 ✨

JarrettSpiker (approved):

PR #927, opened 2024-07-02
[WAYP-2819] Support creating no-code workspaces

paladin-devops:
The registry no-code module resource is updated with this commit to include an additional method which can be used to create workspaces using the registry no-code module.

## Testing plan

Run `go-tfe go test -run TestRegistryNoCodeModule -v ./...` with the necessary environment variables.

### DIFF
--- a/.github/actions/test-go-tfe/action.yml
+++ b/.github/actions/test-go-tfe/action.yml
@@ -95,6 +95,7 @@ runs:
         TFC_RUN_TASK_URL: "http://testing-mocks.tfe:22180/runtasks/pass"
         GITHUB_POLICY_SET_IDENTIFIER: "hashicorp/test-policy-set"
         GITHUB_REGISTRY_MODULE_IDENTIFIER: "hashicorp/terraform-random-module"
+        GITHUB_REGISTRY_NO_CODE_MODULE_IDENTIFIER: "hashicorp/terraform-random-no-code-module"
         OAUTH_CLIENT_GITHUB_TOKEN: "${{ inputs.oauth-client-github-token }}"
         GO111MODULE: "on"
         ENABLE_TFE: ${{ inputs.enterprise }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # UNRELEASED
 
-# v1.61.0
-
 ## Enhancements
 
 * Adds support for creating no-code workspaces by @paladin-devops [#927](https://github.com/hashicorp/go-tfe/pull/927)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # UNRELEASED
 
+# v1.61.0
+
+## Enhancements
+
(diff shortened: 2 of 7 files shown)


registry_no_code_module.go
```diff
@@ -33,6 +33,36 @@ type RegistryNoCodeModules interface {
...
+
+	AutoApply *bool `jsonapi:"attr,auto-apply,omitempty"`
+
+	// Project is the associated project with the workspace. If not provided,
+	// default project of the organization will be assigned to the workspace.
+	Project *Project `jsonapi:"relation,project,omitempty"`
+
+	// Variables is the slice of variables to be configured for the no-code
+	// workspace.
+	Variables []*Variable `jsonapi:"relation,vars,omitempty"`
```
sebasslash, line 61: As a convention, we don't express required fields as pointers or with `omitempty`.

registry_no_code_module.go
```diff
@@ -246,3 +301,14 @@ func (o *RegistryNoCodeModuleUpdateOptions) valid() error {
...
 	return nil
 }
+
+func (o *RegistryNoCodeModuleCreateWorkspaceOptions) valid() error {
+	if !validString(o.Name) {
+		return ErrRequiredName
+	}
+	if !validStringID(o.Name) {
+		return ErrInvalidName
+	}
```
sebasslash, line 311: validStringID ensures the field follows the resource external ID format used by our API (e.g for a workspace it would be `ws-1234abcde`).

So either this field accepts a workspace's external ID or a org/workspace name pair to identify a workspace, but not both. 

However since this creates a workspace using a no-code module, my guess would be to provide the name not an external ID? In which case we only need to call `validString(o.Name)`

sebasslash (commented):
:wave: I know this PR is still a draft, but figured I'd leave some general pointers below.

registry_no_code_module.go
paladin-devops: @sebasslash thanks for the feedback! I've addressed this and your other comment in [this commit](https://github.com/hashicorp/go-tfe/pull/927/commits/e8f7b3144ee48aabe3a327ed2c59e3972da73403).

.github/actions/test-go-tfe/action.yml
```diff
@@ -95,6 +95,7 @@ runs:
         TFC_RUN_TASK_URL: "http://testing-mocks.tfe:22180/runtasks/pass"
         GITHUB_POLICY_SET_IDENTIFIER: "hashicorp/test-policy-set"
         GITHUB_REGISTRY_MODULE_IDENTIFIER: "hashicorp/terraform-random-module"
+        GITHUB_REGISTRY_NO_CODE_MODULE_IDENTIFIER: "hashicorp/terraform-random-no-code-module"
```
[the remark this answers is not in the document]
paladin-devops: I added your team as admin!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # UNRELEASED
 
+# v1.61.0
```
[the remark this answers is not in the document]
paladin-devops: Gotcha - will fix this!

sebasslash (approved):

Issue #862, opened 2024-03-04
Support Run Tasks Integration API


karvounis-form3:
#### Use-cases

Users want to create custom TFE run tasks. However, it seems that the Golang structs, and functions that support this are missing from this repo. This is the API docs for [Run Tasks Integration API](https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration). This leads to users writing their own implementation, which can differ and lead to mistakes, bugs, and slower development cycle.

#### Attempted Solutions

It was attempted to write our own structs, functions etc but I believe the whole community would benefit if those were supported in this official Golang client.

#### Proposal

Support the following:
- [Run task request](https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration#run-task-request) struct. This way, the consumers of the call coming from TFC/E will not have to create their own struct
- [Run Task Callback](https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration#run-task-callback)
  - [Request body](https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration#request-body-1) struct
  - [Outcomes Payload Body](https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration#outcomes-payload-body) struct
  - Function that sends a PATCH request with the above payload back to the TFC/E callback URL

PR #929, opened 2024-07-03
feat: add support for run task results callback

sebasslash:
## Description

This PR aims to address the missing support for Run Tasks Integration API. I have already raised an issue https://github.com/hashicorp/go-tfe/issues/862 about this.
It introduces the following:
- RunTaskRequest struct which can be used to objectify the request you get from a TFC/E run task
- RunTasksCallback interface which can be used to update the run task with status and messages
  - TaskResultsCallbackOptions defines the attributes required in order to perform the update

## Testing plan

Unfortunately, it is not easy to test the Update function because you first need a running server which listens for incoming TFC/E requests. Then, you need a workspace run which is configured to use that run task endpoint.

However, I added the following unit tests and their output when run locally:

- TestTaskResultsCallbackRequestOptions_Validate

Screenshot 2024-06-19 at 8 27 28 AM

- TestTaskResultsCallbackRequestOptions_Marshal

Screenshot 2024-06-19 at 8 28 01 AM

- TestTaskResultsCallbackUpdate_Validate

Screenshot 2024-06-19 at 8 28 25 AM

- TestTaskResultsCallbackUpdate

Screenshot 2024-06-19 at 8 28 52 AM

## External links

https://github.com/hashicorp/terraform-run-task-scaffolding-go this HashiCorp repo contains part of the code that can be found here.


run_task_request.go
```diff
@@ -0,0 +1,54 @@
...
+// RunTaskRequestCapabilitites defines the capabilities that the caller supports.
+type RunTaskRequestCapabilitites struct {
+	Outcomes bool `json:"outcomes"`
+}
+
+// IsEndpointValidation returns true if this is a Request from TFC/E to validate and register this API endpoint.
+// Function copied from: https://github.com/hashicorp/terraform-run-task-scaffolding-go/blob/d7ed63b7d8eacf0897ab687d35d353386e4bd0ac/internal/sdk/api/structs.go#L55-L60
+func (r RunTaskRequest) IsEndpointValidation() bool {
+	return r.AccessToken == VerificationToken
+}
```
brandonc, line 54: This is testing against a const token that should never be valid. What is the point?

run_task_results_callback.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// Compile-time proof of interface implementation.
+var _ RunTasksCallback = (*taskResultsCallback)(nil)
+
+// RunTasksCallback describes all the Run Tasks Integration Callback API methods.
+//
+// TFE API docs:
+// https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration
+type RunTasksCallback interface {
```
brandonc, line 15: What do you think of calling this interface `RunTasksIntegration`? And changing the name of the method to `Callback`?

I feel that would better reflect the Documentation and the README.

run_task_results_callback.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+	// The PATCH request must use the token supplied in the originating request (access_token) for authentication.
+	// https://developer.hashicorp.com/terraform/enterprise/api-docs/run-tasks/run-tasks-integration#request-headers-1
+	req.Header.Set("Authorization", "Bearer "+accessToken)
+	return req.Do(ctx, nil)
+}
+
+func (o *TaskResultCallbackRequestOptions) valid() error {
+	if !validStringID(String(string(o.Status))) || (o.Status != TaskFailed && o.Status != TaskPassed && o.Status != TaskRunning) {
```
brandonc, line 75: I think there is a redundancy here that could be simplified to make the condition easier to read

```suggestion
	if o.Status != TaskFailed && o.Status != TaskPassed && o.Status != TaskRunning {
```

brandonc (commented):
This is a great addition. I just had some minor concern about naming

sebasslash:
@karvounis-form3 I'll address these suggestions on this branch that way we don't have to rebase your changes again.

Maed223 (approved):

sebasslash:
@karvounis-form3 We release roughly every week on Tuesdays, so you can expect these changes to land in `v1.59.0` next week.

karvounis-form3:
Thank you for this!

PR #941, opened 2024-07-29
[WAYP-2855] Enable execution mode option for creating no-code workspaces

paladin-devops:
## Description
This PR adds the option to set the execution mode of a Terraform no-code workspace when creating it.

## Testing plan
Run `go-tfe go test -run TestRegistryNoCodeModule -v ./...` with the necessary environment variables.

### DIFF
--- a/registry_no_code_module.go
+++ b/registry_no_code_module.go
@@ -43,7 +43,6 @@ type RegistryNoCodeModules interface {
 
 type RegistryNoCodeModuleCreateWorkspaceOptions struct {
 	Type string `jsonapi:"primary,no-code-module-workspace"`
-	// Add more create options here
 
 	// Name is the name of the workspace, which can only include letters,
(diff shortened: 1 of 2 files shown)


registry_no_code_module.go
```diff
@@ -68,6 +67,9 @@ type RegistryNoCodeModuleCreateWorkspaceOptions struct {
 
 	// SourceUrl is the URL of the source of the workspace.
 	SourceURL *string `jsonapi:"attr,source-url,omitempty"`
+
+	// ExecutionMode is the execution mode of the workspace.
+	ExecutionMode *string `jsonapi:"attr,execution-mode,omitempty"`
```
briancain, line 72: I think you'll also need to include `AgentPoolID` too, in case someone uses `agent` for its `ExecutionMode`?

briancain (approved):

PR #943, opened 2024-07-31
Add support for Org. Token types

glennsarti:
## Description

HCP Terraform organizations now support different types of tokens. This commit updates the org. auth. token client to support CRUD operations with a token type URL attribute.

Note that the integration tests require an organization that supports audit-logging

## Testing plan

Handled by integration tests

## External links

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/audit-trails-tokens)

## Output from tests

Handled by CI


organization_token.go
```diff
@@ -53,6 +67,35 @@ type OrganizationTokenCreateOptions struct {
 	// Optional: The token's expiration date.
 	// This feature is available in TFE release v202305-1 and later
 	ExpiredAt *time.Time `jsonapi:"attr,expired-at,iso8601,omitempty"`
+	// Optional: What type of token to create
+	// This option is only applicable to HCP Terraform and is ignored by TFE.
+	TokenType *TokenType
+}
+
+// OrganizationTokenDeleteOptions contains the options for deleting an organization token.
```
mrinalirao, line 75:
```suggestion
// OrganizationTokenReadOptions contains the options for reading an organization token.
```

glennsarti: Whoopsy

organization_token.go
```diff
@@ -13,6 +13,14 @@ import (
 // Compile-time proof of interface implementation.
 var _ OrganizationTokens = (*organizationTokens)(nil)
 
+type TokenType = string
```
mrinalirao, line 16: curious as to why you chose to go with type alias (`type TokenType = string`) vs  `type TokenType string`. The former allows you to pass any string type to a function that accepts TokenType as a function param, whereas the latter only allows to pass only `TokenType` type.  The latter is better for type safety

glennsarti: Huh whoops... Because it was typo.

I didn't expect that would compile, but TIL!  Thanks for sharp eyes @mrinalirao !

glennsarti: Fixed

organization_token.go
glennsarti: Fixed

mrinalirao (approved):
lgtm! 👍

mrinalirao (dismissed):

Maed223 (approved):

organization_token.go
```diff
@@ -28,8 +36,14 @@ type OrganizationTokens interface {
...
 	Read(ctx context.Context, organization string) (*OrganizationToken, error)
 
+	// Read an organization token with options.
+	ReadWithOptions(ctx context.Context, organization string, options OrganizationTokenReadOptions) (*OrganizationToken, error)
+
 	// Delete an organization token.
 	Delete(ctx context.Context, organization string) error
+
+	// Delete an organization token with options.
+	DeleteWithOptions(ctx context.Context, organization string, options OrganizationTokenDeleteOptions) error
```
Maed223, line 46: With adding these options, we should have some validations for them with clear errors to surface any invalid token types given.  [You can see here](https://github.com/hashicorp/go-tfe/blob/main/admin_setting_smtp.go) for a general example of how we typically do it.

glennsarti: Alrighty, so yes I could add some validation logic there but I'm having some thoughts about should I ...

Is there any danger in a user supplying their own token-type value? Not really, the server will catch it, and does return a useful HTTP error message. Also, it's a nice escape-hatch i.e. when HCP Terraform supports a new token-type, this library doesn't require an immediate update, instead users can supply the new token-type as a string instead of using our pre-defined constant list.

Which does raise a different question ... Is this a helper enumeration (i.e. "Here's a list of common values you can use!") or is it an explicit list that this project will enforce? (i.e. "Thou shalt only use these approved values"). Right now I'm leaning towards the former.

If we did want to go down the enforcement route, wouldn't a properly typed value be better instead of late validation? Or shouldn't validation occur when creating the options, not when they're used?

organization_token.go
```diff
@@ -53,6 +67,35 @@ type OrganizationTokenCreateOptions struct {
...
+// OrganizationTokenDeleteOptions contains the options for deleting an organization token.
+type OrganizationTokenDeleteOptions struct {
+	// Optional: What type of token to delete
+	// This option is only applicable to HCP Terraform and is ignored by TFE.
+	TokenType *TokenType
+}
+
+// Creates the URL for use against organization token API routes
+func (s *organizationTokens) generateURL(organization string, tokenType *TokenType) string {
+	u := fmt.Sprintf("organizations/%s/authentication-token", url.QueryEscape(organization))
```
brandonc, line 91: url.PathEscape

organization_token.go
```diff
@@ -53,6 +67,35 @@ type OrganizationTokenCreateOptions struct {
...
+	// Optional: What type of token to create
+	// This option is only applicable to HCP Terraform and is ignored by TFE.
+	TokenType *TokenType
+}
+
+// OrganizationTokenReadOptions contains the options for reading an organization token.
+type OrganizationTokenReadOptions struct {
+	// Optional: What type of token to read
+	// This option is only applicable to HCP Terraform and is ignored by TFE.
+	TokenType *TokenType
```
brandonc, line 79: If you annotate these with `url:"token,omitempty"` the underlying library will automatically encode the token type into the query string for each query (if it is not empty). This may be a better choice because it would make extending the parameters easier in the future. I don't think what you have is blocking, though.

brandonc (approved):

organization_token.go
glennsarti: So, I did look at that and unfortunately the annotations are ignored for DELETE and POST calls (tfe.go:273+ https://github.com/hashicorp/go-tfe/blob/fd766c164d112e5cca222d8412c58d83716ab8dc/tfe.go#L287)

Which is why I had to inject the query string (which I did not like doing :-( )

glennsarti: Would you prefer me to update that helper to honor the url annotations ?

It's basically moving that logic out of the "GET" case to before/after the switch

organization_token.go
glennsarti: I'm happy to change it (and it should be) but as an FYI, QueryEscape is littered throughout the codebase :-(

glennsarti:
Blocking merging pending comments from Brandon and I noticed the delete unit test is giving false positives 😢

brandonc:
@glennsarti Yuck! I didn't realize the query param handling was so inconsistent in NewRequest. I tried to make it more consistent in https://github.com/hashicorp/go-tfe/pull/959, as well as replacing all the QueryEscape usage with PathEscape

If you like it, we can get that merged and rebase this on top, utilizing `NewRequestWithAdditionalQueryParams` if that helps.

glennsarti:
@brandonc 100% I'll rebase and fixup once that's merged

organization_token.go
glennsarti: @Maed223 Did you have any thoughts or responses to this?

organization_token.go
glennsarti: Rebased on the new request helpers

organization_token.go
glennsarti: Fixed

glennsarti:
Rebased and modified for main. Back to review @brandonc

brandonc (approved):
Alright nice

glennsarti:
Rebased due to changelog merge conflict

organization_token.go
Maed223: In retrospect this may be a bit of a nit, and the example I provided happens to be a case where those potential values will be unlikely to change. Also with the value is optional, there's not much that could go wrong.

Maed223 (approved):
👍

PR #947, opened 2024-08-02
add new non-TBW timeout attributes

radditude:
## Description

Small cleanup PR to indicate that the `terraform-build-worker-plan-timeout` and `terraform-build-worker-apply-timeout` attributes in the admin API are being phased out and replaced with `plan-timeout` and `apply-timeout`, respectively. The new attributes are entirely interchangeable with the old, only the name is different.

### DIFF
--- a/admin_organization.go
+++ b/admin_organization.go
@@ -53,6 +53,8 @@ type AdminOrganization struct {
 	SsoEnabled                       bool   `jsonapi:"attr,sso-enabled"`
 	TerraformBuildWorkerApplyTimeout string `jsonapi:"attr,terraform-build-worker-apply-timeout"`
 	TerraformBuildWorkerPlanTimeout  string `jsonapi:"attr,terraform-build-worker-plan-timeout"`
+	ApplyTimeout                     string `jsonapi:"attr,apply-timeout"`
+	PlanTimeout                      string `jsonapi:"attr,plan-timeout"`
 	TerraformWorkerSudoEnabled       bool   `jsonapi:"attr,terraform-worker-sudo-enabled"`
 	WorkspaceLimit                   *int   `jsonapi:"attr,workspace-limit"`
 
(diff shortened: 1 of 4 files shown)


admin_organization.go
```diff
@@ -69,6 +71,8 @@ type AdminOrganizationUpdateOptions struct {
 	IsDisabled                       *bool   `jsonapi:"attr,is-disabled,omitempty"`
 	TerraformBuildWorkerApplyTimeout *string `jsonapi:"attr,terraform-build-worker-apply-timeout,omitempty"`
 	TerraformBuildWorkerPlanTimeout  *string `jsonapi:"attr,terraform-build-worker-plan-timeout,omitempty"`
+	ApplyTimeout                     *string `jsonapi:"attr,apply-timeout,omitempty"`
```
natalie-todd, line 74: Is this a phased approach where the `TerraformBuildWorkerApplyTimeout` and `TerraformBuildWorkerPlanTimeout` fields in this file and `admin_setting_general.go` be removed at a later date or will they be kept forever as deprecated values?

natalie-todd (approved):
I added a question for my own understanding of the plan here. Looks good to me though!

admin_organization.go
radditude: Good question! Per our [API stability policy](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/stability-policy#deprecation-policy), we have to keep deprecated endpoints around for at least 3 months before removing them. In practice, it's generally longer than that, unless there's a pressing reason to move quickly.

In this case, we have two bits of complicating info:
1. this is an admin-only API and therefore unavailable to users outside of TFE, and
2. this is a cleanup-only change - there's no urgent reason to retire the old attributes promptly

Given all that, while I hope we won't keep them _forever_ as deprecated attributes, I'm inclined to play it by ear instead of setting a strict removal schedule. I think we should be fine to remove the old ones after ~6 months, though.

PR #953, opened 2024-08-06
polling helpers for stack plan/configuration state

brandonc:
## Description

Stack configurations, stack plans, and stack plan operations are driven through their lifecycle through various states, and these can be repetitive and difficult to correctly handle when developing client applications. It would be really nice if the SDK provided some simple polling helpers to assist with building client applications.

## Testing plan

Utilized this in canary tests. DM for details. Here's an example:

Waiting for configuration prepared:
```go
waitForPrepared := client.StackConfigurations.AwaitPrepared(ctx, stack.LatestStackConfiguration.ID)
for {
    result := <-waitForPrepared
    if result.Error != nil {
        return fmt.Errorf("error waiting for status: %w", result.Error)
    }
    if result.Quit {
        logf("%s reached quit status %s", result.ID, result.Status)
        return nil
    }
    logf("%s had status %s", result.ID, result.Status)
}
```

Fan-in plans triggered by prepared configuration:
```go
plans, err := client.StackPlans.ListByConfiguration(ctx, stack.LatestStackConfiguration.ID, nil)
if err != nil {
    return fmt.Errorf("error listing stack plans: %w", err)
}

c.logger.Debug().Msgf("Found %d plans for stack %s", len(plans.Items), stack.ID)

// Wait on all the plans to finish
planWaits := make([]<-chan tfe.WaitForStatusResult, len(plans.Items))
for idx, plan := range plans.Items {
    planWaits[idx] = client.StackPlans.AwaitTerminalState(ctx, plan.ID)
    logf("Waiting for plan %s to finish...", plan.ID)
}

// Fan-in all waitings
planFinished := make(chan tfe.WaitForStatusResult)
for _, ch := range planWaits {
go func() {
    for result := range ch {
        planFinished <- result
    }
}()
}

done := 0

loop:
for {
    select {
    case <-ctx.Done():
        return fmt.Errorf("context was canceled: %w", ctx.Err())
    case progress := <-planFinished:
        if progress.Error != nil {
            return fmt.Errorf("plan %q could not finish: %w", progress.ID, progress.Error)
        }
	logf("Plan %s has status %s", progress.ID, progress.Status)
	if progress.Quit {
            done += 1
            logf("Plan %s reached quit status %s", progress.ID, progress.Status)
            if done == len(plans.Items) {
                break loop
            }
        } else if progress.Status == "paused_planned" {
            err = client.StackPlans.Approve(ctx, progress.ID)
            if err != nil {
                return fmt.Errorf("error approving plan %q: %w", progress.ID, err)
            }
        }
    }
}

```

### DIFF
--- a/stack_configuration.go
+++ b/stack_configuration.go
@@ -20,6 +20,33 @@ type StackConfigurations interface {
 
 	// JSONSchemas returns a byte slice of the JSON schema for the stack configuration.
 	JSONSchemas(ctx context.Context, stackConfigurationID string) ([]byte, error)
+
+	// AwaitCompleted generates a channel that will receive the status of the
+	// stack configuration as it progresses, until that status is "converged",
+	// "converging", "errored", "canceled".
+	AwaitCompleted(ctx context.Context, stackConfigurationID string) <-chan WaitForStatusResult
+
+	// AwaitPrepared generates a channel that will receive the status of the
+	// stack configuration as it progresses, until that status is "<status>",
+	// "errored", "canceled".
+	AwaitStatus(ctx context.Context, stackConfigurationID string, status StackConfigurationStatus) <-chan WaitForStatusResult
+}
+
+type StackConfigurationStatus string
+
+const (
+	StackConfigurationStatusPending    StackConfigurationStatus = "pending"
+	StackConfigurationStatusQueued     StackConfigurationStatus = "queued"
+	StackConfigurationStatusPreparing  StackConfigurationStatus = "preparing"
+	StackConfigurationStatusEnqueueing StackConfigurationStatus = "enqueueing"
+	StackConfigurationStatusConverged  StackConfigurationStatus = "converged"
+	StackConfigurationStatusConverging StackConfigurationStatus = "converging"
+	StackConfigurationStatusErrored    StackConfigurationStatus = "errored"
+	StackConfigurationStatusCanceled   StackConfigurationStatus = "canceled"
+)
+
+func (s StackConfigurationStatus) String() string {
+	return string(s)
 }
 
 type stackConfigurations struct {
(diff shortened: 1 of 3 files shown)


stack_configuration.go
```diff
@@ -20,6 +20,16 @@ type StackConfigurations interface {
 
 	// JSONSchemas returns a byte slice of the JSON schema for the stack configuration.
 	JSONSchemas(ctx context.Context, stackConfigurationID string) ([]byte, error)
+
```
DanielMSchmidt, line 23: Can we also add an AwaitQueued? This would be helpful since only then the eventstream url is set :)

stack_plan.go
```diff
@@ -32,6 +32,10 @@ type StackPlans interface {
 
 	// PlanDescription returns the plan description for a stack plan.
 	PlanDescription(ctx context.Context, stackPlanID string) (*JSONChangeDesc, error)
+
```
DanielMSchmidt, line 35: Can we add a AwaitQueued? Only then the eventstream URL is set and we need that for showing progress events.

DanielMSchmidt (commented):
Awesome work, there are two functions I would need to have added for the CLI if possible :)

stack_configuration.go
```diff
@@ -64,3 +74,33 @@ func (s stackConfigurations) JSONSchemas(ctx context.Context, stackConfiguration
 
 	return raw.Bytes(), nil
 }
+
+// AwaitPrepared generates a channel that will receive the status of the stack configuration as it progresses.
+// The channel will be closed when the stack configuration reaches a status indicating that or an error occurs. The
+// read will be retried dependending on the configuration of the client. When the channel is closed,
+// the last value will either be a terminal status or an error.
+func (s stackConfigurations) AwaitPrepared(ctx context.Context, stackConfigurationID string) <-chan WaitForStatusResult {
```
sebasslash, line 82: I'm curious what the advantage is to returning the read only channel as opposed to returning either the terminal status or error? The caller will inevitably have to write a range loop to read from it or am I missing some extra context here

sebasslash: Oh wait I see the examples, it allows for easier handling within a `select`.

stack_configuration.go
brandonc: I added `AwaitStatus` that accepts a specific status, and renamed "AwaitPrepared" to "AwaitCompleted"

stack_plan.go
brandonc: I don't think queued is a state for stack plan. Do you mean running_queued?

DanielMSchmidt: I'm not too sure what I meant anymore, just ignore this comment 🙈

DanielMSchmidt (approved):

sebasslash (approved):

PR #954, opened 2024-08-07
Fix returned workspace for created & upgraded no-code workspaces

paladin-devops:
## Description

This change is being made because the create and upgrade methods for no-code workspaces _do not_ properly return the workspace. They _do_ correctly create and upgrade no-code workspaces, respectively; however the workspace they return is a zero-value `tfe.Workspace`.

This PR fixes this issue by removing the `RegistryNoCodeModuleWorkspace` struct, which was a struct that only "wrapped" an embedded `tfe.Workspace` struct field. This was done with the intent of future-proofing the endpoint in the event that other data was returned. However, the documented API response clearly is only a `tfe.Workspace`, and so this PR corrects the method to return that value.

The integration tests passed prior to this PR because the tests only tested the "refreshed" workspace, meaning that the endpoint to fetch the workspace after it was created was used to check that the desired input workspace was created as expected. The test should have used the workspace returned by the "post" endpoints themselves.

## Testing plan

```shell
$ go test -run TestRegistryNoCodeModulesCreateWorkspace -v ./...

$ go test -run TestRegistryNoCodeModulesUpgradeWorkspace -v ./...
```

## External links

Relevant API endpoints:
1. https://developer.hashicorp.com/terraform/cloud-docs/api-docs/no-code-provisioning#create-a-no-code-module-workspace
2. https://developer.hashicorp.com/terraform/cloud-docs/api-docs/no-code-provisioning#initiate-a-no-code-workspace-update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # UNRELEASED
 * Adds `AllowMemberTokenManagement` permission to `Team` by @juliannatetreault [#922](https://github.com/hashicorp/go-tfe/pull/922)
 
+## Bug Fixes
+* Fix returned workspace for no-code workspace creation & upgrade by @paladin-devops [#954](https://github.com/hashicorp/go-tfe/pull/954)
+
 # v1.61.0
 
 ## Enhancements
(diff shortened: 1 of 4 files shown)


brandonc (approved):
I can see this is a bug fix so it's not exactly a "Breaking Change" but hopefully won't cause too much disruption downstream. I may highlight the issue a little brighter in the CHANGELOG before release

PR #956, opened 2024-08-08
Stacks: Decode diagnostics

DanielMSchmidt:
## Description

Decode diagnostics for stack_configuration and stack_plan_operation

## Output from tests

```

...
```


brandonc (approved):
Would be nice to have some tests (marked SkipUnlessBeta) for diagnostics decoding but I also recognize there are a lot of tests we need to backfill as we get more setup automation in place

PR #959, opened 2024-08-13
More flexible parameter handling in NewRequest and friends

brandonc:
@glennsarti noted that go-tfe demonstrated some odd query parameter encoding behavior depending on http method, and that url.QueryEncode was used throughout the code when we actually should have used the slightly more narrow url.PathEncode.


glennsarti:
the tfe.go changes look good. Just need to appease the almighty lint deity!

glennsarti (approved):

glennsarti:
Ship it

PR #973, opened 2024-09-13
Add billable RUM count to state version

shoekstra:
## Description

For our internal accounting we want to track billable resources at workspace level and saw some discrepancies when comparing the usage API value with how we counted billable resources. Via a support ticket we were notified about the `billable-rum-count` attribute which is working for us and would like to continue using 

## Testing plan

This value should always be there now (assumption) so should pop up in the usual integration tests.

## Output from tests

No TFE instance, but in another project where I use this package I'm now able to sum the billable resources across workspaces using this attribute and the total matches the billable resources in the Usage page.

### DIFF
--- a/state_version.go
+++ b/state_version.go
@@ -89,16 +89,17 @@ type StateVersionList struct {
 
 // StateVersion represents a Terraform Enterprise state version.
 type StateVersion struct {
-	ID              string             `jsonapi:"primary,state-versions"`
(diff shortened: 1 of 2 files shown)


brandonc:
@shoekstra Thanks for contributing this. I revised this a little bit and re-opened as #974. I changed the type to a pointer because there are several situations where the RUM count attribute would be missing from the response-- like not being authorized to read it. I wouldn't want that situation to be ambiguous with zero RUM. You're still included as an author on the commit.

PR #978, opened 2024-09-18
Adds ErrWorkspaceLockedStateVersionStillPending

brandonc:
## Description

Adds `ErrWorkspaceLockedStateVersionStillPending`, indicating that a workspace can't be unlocked yet because the latest state version upload is still pending. This will help signal to terraform that it should retry this operation.


Maed223 (approved):
🤘

PR #979, opened 2024-09-23
Add no-code endpoint for fetching registry module variables.

paladin-devops:
## Description

This PR adds support to go-tfe for the HCP Terraform endpoint to retrieve a no-code module's variables.

## Testing plan

1. Create a no-code module with at least one required input variable.
2. Get the ID of the no-code module.
    - expected format: `nocode-abcdefgh12345678`
3. Set the environment variable `NO_CODE_MODULE_ID` to be that ID.
4. Run `go test -run TestRegistryNoCodeModule -v ./...`, with the other necessary testing environment variables.

## Test Output

```terminal
=== RUN   TestRegistryNoCodeModulesReadVariables
--- PASS: TestRegistryNoCodeModulesReadVariables (7.94s)
=== RUN   TestRegistryNoCodeModulesReadVariables/happy_path
    --- PASS: TestRegistryNoCodeModulesReadVariables/happy_path (7.66s)
PASS
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Fixes a bug in `NewRequest` that did not allow query parameters to be specified in the first parameter, which broke several methods: `RegistryModules ReadVersion`, `VariableSets UpdateWorkspaces`, and `Workspaces Readme` by @brandonc [#982](https://github.com/hashicorp/go-tfe/pull/982)
 
+## Enhancements
+
+* Add support for reading a no-code module's variables by @paladin-devops [#979](https://github.com/hashicorp/go-tfe/pull/979)
+
 # v1.67.0
 
 ## Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
+## Enhancements
+
+* Add support for reading a no-code module's variables by @paladin-devops [#979](https://github.com/hashicorp/go-tfe/pull/979)
+
 # v1.67.1
 
 ## Bug Fixes
 
 * Fixes a bug in `NewRequest` that did not allow query parameters to be specified in the first parameter, which broke several methods: `RegistryModules ReadVersion`, `VariableSets UpdateWorkspaces`, and `Workspaces Readme` by @brandonc [#982](https://github.com/hashicorp/go-tfe/pull/982)
-
-## Enhancements
-
-* Add support for reading a no-code module's variables by @paladin-devops [#979](https://github.com/hashicorp/go-tfe/pull/979)
 
 # v1.67.0
 
--- a/registry_no_code_module.go
+++ b/registry_no_code_module.go
@@ -26,6 +26,10 @@ type RegistryNoCodeModules interface {
 	// **Note: This API is still in BETA and subject to change.**
 	Read(ctx context.Context, noCodeModuleID string, options *RegistryNoCodeModuleReadOptions) (*RegistryNoCodeModule, error)
 
+	// ReadVariables returns the variables for a version of a no-code module
+	// **Note: This API is still in BETA and subject to change.**
+	ReadVariables(ctx context.Context, noCodeModuleID, noCodeModuleVersion string, options *RegistryNoCodeModuleReadVariablesOptions) (*RegistryModuleVariableList, error)
+
 	// Update a registry no-code module
 	// **Note: This API is still in BETA and subject to change.**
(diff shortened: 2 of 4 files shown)


registry_no_code_module.go
```diff
@@ -41,6 +45,44 @@ type RegistryNoCodeModules interface {
...
 }
 
+// RegistryModuleVariableList is a list of registry module variables.
+type RegistryModuleVariableList struct {
+	Items []*RegistryModuleVariable
+
+	// NOTE: At the time of authoring this comment, the API endpoint to fetch
+	// registry module variables does not support pagination. This field is
+	// included to satisfy jsonapi unmarshaler implementation here:
+	// https://github.com/hashicorp/go-tfe/blob/3d29602707fa4b10469d1a02685644bd159d3ccc/tfe.go#L859
```
paladin-devops, line 55: I think this could be a bug in the unmarshaler implementation. In the case of the change in this PR, `Items` is a slice, and in the `unmarshalResponse` function, it should be handled by `jsonapi.UnmarshalManyPayload`. However, without the `*Pagination` field of this struct present (which should not actually be here, because the backend service doesn't actually return a `*Pagination) in the response), the conditional [here](ttps://github.com/hashicorp/go-tfe/blob/3d29602707fa4b10469d1a02685644bd159d3ccc/tfe.go#L859) sends the request into the `jsonapi` handler `UnmarshalPayload`, which is documented to support "single request payloads only", leading an attempt to decode a slice to fail.

For now, just including the `*Pagination` field here is a workaround. But if this issue is fixed, then it can be removed (or if pagination is implemented on the backend, it should be kept).

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 * Fixes a bug in `NewRequest` that did not allow query parameters to be specified in the first parameter, which broke several methods: `RegistryModules ReadVersion`, `VariableSets UpdateWorkspaces`, and `Workspaces Readme` by @brandonc [#982](https://github.com/hashicorp/go-tfe/pull/982)
 
+## Enhancements
+
+* Add support for reading a no-code module's variables by @paladin-devops [#979](https://github.com/hashicorp/go-tfe/pull/979)
+
```
sebasslash, line 10: You'll need to move this section above `# v1.67.1`

registry_no_code_module.go
sebasslash: This is a great callout, I'll create a ticket to look into this. :+1:

registry_no_code_module_integration_test.go
```diff
@@ -211,6 +211,34 @@ func TestRegistryNoCodeModulesRead(t *testing.T) {
...
+//
+// This test requires that the environment variable "NO_CODE_MODULE_ID" is set
+// with the ID of an existing no-code module that has variables.
+func TestRegistryNoCodeModulesReadVariables(t *testing.T) {
+	skipUnlessBeta(t)
+	client := testClient(t)
+	ctx := context.Background()
+	r := require.New(t)
+
+	ncmID := os.Getenv("NO_CODE_MODULE_ID")
```
sebasslash, line 225: Making a note of this to add the env to our CI workflow. Would it be possible to fetch the ID using [GITHUB_REGISTRY_NO_CODE_MODULE_IDENTIFIER](https://github.com/hashicorp/go-tfe/blob/0b3525a3372b37b0bada8720941bb8972e4a338d/.github/actions/test-go-tfe/action.yml#L98) ?

registry_no_code_module.go
```diff
@@ -41,6 +45,44 @@ type RegistryNoCodeModules interface {
 	UpgradeWorkspace(ctx context.Context, noCodeModuleID string, workspaceID string, options *RegistryNoCodeModuleUpgradeWorkspaceOptions) (*WorkspaceUpgrade, error)
 }
 
+// RegistryModuleVariableList is a list of registry module variables.
```
sebasslash, line 48: Let's add a BETA label here too given we might yank `*Pagination` from this.

registry_no_code_module.go
```diff
@@ -160,6 +202,14 @@ type RegistryNoCodeModuleReadOptions struct {
 	Include []RegistryNoCodeModuleIncludeOpt `url:"include,omitempty"`
 }
 
+type RegistryNoCodeModuleReadVariablesOptions struct {
```
sebasslash, line 205: Let's add a GODOC comment here as well

sebasslash (changes requested):
One minor change to the changelog, a few extra comments and you're good to go :)

sebasslash (approved):

PR #982, opened 2024-09-25
fixes NewRequest strips query from path parameter

### LINKED ISSUE #981: Beginning with 1.63.0, NewRequest strips query params, breaking some callers
This commit seems to be the culprit https://github.com/hashicorp/go-tfe/commit/417505b3914ee4e34a17d77b609b7f487bf4ceed#diff-2aefd6ea323a50ad81560708158f18e31d0d177f50aed3edd5a94353b7ab0f5aR292

brandonc:
Fixes a bug in `NewRequest` that did not allow query parameters to be specified in the first parameter, which broke several methods: `RegistryModules ReadVersion`, `VariableSets UpdateWorkspaces`, and `Workspaces Readme`

Closes #981

### DIFF
--- a/tfe.go
+++ b/tfe.go
@@ -277,6 +277,8 @@ func (c *Client) NewRequestWithAdditionalQueryParams(method, path string, reqBod
 		}
 	}
 
+	// Will contain combined query values from path parsing and
+	// additionalQueryParams parameter
 	q := make(url.Values)
 
 	// Create a request specific headers map.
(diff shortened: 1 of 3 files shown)


tfe.go
```diff
@@ -310,6 +312,9 @@ func (c *Client) NewRequestWithAdditionalQueryParams(method, path string, reqBod
 		body = reqBody
 	}
 
+	for k, v := range u.Query() {
+		q[k] = v
+	}
```
sebasslash, line 317: :heart:

sebasslash (approved):

PR #987, opened 2024-10-16
Filtering by and updating tag-bindings

brandonc:
- Filter workspaces and projects by kv tags
- Create workspaces and projects with kv tags
- Replace kv tags on workspace and projects


workspace.go
```diff
@@ -610,6 +621,10 @@ type WorkspaceUpdateOptions struct {
 	// Associated Project with the workspace. If not provided, default project
 	// of the organization will be assigned to the workspace
 	Project *Project `jsonapi:"relation,project,omitempty"`
+
+	// Associated TagBindings of the project. Note that this will replace
```
ctrombley, line 625: nit: consider "Note that _populating_ this _field_ will..."?

workspace.go
```diff
@@ -715,6 +736,30 @@ func (s *workspaces) List(ctx context.Context, organization string, options *Wor
...
+	}
+
+	u := fmt.Sprintf("workspaces/%s/tag-bindings", url.PathEscape(workspaceID))
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var list struct {
+		*Pagination
```
ctrombley, line 751: Same as above, I don't think we are paginating the api response.

workspace.go
```diff
@@ -1436,7 +1481,7 @@ func (o WorkspaceCreateOptions) valid() error {
 	if o.AgentPoolID == nil && (o.ExecutionMode != nil && *o.ExecutionMode == "agent") {
 		return ErrRequiredAgentPoolID
 	}
-	if o.TriggerPrefixes != nil && len(o.TriggerPrefixes) > 0 &&
+	if len(o.TriggerPrefixes) > 0 &&
```
ctrombley, line 1484: This change had me do a double take.  I didn't realize nil slices allowed calls to `len`!
You can't do that with other nils. TIL.

ctrombley (commented):
This looks great, just a few minor comments but no blockers. 

Do we want to test around including both kinds of filters in the same request? Or enforce mutually exclusivity?

ctrombley (approved):

project.go
```diff
@@ -166,6 +185,30 @@ func (s *projects) Read(ctx context.Context, projectID string) (*Project, error)
...
+	}
+
+	u := fmt.Sprintf("projects/%s/tag-bindings", url.PathEscape(projectID))
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var list struct {
+		*Pagination
```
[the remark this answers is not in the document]
brandonc: I should have highlighted this in the notes. Proper decoding inside go-tfe depends on the collection being this type, but there is no pagination. That's why I kind of hid it inside this anonymous type and didn't return a List type with pagination.

PR #988, opened 2024-10-17
Enable alternative reading of a registry module by its ID

dsa0x:
## Description

This enables an alternative way to read a registry module by its ID instead of its address (namespace/name/system). When the external ID is given, all other attributes are ignored.

## Testing plan

- Create a token in your HCP TF instance
- Add the TFE_ADDRESS and TFE_TOKEN variables to your environment
- Running `go test -v -run ^TestRegistryModulesRead$ github.com/hashicorp/go-tfe` should succeed.

## External links

- [JIRA](https://hashicorp.atlassian.net/browse/TFECO-6399)

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test -v -run ^TestRegistryModulesRead$ github.com/hashicorp/go-tfe
=== RUN   TestRegistryModulesRead/with_a_external_ID_field_for_private_module
=== RUN   TestRegistryModulesRead/with_a_external_ID_field_for_private_module/permissions_are_properly_decoded
=== RUN   TestRegistryModulesRead/with_a_external_ID_field_for_private_module/timestamps_are_properly_decoded
...
    --- PASS: TestRegistryModulesRead/with_a_external_ID_field_for_private_module (0.21s)
        --- PASS: TestRegistryModulesRead/with_a_external_ID_field_for_private_module/permissions_are_properly_decoded (0.00s)
        --- PASS: TestRegistryModulesRead/with_a_external_ID_field_for_private_module/timestamps_are_properly_decoded (0.00s)
...
PASS
ok      github.com/hashicorp/go-tfe     3.926s

...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -530,7 +530,6 @@ func (r *registryModules) Read(ctx context.Context, moduleID RegistryModuleID) (
 	if moduleID.ExternalID != "" {
 		u = fmt.Sprintf("registry-modules/%s", url.PathEscape(moduleID.ExternalID))
 	} else {
-
 		if moduleID.RegistryName == "" {
 			log.Println("[WARN] Support for using the RegistryModuleID without RegistryName is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the RegistryName in RegistryModuleID.")
 			moduleID.RegistryName = PrivateRegistry
--- a/registry_module.go
+++ b/registry_module.go
@@ -111,8 +111,8 @@ const (
 // Use NewPublicRegistryModuleID or NewPrivateRegistryModuleID to build one
 
 type RegistryModuleID struct {
-	// The unique external ID of the organization. If given, the other fields are ignored.
-	ExternalID string
+	// The unique ID of the module. If given, the other fields are ignored.
+	ID string
 	// The organization the module belongs to, see RegistryModule.Organization.Name
 	Organization string
 	// The name of the module, see RegistryModule.Name
@@ -527,7 +527,7 @@ func (r *registryModules) Read(ctx context.Context, moduleID RegistryModuleID) (
 	}
 
 	var u string
-	if moduleID.ExternalID == "" {
+	if moduleID.ID == "" {
 		if moduleID.RegistryName == "" {
 			log.Println("[WARN] Support for using the RegistryModuleID without RegistryName is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the RegistryName in RegistryModuleID.")
 			moduleID.RegistryName = PrivateRegistry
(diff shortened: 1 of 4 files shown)


tfe.go
```diff
@@ -319,7 +319,7 @@ func (c *Client) NewRequestWithAdditionalQueryParams(method, path string, reqBod
 		q[k] = v
 	}
 
-	u.RawQuery = encodeQueryParams(q)
+	u.RawQuery = strings.Join([]string{u.RawQuery, encodeQueryParams(q)}, "&")
```
dsa0x, line 322: This fixes a bug where the original set query param is overwritten.

tfe.go
```diff
@@ -319,7 +319,12 @@ func (c *Client) NewRequestWithAdditionalQueryParams(method, path string, reqBod
 		q[k] = v
 	}
 
-	u.RawQuery = encodeQueryParams(q)
+	encodedParams := encodeQueryParams(q)
+	if u.RawQuery == "" {
+		u.RawQuery = encodedParams
+	} else if encodedParams != "" {
+		u.RawQuery = strings.Join([]string{u.RawQuery, encodedParams}, "&")
+	}
```
dsa0x, line 327: This fixes a bug where the original set query param is overwritten.

registry_module.go
```diff
@@ -523,24 +526,29 @@ func (r *registryModules) Read(ctx context.Context, moduleID RegistryModuleID) (
...
 	}
 
-	if moduleID.RegistryName == "" {
-		log.Println("[WARN] Support for using the RegistryModuleID without RegistryName is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the RegistryName in RegistryModuleID.")
-		moduleID.RegistryName = PrivateRegistry
-	}
+	var u string
+	if moduleID.ID == "" {
+		if moduleID.RegistryName == "" {
+			log.Println("[WARN] Support for using the RegistryModuleID without RegistryName is deprecated as of release 1.5.0 and may be removed in a future version. The preferred method is to include the RegistryName in RegistryModuleID.")
```
brandonc, line 532: I know you didn't add this, but registry_module.go seems to be the only file in go-tfe which is using the `log`  package, and, as a library, probably shouldn't be logging to stderr. It would be nice if you could go ahead and remove all these log.Print calls from this file, but I'll take care of it if you don't feel like waiting for another review!

brandonc (approved):

registry_module.go
dsa0x: Thank you. I will go ahead and merge this. Is there a convention employed in go-tfe for including deprecated messages?. Perhaps we could just remove them entirely, and have the comments indicate the deprecated attributes.

PR #1001, opened 2024-11-11
Update registry module object

kd-hcp:
## Description

Updated the RegistryModule structure in registry_module.go to include the "relation" NoCodeModules if it included in the returned json. Omitempty was included to ignore this in the event the json return does not include the no-code-module relationship.

## Testing plan

Make a call to RegistryModules.List(). 
The returned object will now include the param NoCodeModules which is an array object containing any associated no code module data related to the parent module.

## External links

- [API documentation](https://developer.hashicorp.com/terraform/enterprise/api-docs/private-registry/modules)

## Output from tests

```

=== RUN   TestRegistryModulesListCommit
    registry_module_integration_test.go:681: Skipping test related to a HCP Terraform beta feature. Set ENABLE_BETA=1 to run.
--- SKIP: TestRegistryModulesListCommit (0.00s)
PASS
ok      github.com/hashicorp/go-tfe     0.287s

...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -156,7 +156,8 @@ type RegistryModule struct {
 	UpdatedAt           string                          `jsonapi:"attr,updated-at"`
 
 	// Relations
-	Organization *Organization `jsonapi:"relation,organization"`
+	Organization  *Organization           `jsonapi:"relation,organization"`
+	NoCodeModules []*RegistryNoCodeModule `jsonapi:"relation,no-code-modules"`
 }
 
 // Commit represents a commit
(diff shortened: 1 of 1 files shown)


brandonc:
I think this is being added in https://github.com/hashicorp/go-tfe/pull/1003

kd-hcp:
> I think this is being added in #1003

so are you not taking my change because someone made the change after me?

brandonc:
Right, I'm saying this duplicates a line of code in a PR that was already merged

PR #1007, opened 2024-11-15
Adds dd test optimization integration

brandonc:
See comment below 🤞👇  ¯\_(ツ)_/¯ 

The report below will outline tests coverage % decreases, flaky tests, etc whenever they apply to the PR. Similar to this:

image (2)

### DIFF
--- a/.github/actions/test-go-tfe/action.yml
+++ b/.github/actions/test-go-tfe/action.yml
@@ -100,11 +100,13 @@ runs:
         GO111MODULE: "on"
         ENABLE_TFE: ${{ inputs.enterprise }}
       run: |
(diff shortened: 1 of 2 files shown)


barrettclark:
Should the `cover.out` file be included? Will it be updated as code and tests evolve?

brandonc:
No, it's transient and used to report the number to datadog

barrettclark:
That's what I thought. It looks like you included it in this PR, though. If it shouldn't be then it could also be helpful to include that in the `.gitignore` file.

brandonc:
OOPS!

barrettclark (approved):

Maed223 (approved):

PR #1008, opened 2024-11-18
CODEOWNERS needs to have a default owner

barrettclark:
CODEOWNERS needs to have a default owner, so I copied the settings from terraform-provider-tfe.

https://hashicorp.atlassian.net/browse/TF-21760


sebasslash (approved):
:rocket: :cloud:

PR #1017, opened 2024-12-05
feat(stacks): Add trigger disabled attribute to StackVCSRepo

mjyocca:
## Description

Allows Stacks public beta customers to enable/disable webhook processing. 

## Output from tests

```

...
```


matejrisek (dismissed):

stack.go
```diff
@@ -71,6 +71,7 @@ type StackVCSRepo struct {
 	Branch            string `jsonapi:"attr,branch,omitempty"`
 	GHAInstallationID string `jsonapi:"attr,github-app-installation-id,omitempty"`
 	OAuthTokenID      string `jsonapi:"attr,oauth-token-id,omitempty"`
+	TriggerDisabled   bool   `jsonapi:"attr,trigger-disabled"`
```
ctrombley, line 74: Would you mind adding an assertion (maybe to `TestStackReadUpdateDelete`) that validates this is being unmarshalled properly?

ctrombley (changes requested):

brandonc:
If this is still needed please modify the OpenAPI spec in the HCP Terraform platform code.

PR #1018, opened 2024-12-05
Add support for the Explorer API

sebasslash:
## Description

Adds support for the two current GA endpoints of the Explorer API:

* [Execute a Query
](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/explorer#execute-a-query) 
* [Export Data as CSV](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/explorer#export-data-as-csv)

## Testing plan

To run the tests, you will need an organization available with enough activity to test several different types of queries. The organization must be specified using the `EXPLORER_TEST_ORGANIZATION` environment variable; otherwise, the tests will be skipped. Test assertions rely on the current state of the specified organization, and failures may occur if the organization does not meet the required query conditions enumerated in the integration tests. For guidance on writing custom queries to smoke test on your own, refer to the code examples provided in the `./examples/explorer` directory.

## Output from tests

```

...
```


github_app_installation.go
```diff
@@ -51,7 +51,6 @@ type GHAInstallationListOptions struct {
 func (s *gHAInstallations) List(ctx context.Context, options *GHAInstallationListOptions) (*GHAInstallationList, error) {
 	u := "github-app/installations"
 	req, err := s.client.NewRequest("GET", u, options)
-	fmt.Println(u)
```
ctrombley, line 54: ❤️

ctrombley (dismissed):
Smoke tested, looks great! Thanks for doing this! ✨

notchairmk:
This is cool, but might lock us into a path that we don't want to go down quite yet. One thing that we want to get to is an explorer discovery endpoint that could be used to look up type and field information, so that we don't have to hardcode all of the types and supported fields here and can instead read them from the backend and get some metadata around supported data types and operations.

There's a decent amount that we could use in the future, but I'd like to hold off on this for the moment just so we don't have to undo things in the near future. Thank you for taking a first swipe at this!

PR #1022, opened 2024-12-20
[TF-18527] Add Archs to AdminTerraformVersionCreateOptions

natalie-todd:
## Description

The purpose of this PR is to update the `AdminTerraformVersionCreateOptions` struct with an `Archs` field so [Terraform Releases](https://github.com/hashicorp/terraform-releases) can add Terraform versions with Linux amd64 and arm64 architectures. 

## External links

- [API documentation](https://hermes.hashicorp.services/document/133PNB2FQKSFKtLa6BmEk9T2NHAwiBKiVLmKSxHfk3Ic)
- [Related PR](https://github.com/hashicorp/terraform-releases/pull/150)
- [Jira](https://hashicorp.atlassian.net/browse/TF-18527?atlOrigin=eyJpIjoiOGJiMDEyMDA5OThiNGMxNmFlMGE0MWY0ODk0YWI4Y2QiLCJwIjoiaiJ9)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Enhancements
+
+* Add `Archs` field to `AdminTerraformVersionCreateOptions` by @natalie-todd [#1022](https://github.com/hashicorp/go-tfe/pull/1022)
+
 # v1.71.0
 
 ## Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## Enhancements
 
 * Add support for project level auto destroy settings @simonxmh [#1011](https://github.com/hashicorp/go-tfe/pull/1011)
-* Add `Archs` field to `AdminTerraformVersionCreateOptions` by @natalie-todd [#1022](https://github.com/hashicorp/go-tfe/pull/1022)
-
+* Add BETA support for Linux arm64 agents, which is EXPERIMENTAL, SUBJECT TO CHANGE, and may not be available to all users @natalie-todd [#1022](https://github.com/hashicorp/go-tfe/pull/1022)
+  
 # v1.71.0
 
 ## Enhancements
--- a/admin_terraform_version.go
+++ b/admin_terraform_version.go
@@ -14,6 +14,12 @@ import (
 // Compile-time proof of interface implementation.
 var _ AdminTerraformVersions = (*adminTerraformVersions)(nil)
 
+const (
+	linux = "linux"
+	amd64 = "amd64"
+	arm64 = "arm64"
+)
+
 // AdminTerraformVersions describes all the admin terraform versions related methods that
 // the Terraform Enterprise API supports.
 // Note that admin terraform versions are only available in Terraform Enterprise.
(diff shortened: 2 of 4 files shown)


admin_terraform_version.go
```diff
@@ -70,29 +79,31 @@ type AdminTerraformVersionsListOptions struct {
...
+	Type             string                     `jsonapi:"primary,terraform-versions"`
+	Version          *string                    `jsonapi:"attr,version,omitempty"`
+	URL              *string                    `jsonapi:"attr,url,omitempty"`
+	Sha              *string                    `jsonapi:"attr,sha,omitempty"`
+	Official         *bool                      `jsonapi:"attr,official,omitempty"`
+	Deprecated       *bool                      `jsonapi:"attr,deprecated,omitempty"`
+	DeprecatedReason *string                    `jsonapi:"attr,deprecated-reason,omitempty"`
+	Enabled          *bool                      `jsonapi:"attr,enabled,omitempty"`
+	Beta             *bool                      `jsonapi:"attr,beta,omitempty"`
+	Archs            []*ToolVersionArchitecture `jsonapi:"attr,archs"`
```
sebasslash, line 106: Just clarifying on the omission of `omitempty` in the struct tag: is this field required in the request body? If not, we will end up serializing an empty array potentially overriding any existing architectures persisted in the API. Normally for optional fields, we use `omitempty` to prevent this from occurring. 

Furthermore, if we do want to allow deleting all architectures for a given TF version then we will need to define a separate method that intentionally allows an empty slice to be sent over the wire, i.e does not include `omitempty`. See #1023 as an example.

sebasslash (changes requested):
Thanks for adding this! A few blocking things before we merge:
- We should add tests to `admin_terraform_version_integration_test.go` to ensure this new field is behaving as expected when we create/update a version. 
- Please add a changelog entry under the `# Unreleased` section
- See comment below 

Thank you :+1:

admin_terraform_version.go
natalie-todd: Turns out I don't need to edit the `AdminTerraformVersionUpdateOptions` struct for what I am trying to accomplish. I have added `omitempty` to the `Archs` field on the `AdminTerraformVersionCreateOptions` struct. Thanks for catching this.

natalie-todd:
> Thanks for adding this! A few blocking things before we merge:
> 
> * We should add tests to `admin_terraform_version_integration_test.go` to ensure this new field is behaving as expected when we create/update a version.
> * Please add a changelog entry under the `# Unreleased` section
> * See comment below
> 
> Thank you 👍

Hi @sebasslash, I think I addressed these bullet points but let me know if there is anything else I'm missing. Thanks!

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Unreleased
 
 ## Enhancements
+
 * Add support for project level auto destroy settings @simonxmh [#1011](https://github.com/hashicorp/go-tfe/pull/1011)
+* Add `Archs` field to `AdminTerraformVersionCreateOptions` by @natalie-todd [#1022](https://github.com/hashicorp/go-tfe/pull/1022)
```
radditude, line 6: Question for core cloud folks: do we need to note that this is in beta to give any TFE users a heads up that they won't be able to use that field yet?

admin_terraform_version.go
```diff
@@ -194,7 +203,7 @@ func (a *adminTerraformVersions) Delete(ctx context.Context, id string) error {
 }
 
 func (o AdminTerraformVersionCreateOptions) valid() error {
```
radditude, line 205: Do we want to add a validation that checks all fields are provided for architectures, if present? I could also make a case for waiting to add that handling along with update support once the feature is released and the API is fully stable!

radditude (approved):
A couple of very minor questions, but this makes sense to me as a first pass to enable basic usage of the create endpoint while the feature is still flagged!

admin_terraform_version.go
natalie-todd: I just added a validation to check whether at least one valid arch was provided or a valid URL and SHA. Let me know if this seems good or if you were thinking of something else.

radditude: That's perfect! Just a little nicer to skip making the request if we already know it's going to fail.

radditude (approved):

PR #1023, opened 2024-12-26
Add support to remove tag bindings from projects/workspaces

sebasslash:
## Description

In order to remove tag bindings using the workspace or projects API you need to serialize an empty array of bindings (as a jsonapi relationship) and pass that to the PATCH endpoint for either target resource. The problem is the current update calls for either resource sets the tag bindings field to be `omitempty`. So if you serialize an empty slice nothing will occur. This change adds a new method to `Projects` and `Workspaces` called `DeleteTagBindings()` which purposefully serializes an empty slice and sends it over the wire. 

## Output from tests

```

...
```

### DIFF
--- a/project.go
+++ b/project.go
@@ -44,6 +44,9 @@ type Projects interface {
 
 	// AddTagBindings adds or modifies the value of existing tag binding keys for a project.
 	AddTagBindings(ctx context.Context, projectID string, options ProjectAddTagBindingsOptions) ([]*TagBinding, error)
+
+	// DeleteAllTagBindings removes all existing tag bindings for a project.
+	DeleteAllTagBindings(ctx context.Context, projectID string) error
 }
 
 // projects implements Projects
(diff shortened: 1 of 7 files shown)


project.go
```diff
@@ -44,6 +44,9 @@ type Projects interface {
 
 	// AddTagBindings adds or modifies the value of existing tag binding keys for a project.
 	AddTagBindings(ctx context.Context, projectID string, options ProjectAddTagBindingsOptions) ([]*TagBinding, error)
+
+	// DeleteTagBindings removes all existing tag bindings for a project.
+	DeleteTagBindings(ctx context.Context, projectID string) error
```
brandonc, line 49: Can we call this `DeleteAllTagBindings` just in case we ever add the ability to delete specific ones?

workspace.go
```diff
@@ -829,6 +832,31 @@ func (s *workspaces) AddTagBindings(ctx context.Context, workspaceID string, opt
 	return response.Items, err
 }
 
+// DeleteTagBindings removes all tag bindings associated with a workspace. This
+// method will not remove any inherited tags.
```
ctrombley, line 836: 💅 
```suggestion
// method will not remove any inherited tag bindings, which must be 
// removed explicitly removed from the parent resource.
```

workspace.go
```diff
@@ -829,6 +832,31 @@ func (s *workspaces) AddTagBindings(ctx context.Context, workspaceID string, opt
...
+func (s *workspaces) DeleteTagBindings(ctx context.Context, workspaceID string) error {
+	if !validStringID(&workspaceID) {
+		return ErrInvalidWorkspaceID
+	}
+
+	type aliasOpts struct {
```
ctrombley, line 842: No suggestions, just curious about the intent behind the name `aliasOpts`? What is "alias" referring to?

ctrombley (approved):
Nice work! Smoke tested & approved with minor suggestions.

workspace.go
sebasslash: I could have probably used a more descriptive name, but normally structs that "shadow" another -- for example to override struct tag definitions -- are referred to as aliases from what I've seen. 

A more descriptive name would be `aliasUpdateOptions`, but given this wasn't part of the public interface I decided for brevity. An alternative declaration could look like:
```go
opts :=  struct {
		Type        string        `jsonapi:"primary,workspaces"`
		TagBindings []*TagBinding `jsonapi:"relation,tag-bindings"`
	} {
		TagBindings: []*TagBinding{},
	} 
```

Let me know if you would prefer this change

ctrombley: It's ok, no changes needed from my end.

ctrombley (approved):
✨

PR #1042, opened 2025-01-22
[TF-23534] Fix test panic

natalie-todd:
## Description

A new test with a panic was introduced in [#1030](https://github.com/hashicorp/go-tfe/pull/1030). This PR fixes the panic by removing the nil pointer dereference causing it. In the coming weeks there will be a follow up PR which will support the comparison of the architectures returned by the terraform versions CREATE endpoint.


ctrombley (approved):
Thanks!

PR #1043, opened 2025-01-23
Add effective tag binding include relation to projects and workspaces

sebasslash:
## Description

Allow the user to include `EffectiveTagBindings` when listing projects or workspaces. Adds the relation to each respective model. 

## Output from tests

```

...
```


Maed223 (approved):
LGTM 👍

PR #1055, opened 2025-02-07
Add Github action and automated changelog using changie

shwetamurali:
## Description

The Core team recently made a change to the terraform repository to use `changie`, an automated changelog tool, to generate new changelog files for each PR. In addition, they added a new GitHub Action to ensure that changie has been used to generate a changelog, and fails if this has not been done. This ensures that the changelog is completely up to date with all new changes for each release (and users can opt to add the `no-changelog-needed` label to their PRs if they don't see value in that).

I evaluated the idea of using `changie`, but ended up going with just doing the GitHub action. We currently have contributors add to the `CHANGELOG.md` file directly, and this isn't something that I feel gets signficantly easier with the addition of `changie`. However, automation can be super helpful, so if anyone thinks I should add the changie piece too I'd love to discuss!

## Testing
Because the GitHub action will not run until this is merged into main, I tested this on a private repository to ensure that all workflows work as expected.


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,115 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  check-changelog-entry:
+    name: "Check Changelog Entry"
+    runs-on: ubuntu-latest
+    concurrency:
+      group: changelog-${{ github.ref }}
```
DanielMSchmidt, line 28: pull_request_target sets the ref to the base branch (so e.g. main) so we need to use head_ref here

```suggestion
      group: changelog-${{ github.head_ref }}
```

DanielMSchmidt (commented):
Just to double check: When you expect users to edit the changelog it often leads to PRs having merge conflict with the changelog, is that acceptable in your case?

shwetamurali:
> Just to double check: When you expect users to edit the changelog it often leads to PRs having merge conflict with the changelog, is that acceptable in your case?



> Just to double check: When you expect users to edit the changelog it often leads to PRs having merge conflict with the changelog, is that acceptable in your case?

@DanielMSchmidt  After talking to @uturunku1 we decided that it is acceptable for now, and we will create a follow up PR in the future if we want to change this. Thanks for checking!

DanielMSchmidt (approved):

PR #1057, opened 2025-02-07
Support proxied Terraform registry endpoint

paladin-devops:
## Description

This PR adds support for the proxied Terraform Registry endpoint to go-tfe, for fetching modules, documented [here](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#hcp-terraform-registry-implementation).

## Testing plan

Run integration tests, setting the TFE_TOKEN, OAUTH_CLIENT_GITHUB_TOKEN, and GITHUB_REGISTRY_NO_CODE_MODULE_IDENTIFIER environment variables.

## External links

https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#hcp-terraform-registry-implementation

## Output from tests

```
$ go test ./... -v -run TestFunctionsAffectedByChange
=== RUN   TestRegistryModulesReadRegistry
--- PASS: TestRegistryModulesReadRegistry (46.70s)
=== RUN   TestRegistryModulesReadRegistry/fetch_module_from_registry
    --- PASS: TestRegistryModulesReadRegistry/fetch_module_from_registry (39.41s)
PASS

...
```


registry_module.go
```diff
@@ -70,6 +74,66 @@ type RegistryModules interface {
...
+	Version         string `json:"version"`
+	Provider        string `json:"provider"`
+	ProviderLogoURL string `json:"provider_logo_url"`
+	Description     string `json:"description"`
+	Source          string `json:"source"`
+	Tag             string `json:"tag"`
+	PublishedAt     string `json:"published_at"`
+	Downloads       int    `json:"downloads"`
+	Verified        bool   `json:"verified"`
+	Root            `json:"root"`
```
briancain, line 92: Should this be `Root Root ...` ?

registry_module.go
```diff
@@ -572,6 +636,38 @@ func (r *registryModules) Read(ctx context.Context, moduleID RegistryModuleID) (
 
 	return rm, nil
 }
+
+// ReadRegistry fetches a registry module from the Terraform Registry.
+func (r *registryModules) ReadRegistry(ctx context.Context, moduleID RegistryModuleID, version string) (*TerraformRegistryModule, error) {
```
briancain, line 641: Should this be called `ReadByVersion`? It's returning a module but the function is called `ReadRegistry`.

registry_module.go
paladin-devops: As written, `Root`'s fields are "promoted" to the top level. That shortens referencing the `Root` of an example `TerraformRegistryModule`, `tfrm` in this case, to `tfrm.Readme` instead of `tfrm.Root.Readme` (the latter also still works, though). 

So I'd say it's a matter of preference, so long as `Root`'s fields don't conflict with those of the parent struct. I'll make it explicitly `Root Root` though.

registry_module.go
Uk1288: The method takes a `version` param, does it read the module or module-version?

A name suggestion would be something like `ReadTerraformRegistryModule`

registry_module.go
```diff
@@ -70,6 +74,66 @@ type RegistryModules interface {
...
+	Provider        string `json:"provider"`
+	ProviderLogoURL string `json:"provider_logo_url"`
+	Description     string `json:"description"`
+	Source          string `json:"source"`
+	Tag             string `json:"tag"`
+	PublishedAt     string `json:"published_at"`
+	Downloads       int    `json:"downloads"`
+	Verified        bool   `json:"verified"`
+	Root            Root   `json:"root"`
+	// Submodules []Submodule `json:"submodules"`
```
Uk1288, line 93: for the commented out attributes, do they need to be removed? Is there a reason why they are commented out?

registry_module_integration_test.go
```diff
@@ -1323,6 +1324,60 @@ func TestRegistryModulesRead(t *testing.T) {
 	})
 }
 
+func TestRegistryModulesReadRegistry(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+	r := require.New(t)
+
+	// create an org that will be deleted later. the wskp will live here
```
Uk1288, line 1332: seems there is a typo in the comment. Is this comment needed? `orgTestCleanup` explains the cleanup

registry_module_integration_test.go
```diff
@@ -1323,6 +1324,60 @@ func TestRegistryModulesRead(t *testing.T) {
...
+			OrganizationName:  String(org.Name),
+			Identifier:        String(githubIdentifier),
+			Tags:              Bool(true),
+			OAuthTokenID:      String(token.ID),
+			DisplayIdentifier: String(githubIdentifier),
+		},
+	}
+
+	version := "1.0.0"
+	rm, err := client.RegistryModules.CreateWithVCSConnection(ctx, rmOpts)
```
Uk1288, line 1359: a test should be added when the module is type PublicRegistry as well.

registry_module_integration_test.go
```diff
@@ -1323,6 +1324,60 @@ func TestRegistryModulesRead(t *testing.T) {
...
+			Name:         rm.Name,
+			Provider:     rm.Provider,
+			Namespace:    rm.Namespace,
+			RegistryName: rm.RegistryName,
+		}
+		tfm, err := client.RegistryModules.ReadRegistry(ctx, rmID, version)
+		r.NoError(err)
+		r.NotNil(tfm)
+		r.Equal(fmt.Sprintf("%s/%s/%s/%s", orgTest.Name, rm.Name, rm.Provider, version), tfm.ID)
+		r.Equal(version, tfm.Version)
```
Uk1288, line 1377: what do you think about adding checks for expected `TerraformRegistryModule` attributes in addition to the Version attribute?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Enhancements
 
+* Add support for proxied Terraform Registry endpoints for reading detailed module information by @paladin-devops [#1057](https://github.com/hashicorp/go-tfe/pull/1057)
```
Uk1288, line 5:
```suggestion
* Adds `ReadTerraformRegistryModule` to support reading a registry module from Terraform Registry's proxied endpoints by @paladin-devops [#1057](https://github.com/hashicorp/go-tfe/pull/1057) 
```

registry_module.go
paladin-devops: Ah, I left these commented out because the test module I was working with didn't populate these fields. Not knowing their JSON structure, I commented out the fields. I will remove them for now.

registry_module.go
```diff
@@ -45,6 +45,10 @@ type RegistryModules interface {
 	// ReadVersion Read a registry module version
 	ReadVersion(ctx context.Context, moduleID RegistryModuleID, version string) (*RegistryModuleVersion, error)
 
+	// ReadRegistry Reads a registry module from the Terraform Registry.
```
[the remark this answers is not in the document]
paladin-devops: Agreed! I'll update this.

registry_module.go
paladin-devops: It reads the module.

I will update this to be that name instead!

registry_module_integration_test.go
paladin-devops: Not really, I'll remove this.

registry_module_integration_test.go
```diff
@@ -1323,6 +1324,60 @@ func TestRegistryModulesRead(t *testing.T) {
...
+func TestRegistryModulesReadRegistry(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+	r := require.New(t)
+
+	// create an org that will be deleted later. the wskp will live here
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	defer orgTestCleanup()
+
+	org, err := client.Organizations.Read(ctx, orgTest.Name)
```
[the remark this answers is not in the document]
paladin-devops: It's not necessary, removing.

CHANGELOG.md
paladin-devops: Done!

registry_module_integration_test.go
paladin-devops: I added several more checks. I did not update the test for every attribute of the module, but many of them. Also, several of the test values are hard-coded, making this test very brittle to changes in the [module under test](https://github.com/hashicorp/terraform-random-module). For example, if a new version of that module is published with one additional input variable, then the `Len` check for the `Inputs` field will fail.

Just wanted to call that out! 😄 I still think it's good that other attributes are being verified.

Uk1288 (approved):

PR #1060, opened 2025-02-18
Support unsetting an organization's default project

mkam:
## Description

Allows explicitly unsetting an organization's default project by setting it as a `NullableRelationship` rather than just a `Project`, as it's not required to have a default project.

## Testing plan
1. Create an organization.
1. Create a new project in the organization
1. Update the organization’s default project to the new project
1. Verify in the UI on the organization's settings that the default project has been updated
1. Delete the default project setting
1. Verify in the UI on the organization's settings that there is not a default project set

## Output from tests
```
-> % ENABLE_BETA=1 go test ./... -v -run "TestOrganizationsUpdate/with_different_default_project"
?   	github.com/hashicorp/go-tfe/examples/backing_data	[no test files]
?   	github.com/hashicorp/go-tfe/examples/configuration_versions	[no test files]
?   	github.com/hashicorp/go-tfe/examples/organizations	[no test files]
?   	github.com/hashicorp/go-tfe/examples/projects	[no test files]
?   	github.com/hashicorp/go-tfe/examples/registry_modules	[no test files]
?   	github.com/hashicorp/go-tfe/examples/run_errors	[no test files]
?   	github.com/hashicorp/go-tfe/examples/state_versions	[no test files]
?   	github.com/hashicorp/go-tfe/examples/users	[no test files]
?   	github.com/hashicorp/go-tfe/examples/workspaces	[no test files]
?   	github.com/hashicorp/go-tfe/mocks	[no test files]
=== RUN   TestOrganizationsUpdate
=== RUN   TestOrganizationsUpdate/with_different_default_project
--- PASS: TestOrganizationsUpdate (2.56s)
    --- PASS: TestOrganizationsUpdate/with_different_default_project (2.22s)
PASS
ok  	github.com/hashicorp/go-tfe	2.851s
...
```

PR #1064, opened 2025-02-24
Add nullable relationship to default Project

netramali:
## Description

There was 1 place to be updated that was missed in this PR: https://github.com/hashicorp/go-tfe/pull/1060/files so just adding that in this PR.

## Testing plan

Create an organization.
Create a new project in the organization
Update the organization’s default project to the new project
Verify in the UI on the organization's settings that the default project has been updated
Delete the default project setting
Verify in the UI on the organization's settings that there is not a default project set

## Output from tests

```

...
```


netramali:
No longer needed.

PR #1074, opened 2025-03-06
BREAKING: removes TotalCount/TotalPages from ListForOrganization

### LINKED ISSUE #1073: Remove TotalCount from Run ListForOrganization interface
#### go-tfe version

```plaintext
v1.76.0
```

## Description

The `TotalCount` pagination parameter is being removed prior to releasing it to TFE and requires a breaking change to go-tfe.

brandonc:
Closes #1073 

## Description

As we seek to release the Organization Runs feature more widely, it needs to be modified to remove the TotalCount/TotalPages pagination attributes. This feature was added in the last release of go-tfe.

## Testing plan

Today, you'll find that the TotalCount/TotalPages are still in the API response but will be removed very soon.

## Notes

The unmarshaler had to be modified to check for a struct field called "PaginationPrevNext" and preferring that with the missing fields. The implicit assumption is that any `Items` struct will either define `*Pagination` _or_ `*PaginationNextPrev`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## BREAKING CHANGES
+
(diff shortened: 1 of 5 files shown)


brandonc:
@Maed223 Well it's no coincidence I approved it 😄

PR #1126, opened 2025-06-11
Use go-tfe client address for registry URL

paladin-devops:
## Description

Update the `ReadTerraformRegistryModule` method of `registryModules` to use the HCP TF address configured on the go-tfe client.

## Output from tests

```

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+* Add support for HCP Terraform `/api/v2/workspaces/{external_id}/all-vars` API endpoint to fetch the list of all variables available to a workspace (include inherited variables from varsets) by @debrin-hc [#1105](https://github.com/hashicorp/go-tfe/pull/1105)
+
+* Adds BETA support for listing `StackDeploymentGroups`, which is EXPERIMENTAL, SUBJECT TO CHANGE, and may not be available to all users by @hwatkins05-hashicorp [#1128](https://github.com/hashicorp/go-tfe/pull/1128)
 
 # v1.82.0
 
--- a/helper_test.go
+++ b/helper_test.go
@@ -2209,19 +2209,48 @@ func createTeamTokenWithOptions(t *testing.T, client *Client, tm *Team, options 
 }
 
 func createVariable(t *testing.T, client *Client, w *Workspace) (*Variable, func()) {
+	options := VariableCreateOptions{
+		Key:         String(randomString(t)),
+		Value:       String(randomString(t)),
+		Category:    Category(CategoryTerraform),
+		Description: String(randomString(t)),
+	}
+	return createVariableWithOptions(t, client, w, options)
+}
+
+func createVariableWithOptions(t *testing.T, client *Client, w *Workspace, options VariableCreateOptions) (*Variable, func()) {
 	var wCleanup func()
 
 	if w == nil {
 		w, wCleanup = createWorkspace(t, client, nil)
 	}
 
+	if options.Key == nil {
+		options.Key = String(randomString(t))
+	}
+
+	if options.Value == nil {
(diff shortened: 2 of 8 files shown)


registry_module.go
```diff
@@ -635,23 +635,28 @@ func (r *registryModules) Read(ctx context.Context, moduleID RegistryModuleID) (
 	return rm, nil
 }
 
-// ReadRegistry fetches a registry module from the Terraform Registry.
+// ReadTerraformRegistryModule fetches a registry module from the Terraform Registry.
 func (r *registryModules) ReadTerraformRegistryModule(ctx context.Context, moduleID RegistryModuleID, version string) (*TerraformRegistryModule, error) {
-	u := fmt.Sprintf(
-		"https://app.terraform.io/api/registry/v1/modules/%s/%s/%s/%s",
+	u, err := url.JoinPath(r.client.baseURL.Scheme+"://"+r.client.baseURL.Host, fmt.Sprintf("api/registry/v1/modules/%s/%s/%s/%s",
```
brandonc: It turns out you can configure the client with a [RegistryBasePath](https://github.com/hashicorp/go-tfe/blob/main/tfe.go#L64-L65). By default, the path is `api/registry`

We should probably be sensitive to this configuration:

`url.JoinPath(r.client.registryBaseURL, fmt.Sprintf("v1/modules/%s/%s/%s/%s", ...)`

paladin-devops: Agreed! Perhaps this would work better instead.
```suggestion
	u, err := url.JoinPath(r.client.baseURL.Scheme+"://"+r.client.baseURL.Host, fmt.Sprintf("%s/v1/modules/%s/%s/%s/%s",
		r.client.registryBaseURL,
		moduleID.Namespace,
		moduleID.Name,
		moduleID.Provider,
		version,
	))
```

sebasslash: If I'm not mistaken all we need to drop here is `https://app.terraform.io` in the url string we pass to `NewRequest`
```suggestion
	u := fmt.Sprintf("api/registry/v1/modules/%s/%s/%s/%s", ...)
```

and for public registry:

```suggestion
	u := fmt.Sprintf("api/registry/public/v1/modules/%s/%s/%s/%s", ...)
```
We parse the client's address for the registry [here](https://github.com/hashicorp/go-tfe/blob/7044c101ef18b67b72e4dd6de7cd4d4397eeae3d/tfe.go#L388-L396). 

See [example](https://github.com/hashicorp/go-tfe/blob/7044c101ef18b67b72e4dd6de7cd4d4397eeae3d/gpg_key.go#L95-L96)

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * Adds BETA support for speculative runs with `Stacks` resources and removes VCS repo validity check on `Stack` creation, which is EXPERIMENTAL, SUBJECT TO CHANGE, and may not be available to all users by @hwatkins05-hashicorp [#1119](https://github.com/hashicorp/go-tfe/pull/1119)
 * Adds support for listing team tokens, by @mkam [#1109](https://github.com/hashicorp/go-tfe/pull/1109)
 
+## Fixes
```
brandonc:
```suggestion
## Bug Fixes
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
...
+
+* Use go-tfe client address for ReadTerraformRegistryModule, by @paladin-devops [#1126](https://github.com/hashicorp/go-tfe/pull/1126)
```
brandonc:
```suggestion
* Fixes hard coded public terraform registry URL for ReadTerraformRegistryModule, by @paladin-devops [#1126](https://github.com/hashicorp/go-tfe/pull/1126)
```

PR #1131, opened 2025-06-16
Remove omission of vcs_repo in stack update

Maed223:
## Description

Previously forgotten in the PR removing the VCS repo validation was to remove the `omit-empty` jsonapi tag in the `StackUpdateOptions`. This is needed as we need to explicitly give `nil` in the request to remove the VCS backing of a Stack. Before this wouldn't work as the nil was being omitted from the request. This is a requirement for the tfe provider work.

## Testing plan

Added a test to cover this change

## Output from tests
Screenshot 2025-06-16 at 4 25 16 PM
Screenshot 2025-06-16 at 4 34 20 PM
Screenshot 2025-06-16 at 4 35 06 PM


ctrombley (approved):
LGTM!

PR #1138, opened 2025-06-19
Stack configuration list endpoint

shwetamurali:
## Description

Adding an endpoint to get the list of corresponding configurations for a given `stack ID` per the [OpenAPI Spec](https://github.com/hashicorp/atlas/blob/main/openapi/spec/resources/stacks/stack-configurations/collection.yml)

## Testing plan

1. Fork `pet-nulls-stack` or any repo with stacks
2. Disable stacks-ga flag in local atlas
3. Run test

## Output from tests
`go test -v -run TestStackConfigurationList`
```
=== RUN   TestStackConfigurationList
=== RUN   TestStackConfigurationList/with_pagination_options
--- PASS: TestStackConfigurationList (8.50s)
    --- PASS: TestStackConfigurationList/with_pagination_options (0.13s)
PASS
ok      github.com/hashicorp/go-tfe     8.894s
```

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -131,6 +131,8 @@ type StackConfiguration struct {
 	ErrorMessage         *string                             `jsonapi:"attr,error-message"`
 	EventStreamURL       string                              `jsonapi:"attr,event-stream-url"`
 	Diagnostics          []*StackDiagnostic                  `jsonapi:"attr,diags"`
+
+	Relationships map[string]interface{} `json:"relationships"`
 }
 
 // StackDeployment represents a stack deployment, specified by configuration
--- a/stack.go
+++ b/stack.go
@@ -132,7 +132,7 @@ type StackConfiguration struct {
 	EventStreamURL       string                              `jsonapi:"attr,event-stream-url"`
 	Diagnostics          []*StackDiagnostic                  `jsonapi:"attr,diags"`
 
-	Relationships map[string]interface{} `json:"relationships"`
+	Stack *Stack `jsonapi:"relation,stack"`
 }
 
 // StackDeployment represents a stack deployment, specified by configuration
--- a/stack_configuration_integration_test.go
+++ b/stack_configuration_integration_test.go
@@ -55,6 +55,15 @@ func TestStackConfigurationList(t *testing.T) {
 		require.NotEmpty(t, cfg.ID)
 		require.NotEmpty(t, cfg.Status)
 		require.GreaterOrEqual(t, cfg.SequenceNumber, 1)
+
+		if cfg.Relationships != nil {
+			stackRel, ok := cfg.Relationships["stack"].(map[string]interface{})
+			require.True(t, ok)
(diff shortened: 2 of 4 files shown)


stack_configuration_integration_test.go
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package tfe
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	. "github.com/stretchr/testify/require"
```
jpadrianoGo: minor: Noticed use of dot importer, for files with large number of exported fubctions, there is a slight difference in compile-time as it may need track all symbols. Collisions may be avoided for when using multiple package imports.

stack_configuration_integration_test.go
```diff
@@ -0,0 +1,74 @@
...
+	orgTest, orgTestCleanup := createOrganization(t, client)
+	t.Cleanup(orgTestCleanup)
+
+	oauthClient, cleanup := createOAuthClient(t, client, orgTest, nil)
+	t.Cleanup(cleanup)
+
+	stack, err := client.Stacks.Create(ctx, StackCreateOptions{
+		Name: "test-stack-list",
+		VCSRepo: &StackVCSRepoOptions{
+			Identifier:   "shwetamurali/pet-nulls-stack",
```
jpadrianoGo: minor: can be hashicorp-guide to avoid pushing to repo

stack_configuration_integration_test.go
```diff
@@ -0,0 +1,86 @@
...
+	_, err = client.Stacks.UpdateConfiguration(ctx, stack.ID)
+	require.NoError(t, err)
+
+	list, err := client.StackConfigurations.List(ctx, stack.ID, nil)
+	require.NoError(t, err)
+	require.NotNil(t, list)
+	assert.Equal(t, len(list.Items), 2)
+
+	// Assert attributes for each configuration
+	for _, cfg := range list.Items {
```
hwatkins05-hashicorp, line 54: list item assertions look really good! 💯

hwatkins05-hashicorp (approved):

stack.go
```diff
@@ -131,6 +131,8 @@ type StackConfiguration struct {
 	ErrorMessage         *string                             `jsonapi:"attr,error-message"`
 	EventStreamURL       string                              `jsonapi:"attr,event-stream-url"`
 	Diagnostics          []*StackDiagnostic                  `jsonapi:"attr,diags"`
+
+	Relationships map[string]interface{} `json:"relationships"`
```
markdecrane: Rather using a `map[string]interface{}` for the relationships, we typically denote the specific item in the relationship. Then in the `jsonapi` tags we mark it as coming from the relationship. You can look at the `StackDeployment` struct below to get an idea on how that looks (attribute `CurrentStackState *StackState` specifically). The only present relationship here will be the configuration's Stack.

Maed223 (approved):
💯

PR #1156, opened 2025-07-14
Fix primary id keys in stacks related struct types

Maed223:
## Description

Fixing the incorrect primary id keys in some stacks related structs.

## Testing plan

Existing tests should pass


ctrombley (dismissed):
👍

ctrombley (approved):
👍

PR #1162, opened 2025-07-25
[TF-18724] HYOK Configurations and OIDC Configurations support

iuri-slywitch-hashicorp:
Added support for HYOK Configuration and OIDC Configuration.

## Description

These changes add HYOK functionalities to the go-tfe library.

## Testing plan

Test files ensure creating, reading, updating and deleting OIDC Configurations.

## External links

- [JIRA OIDC Configuration](https://hashicorp.atlassian.net/browse/TF-27659)
- [JIRA HYOK Configuration](https://hashicorp.atlassian.net/browse/TF-27658)

## Output from tests

```

...
```

### DIFF
--- a/aws_oidc_configuration.go
+++ b/aws_oidc_configuration.go
@@ -1,0 +1,157 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+type AwsOidcConfigurations interface {
+	Create(ctx context.Context, organization string, options AwsOidcConfigurationsCreateOptions) (*AwsOidcConfiguration, error)
+
+	Read(ctx context.Context, hyokID string, options *AwsOidcConfigurationsReadOptions) (*AwsOidcConfiguration, error)
+
+	Update(ctx context.Context, hyokID string, options AwsOidcConfigurationsUpdateOptions) (*AwsOidcConfiguration, error)
+
+	Delete(ctx context.Context, oidcID string) error
+}
+
+type awsOidcConfigurations struct {
+	client *Client
+}
+
+var _ AwsOidcConfigurations = &awsOidcConfigurations{}
+
+type AwsOidcConfiguration struct {
+	Type    string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleArn string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AwsOidcConfigurationsCreateOptions struct {
+	Type    string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleArn string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AwsOidcConfigurationsIncludeOpt string
+
+const (
+	AwsOidcConfigurationsIncludeOrganization            AwsOidcConfigurationsIncludeOpt = "organization"
+	AwsOidcConfigurationsIncludeProject                 AwsOidcConfigurationsIncludeOpt = "project"
+	AwsOidcConfigurationsIncludeLatestHyokConfiguration AwsOidcConfigurationsIncludeOpt = "latest_hyok_configuration"
+	AwsOidcConfigurationsIncludeHyokDiagnostics         AwsOidcConfigurationsIncludeOpt = "hyok_diagnostics"
+)
+
+type AwsOidcConfigurationsReadOptions struct {
+	Include []AwsOidcConfigurationsIncludeOpt `url:"include,omitempty"`
+}
+
+type AwsOidcConfigurationsUpdateOptions struct {
+	Type    string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleArn string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+func (o *AwsOidcConfigurationsCreateOptions) valid() error {
+	if o.RoleArn == "" {
+		return ErrRequiredRoleArn
+	}
+
+	return nil
+}
+
+func (aoc *awsOidcConfigurations) Create(ctx context.Context, organization string, options AwsOidcConfigurationsCreateOptions) (*AwsOidcConfiguration, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := aoc.client.NewRequest("POST", fmt.Sprintf("organizations/%s/oidc-configurations", organization), &options)
+	if err != nil {
+		return nil, err
+	}
+
+	awsOidcConfiguration := &AwsOidcConfiguration{}
+	err = req.Do(ctx, awsOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return awsOidcConfiguration, nil
+}
+
+func (o *AwsOidcConfigurationsReadOptions) valid() error {
+	return nil
+}
+
+func (aoc *awsOidcConfigurations) Read(ctx context.Context, oidcID string, options *AwsOidcConfigurationsReadOptions) (*AwsOidcConfiguration, error) {
+	if options != nil {
+		if err := options.valid(); err != nil {
+			return nil, err
+		}
+	}
+
+	req, err := aoc.client.NewRequest("GET", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), options)
+	if err != nil {
+		return nil, err
+	}
+
+	awsOidcConfiguration := &AwsOidcConfiguration{}
+	err = req.Do(ctx, awsOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return awsOidcConfiguration, nil
+}
+
+func (o *AwsOidcConfigurationsUpdateOptions) valid() error {
+	if o.RoleArn == "" {
+		return ErrRequiredRoleArn
+	}
+
+	return nil
+}
+
+func (aoc *awsOidcConfigurations) Update(ctx context.Context, oidcID string, options AwsOidcConfigurationsUpdateOptions) (*AwsOidcConfiguration, error) {
+	if !validStringID(&oidcID) {
+		return nil, ErrInvalidOidc
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := aoc.client.NewRequest("PATCH", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), &options)
+	if err != nil {
+		return nil, err
+	}
+
+	awsOidcConfiguration := &AwsOidcConfiguration{}
+	err = req.Do(ctx, awsOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return awsOidcConfiguration, nil
+}
+
+func (aoc *awsOidcConfigurations) Delete(ctx context.Context, oidcID string) error {
+	if !validStringID(&oidcID) {
+		return ErrInvalidOidc
+	}
+
+	req, err := aoc.client.NewRequest("DELETE", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), nil)
+	if err != nil {
+		return err
+	}
+
+	return req.Do(ctx, nil)
+}
--- a/azure_oidc_configuration.go
+++ b/azure_oidc_configuration.go
@@ -1,0 +1,179 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+type AzureOidcConfigurations interface {
+	Create(ctx context.Context, organization string, options AzureOidcConfigurationsCreateOptions) (*AzureOidcConfiguration, error)
+
+	Read(ctx context.Context, oidcID string, options *AzureOidcConfigurationsReadOptions) (*AzureOidcConfiguration, error)
+
+	Update(ctx context.Context, oidcID string, options AzureOidcConfigurationsUpdateOptions) (*AzureOidcConfiguration, error)
+
+	Delete(ctx context.Context, oidcID string) error
+}
+
+type azureOidcConfigurations struct {
+	client *Client
+}
+
+var _ AzureOidcConfigurations = &azureOidcConfigurations{}
+
+type AzureOidcConfiguration struct {
+	Type           string `jsonapi:"primary,azure-oidc-configurations"`
+	ClientID       string `jsonapi:"attr,client-id"`
+	SubscriptionID string `jsonapi:"attr,subscription-id"`
+	TenantID       string `jsonapi:"attr,tenant-id"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AzureOidcConfigurationsCreateOptions struct {
+	Type           string `jsonapi:"primary,azure-oidc-configurations"`
+	ClientID       string `jsonapi:"attr,client-id"`
+	SubscriptionID string `jsonapi:"attr,subscription-id"`
+	TenantID       string `jsonapi:"attr,tenant-id"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AzureOidcConfigurationsIncludeOpt string
+
+const (
+	AzureOidcConfigurationsIncludeOrganization            AzureOidcConfigurationsIncludeOpt = "organization"
+	AzureOidcConfigurationsIncludeProject                 AzureOidcConfigurationsIncludeOpt = "project"
+	AzureOidcConfigurationsIncludeLatestHyokConfiguration AzureOidcConfigurationsIncludeOpt = "latest_hyok_configuration"
+	AzureOidcConfigurationsIncludeHyokDiagnostics         AzureOidcConfigurationsIncludeOpt = "hyok_diagnostics"
+)
+
+type AzureOidcConfigurationsReadOptions struct {
+	Include []AzureOidcConfigurationsIncludeOpt `url:"include,omitempty"`
+}
+
+type AzureOidcConfigurationsUpdateOptions struct {
+	Type           string `jsonapi:"primary,azure-oidc-configurations"`
+	ClientID       string `jsonapi:"attr,client-id"`
+	SubscriptionID string `jsonapi:"attr,subscription-id"`
+	TenantID       string `jsonapi:"attr,tenant-id"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+func (o *AzureOidcConfigurationsCreateOptions) valid() error {
+	if o.ClientID == "" {
+		return ErrRequiredClientID
+	}
+
+	if o.SubscriptionID == "" {
+		return ErrRequiredSubscriptionID
+	}
+
+	if o.TenantID == "" {
+		return ErrRequiredTenantID
+	}
+
+	return nil
+}
+
+func (aoc *azureOidcConfigurations) Create(ctx context.Context, organization string, options AzureOidcConfigurationsCreateOptions) (*AzureOidcConfiguration, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := aoc.client.NewRequest("POST", fmt.Sprintf("organizations/%s/oidc-configurations", organization), &options)
+	if err != nil {
+		return nil, err
+	}
+
+	azureOidcConfiguration := &AzureOidcConfiguration{}
+	err = req.Do(ctx, azureOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return azureOidcConfiguration, nil
+}
+
+func (o *AzureOidcConfigurationsReadOptions) valid() error {
+	return nil
+}
+
+func (aoc *azureOidcConfigurations) Read(ctx context.Context, oidcID string, options *AzureOidcConfigurationsReadOptions) (*AzureOidcConfiguration, error) {
+	if options != nil {
+		if err := options.valid(); err != nil {
+			return nil, err
+		}
+	}
+
+	req, err := aoc.client.NewRequest("GET", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), options)
+	if err != nil {
+		return nil, err
+	}
+
+	azureOidcConfiguration := &AzureOidcConfiguration{}
+	err = req.Do(ctx, azureOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return azureOidcConfiguration, nil
+}
+
+func (o *AzureOidcConfigurationsUpdateOptions) valid() error {
+	if o.ClientID == "" {
+		return ErrRequiredClientID
+	}
+
+	if o.SubscriptionID == "" {
+		return ErrRequiredSubscriptionID
+	}
+
+	if o.TenantID == "" {
+		return ErrRequiredTenantID
+	}
+
+	return nil
+}
+
+func (aoc *azureOidcConfigurations) Update(ctx context.Context, oidcID string, options AzureOidcConfigurationsUpdateOptions) (*AzureOidcConfiguration, error) {
+	if !validStringID(&oidcID) {
+		return nil, ErrInvalidOidc
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := aoc.client.NewRequest("PATCH", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), &options)
+	if err != nil {
+		return nil, err
+	}
+
+	azureOidcConfiguration := &AzureOidcConfiguration{}
+	err = req.Do(ctx, azureOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return azureOidcConfiguration, nil
+}
+
+func (aoc *azureOidcConfigurations) Delete(ctx context.Context, oidcID string) error {
+	if !validStringID(&oidcID) {
+		return ErrInvalidOidc
+	}
+
+	req, err := aoc.client.NewRequest("DELETE", fmt.Sprintf("oidc-configurations/%s", url.PathEscape(oidcID)), nil)
+	if err != nil {
+		return err
+	}
+
+	return req.Do(ctx, nil)
+}
--- a/errors.go
+++ b/errors.go
@@ -240,6 +240,8 @@ var (
 	ErrInvalidTaskResultsCallbackStatus = fmt.Errorf("invalid value for task result status. Must be either `%s`, `%s`, or `%s`", TaskFailed, TaskPassed, TaskRunning)
 
 	ErrInvalidDescriptionConflict = errors.New("invalid attributes\n\nValidation failed: Description has already been taken")
+
+	ErrInvalidOidc = errors.New("invalid value for OIDC configuration ID")
 )
 
 var (
@@ -399,4 +401,26 @@ var (
 	ErrRequiredRawState = errors.New("RawState is required")
 
 	ErrStateVersionUploadNotSupported = errors.New("upload not supported by this version of Terraform Enterprise")
+
+	ErrRequiredRoleArn = errors.New("role-arn is required for AWS OIDC configuration")
+
+	ErrRequiredServiceAccountEmail = errors.New("service-account-email is required for GCP OIDC configuration")
+
+	ErrRequiredProjectNumber = errors.New("project-number is required for GCP OIDC configuration")
+
+	ErrRequiredWorkloadProviderName = errors.New("workload-provider-name is required for GCP OIDC configuration")
+
+	ErrRequiredClientID = errors.New("client-id is required for Azure OIDC configuration")
+
+	ErrRequiredSubscriptionID = errors.New("subscription-id is required for Azure OIDC configuration")
+
+	ErrRequiredTenantID = errors.New("tenant-id is required for Azure OIDC configuration")
+
+	ErrRequiredAddress = errors.New("address is required for Vault OIDC configuration")
+
+	ErrRequiredRoleName = errors.New("role is required for Vault OIDC configuration")
+
+	ErrRequiredJWTAuthPath = errors.New("auth_path is required for Vault OIDC configuration")
+
+	ErrRequiredTLSCACertificate = errors.New("encoded_cacert is required for Vault OIDC configuration")
 )
--- a/gcp_oidc_configuration.go
+++ b/gcp_oidc_configuration.go
@@ -1,0 +1,179 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+type GcpOidcConfigurations interface {
+	Create(ctx context.Context, organization string, options GcpOidcConfigurationsCreateOptions) (*GcpOidcConfiguration, error)
+
+	Read(ctx context.Context, oidcID string, options *GcpOidcConfigurationsReadOptions) (*GcpOidcConfiguration, error)
+
+	Update(ctx context.Context, oidcID string, options GcpOidcConfigurationsUpdateOptions) (*GcpOidcConfiguration, error)
+
+	Delete(ctx context.Context, oidcID string) error
+}
+
+type gcpOidcConfigurations struct {
+	client *Client
+}
+
+var _ GcpOidcConfigurations = &gcpOidcConfigurations{}
+
+type GcpOidcConfiguration struct {
+	Type                 string `jsonapi:"primary,gcp-oidc-configurations"`
+	ServiceAccountEmail  string `jsonapi:"attr,service-account-email"`
+	ProjectNumber        string `jsonapi:"attr,project-number"`
+	WorkloadProviderName string `jsonapi:"attr,workload-provider-name"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type GcpOidcConfigurationsCreateOptions struct {
+	Type                 string `jsonapi:"primary,gcp-oidc-configurations"`
+	ServiceAccountEmail  string `jsonapi:"attr,service-account-email"`
+	ProjectNumber        string `jsonapi:"attr,project-number"`
+	WorkloadProviderName string `jsonapi:"attr,workload-provider-name"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type GcpOidcConfigurationsIncludeOpt string
+
+const (
+	GcpOidcConfigurationsIncludeOrganization            GcpOidcConfigurationsIncludeOpt = "organization"
+	GcpOidcConfigurationsIncludeProject                 GcpOidcConfigurationsIncludeOpt = "project"
+	GcpOidcConfigurationsIncludeLatestHyokConfiguration GcpOidcConfigurationsIncludeOpt = "latest_hyok_configuration"
+	GcpOidcConfigurationsIncludeHyokDiagnostics         GcpOidcConfigurationsIncludeOpt = "hyok_diagnostics"
+)
+
+type GcpOidcConfigurationsReadOptions struct {
+	Include []GcpOidcConfigurationsIncludeOpt `url:"include,omitempty"`
+}
+
+type GcpOidcConfigurationsUpdateOptions struct {
+	Type                 string `jsonapi:"primary,gcp-oidc-configurations"`
+	ServiceAccountEmail  string `jsonapi:"attr,service-account-email"`
+	ProjectNumber        string `jsonapi:"attr,project-number"`
+	WorkloadProviderName string `jsonapi:"attr,workload-provider-name"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+func (o *GcpOidcConfigurationsCreateOptions) valid() error {
+	if o.ServiceAccountEmail == "" {
+		return ErrRequiredServiceAccountEmail
+	}
+
+	if o.ProjectNumber == "" {
+		return ErrRequiredProjectNumber
+	}
+
+	if o.WorkloadProviderName == "" {
+		return ErrRequiredWorkloadProviderName
+	}
+
+	return nil
+}
+
+func (goc *gcpOidcConfigurations) Create(ctx context.Context, organization string, options GcpOidcConfigurationsCreateOptions) (*GcpOidcConfiguration, error) {
+	if !validStringID(&organization) {
+		return nil, ErrInvalidOrg
+	}
+
+	if err := options.valid(); err != nil {
+		return nil, err
+	}
+
+	req, err := goc.client.NewRequest("POST", fmt.Sprintf("organizations/%s/oidc-configurations", organization), &options)
+	if err != nil {
+		return nil, err
+	}
+
+	gcpOidcConfiguration := &GcpOidcConfiguration{}
+	err = req.Do(ctx, gcpOidcConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return gcpOidcConfiguration, nil
+}
+
+func (o *GcpOidcConfigurationsReadOptions) valid() error {
+	return nil
+}
+
+func (goc *gcpOidcConfigurations) Read(ctx context.Context, oidcID string, options *GcpOidcConfigurationsReadOptions) (*GcpOidcConfiguration, error) {
+	if options != nil {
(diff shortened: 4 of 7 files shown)


aws_oidc_configuration.go
```diff
@@ -0,0 +1,134 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+type AwsOidcConfigurations interface {
```
helenjw:
```suggestion
type AWSOIDCConfigurations interface {
```

Looking at the Go Wiki (for PRs), it seems like the preferred way of using [initialisms](https://go.dev/wiki/CodeReviewComments#initialisms) is to keep it case-consistent. In our own repo as well, there are examples of using `RSAPublicKey` and `GPGKeys`.

Ditto for all other providers (eg. `GCPOIDCConfiguration`, `VaultOIDCConfiguration`), and any mentions of `OIDC`.

aws_oidc_configuration.go
```diff
@@ -0,0 +1,134 @@
...
+	Create(ctx context.Context, organization string, options AwsOidcConfigurationsCreateOptions) (*AwsOidcConfiguration, error)
+
+	Read(ctx context.Context, hyokID string) (*AwsOidcConfiguration, error)
+
+	Update(ctx context.Context, hyokID string, options AwsOidcConfigurationsUpdateOptions) (*AwsOidcConfiguration, error)
+
+	Delete(ctx context.Context, oidcID string) error
+}
+
+type awsOidcConfigurations struct {
```
helenjw:
```suggestion
type awsOIDCConfigurations struct {
```
Since this isn't exported, we can keep `aws` lowercase

aws_oidc_configuration.go
```diff
@@ -0,0 +1,134 @@
...
+	client *Client
+}
+
+var _ AwsOidcConfigurations = &awsOidcConfigurations{}
+
+type AwsOidcConfiguration struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleArn string `jsonapi:"attr,role-arn"`
```
helenjw:
```suggestion
	RoleARN string `jsonapi:"attr,role-arn"`
```
ARN stands for Amazon Resource Name

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+type KMSOptions struct {
+	KeyRegion   string `jsonapi:"attr,key-region,omitempty"`   // AWS KMS
+	KeyLocation string `jsonapi:"attr,key-location,omitempty"` // GCP KMS
+	KeyRingID   string `jsonapi:"attr,key-ring-id,omitempty"`  // GCP KMS
+}
+
+type HyokConfigurationsCustomerKeyVersion struct {
+	ID   string `jsonapi:"primary,hyok-customer-key-versions"`
+	Type string `jsonapi:"attr,type"`
+}
```
helenjw:
```suggestion
```

This should not be in the HYOK config file. It's a separate resource that will be implemented with a different PR. It's fine to leave this out for now.

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+	AwsOidcConfiguration   *AwsOidcConfiguration
+	GcpOidcConfiguration   *GcpOidcConfiguration
+	AzureOidcConfiguration *AzureOidcConfiguration
+	VaultOidcConfiguration *VaultOidcConfiguration
+}
+
+type KMSOptions struct {
+	KeyRegion   string `jsonapi:"attr,key-region,omitempty"`   // AWS KMS
+	KeyLocation string `jsonapi:"attr,key-location,omitempty"` // GCP KMS
+	KeyRingID   string `jsonapi:"attr,key-ring-id,omitempty"`  // GCP KMS
```
helenjw: Nit: I think it may be better grouped visually like this
```suggestion
	// AWS
	KeyRegion   string `jsonapi:"attr,key-region,omitempty"` 
	// GCP
	KeyLocation string `jsonapi:"attr,key-location,omitempty"`
	KeyRingID   string `jsonapi:"attr,key-ring-id,omitempty"`
```

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+type HyokConfigurationsUpdateOptions struct {
+	ID   string `jsonapi:"primary,hyok-configurations"`
+	Type string `jsonapi:"attr,type"`
+
+	// Attributes
+	KekID      string      `jsonapi:"attr,kek-id"`
+	KMSOptions *KMSOptions `jsonapi:"attr,kms-options"`
+	Name       string      `jsonapi:"attr,name"`
+	Primary    bool        `jsonapi:"attr,primary"`
+	Status     string      `jsonapi:"attr,status"`
```
helenjw: Can narrow down what strings we expect by defining a new type and listing the specific statuses that we can expect.
```suggestion
	Status     HYOKConfigurationStatus      `jsonapi:"attr,status"`
```

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+
+	// Relationships
+	Organization      *Organization            `jsonapi:"relation,organization"`
+	OidcConfiguration *OidcConfigurationChoice `jsonapi:"polyrelation,oidc-configuration"`
+	AgentPool         *AgentPool               `jsonapi:"relation,agent-pool"`
+}
+
+func (h *HyokConfigurationsListOptions) valid() error {
+	return nil
+}
```
helenjw: Remove this and remove the `options.valid()` call if it's not being used
```suggestion
```

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+const (
+	HyokConfigurationsIncludeHyokCustomerKeyVersions HyokConfigurationsIncludeOpt = "hyok_customer_key_versions"
+	HyokConfigurationsIncludeOidcCconfiguration      HyokConfigurationsIncludeOpt = "oidc_configuration"
+)
+
+type HyokConfigurationsListOptions struct {
+	ListOptions
+	ProjectID    string                         `url:"filter[project[id]],omitempty"`
+	Sort         HyokConfigurationsSortColumn   `url:"sort,omitempty"`
+	SearchByName string                         `url:"search[name],omitempty"`
```
helenjw: Just for my information, can you show me how you got this to work in your testing? I have not set this up locally, but I'm not aware that you can do this from the UI either. The only thing I can think of is using the `q` query parameter to get a fuzzy match of the name

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
...
+	err = req.Do(ctx, hyokConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return hyokConfiguration, nil
+}
+
+func (h *HyokConfigurationsUpdateOptions) valid() error {
+	return nil
```
helenjw: I think we can add some logic here to update an HYOK config, probably similar to the create option

hyok_configuration.go
```diff
@@ -0,0 +1,342 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+type HyokConfigurations interface {
```
helenjw:
```suggestion
type HYOKConfigurations interface {
```

helenjw (changes requested):
Left some comments on your draft!

iuri-slywitch-hashicorp:
it's interesting that looking at the`hyok_configuration.rb` in Atlas project (/Users/iurislywitch/Desktop/atlas/app/models/hyok_configuration.rb), we have different initialisms...

`sig { returns(T::Boolean) }
  def aws_configured?
    oidc_configuration_type == 'AwsOidcConfiguration'
  end

  sig { returns(T::Boolean) }
  def gcp_configured?
    oidc_configuration_type == 'GcpOidcConfiguration'
  end

  sig { returns(T::Boolean) }
  def azure_configured?
    oidc_configuration_type == 'AzureOidcConfiguration'
  end

  sig { returns(T::Boolean) }
  def vault_configured?
    oidc_configuration_type == 'VaultOidcConfiguration'
  end`

hyok_configuration.go
iuri-slywitch-hashicorp: yes... I just don't know what exactly. Talked to Jarrett and he didn't recommend writing validations for parameters here. Not sure if we could write if-statement cases for arguments that could be missing. If so, which ones would it be?

aws_oidc_configuration.go
```diff
@@ -0,0 +1,134 @@
...
+var _ AWSOIDCConfigurations = &awsOIDCConfigurations{}
+
+type AWSOIDCConfiguration struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleARN string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AWSOIDCConfigurationsCreateOptions struct {
```
helenjw: Since we are creating a singular AWS OIDC config:
```suggestion
type AWSOIDCConfigurationCreateOptions struct {
```

Ditto for other places too

hyok_configuration.go
```diff
@@ -0,0 +1,334 @@
...
+	// GCP
+	KeyLocation string `jsonapi:"attr,key-location,omitempty"`
+	KeyRingID   string `jsonapi:"attr,key-ring-id,omitempty"`
+}
+
+type HYOKConfiguration struct {
+	ID string `jsonapi:"primary,hyok-configurations"`
+
+	// Attributes
+	KekID      string                  `jsonapi:"attr,kek-id"`
```
helenjw: Key Encryption Key:
```suggestion
	KEKID      string                  `jsonapi:"attr,kek-id"`
```

helenjw:
> it's interesting that looking at the`hyok_configuration.rb` in Atlas project (/Users/iurislywitch/Desktop/atlas/app/models/hyok_configuration.rb), we have different initialisms...
> 
> ```
> sig { returns(T::Boolean) }
> def aws_configured?
>   oidc_configuration_type == 'AwsOidcConfiguration'
> end
> 
> sig { returns(T::Boolean) }
> def gcp_configured?
>   oidc_configuration_type == 'GcpOidcConfiguration'
> end
> 
> sig { returns(T::Boolean) }
> def azure_configured?
>   oidc_configuration_type == 'AzureOidcConfiguration'
> end
> 
> sig { returns(T::Boolean) }
> def vault_configured?
>   oidc_configuration_type == 'VaultOidcConfiguration'
> end
> ```

@iuri-slywitch-hashicorp I see your point, however I don't think Atlas conventions are applicable here. Again, we already have examples of having same-case initialisms in this repo (see: `RSAPublicKey`, `GPGKey`, and `SSHKey`).

agent_pool.go
```diff
@@ -58,9 +58,10 @@ type AgentPool struct {
 	OrganizationScoped bool   `jsonapi:"attr,organization-scoped"`
 
 	// Relations
-	Organization      *Organization `jsonapi:"relation,organization"`
-	Workspaces        []*Workspace  `jsonapi:"relation,workspaces"`
-	AllowedWorkspaces []*Workspace  `jsonapi:"relation,allowed-workspaces"`
+	Organization       *Organization        `jsonapi:"relation,organization"`
+	HYOKConfigurations []*HYOKConfiguration `jsonapi:"relation,hyok-configurations"`
+	Workspaces         []*Workspace         `jsonapi:"relation,workspaces"`
+	AllowedWorkspaces  []*Workspace         `jsonapi:"relation,allowed-workspaces"`
```
helenjw: I see that you made some changes to:

1.  `agent_pool.go`
2. `workspace.go`
3. `organization.go`

There's nothing wrong with the changes in there, but I encourage you to put them up a separate PR since they address https://hashicorp.atlassian.net/browse/TF-27661. We want to avoid making the PR overly large and make it hard to keep track of what each PR implements.

errors.go
```diff
@@ -399,4 +403,42 @@ var (
...
+
+	ErrRequiredSubscriptionID = errors.New("subscription-id is required for Azure OIDC configuration")
+
+	ErrRequiredTenantID = errors.New("tenant-id is required for Azure OIDC configuration")
+
+	ErrRequiredAddress = errors.New("address is required for Vault OIDC configuration")
+
+	ErrRequiredRoleName = errors.New("role is required for Vault OIDC configuration")
+
+	ErrRequiredJWTAuthPath = errors.New("auth_path is required for Vault OIDC configuration")
```
helenjw: A couple of these are unreferenced. Do we still need them?

hyok_configuration.go
```diff
@@ -0,0 +1,334 @@
...
+	HYOKConfigurationsIncludeOIDCCconfiguration      HYOKConfigurationsIncludeOpt = "oidc_configuration"
+)
+
+type HYOKConfigurationsListOptions struct {
+	ListOptions
+	ProjectID    string                         `url:"filter[project[id]],omitempty"`
+	Sort         HYOKConfigurationsSortColumn   `url:"sort,omitempty"`
+	SearchByName string                         `url:"search[name],omitempty"`
+	Include      []HYOKConfigurationsIncludeOpt `url:"include,omitempty"`
+}
```
helenjw: I'm not sure that we implement listing it by `ProjectID`, `Sort`, or `SearchByName`. The only thing I know we do is using the `q` parameter to get hyok configs with names that kinda match

helenjw (dismissed):
Please add integration tests to your PR! Creating, getting/listing, updating, and deleting OIDC and HYOK configs should be possible with what you've implemented.

If not, adding some basic tests like the one in `admin_setting_cost_estimation_integration_test.go` would be nice too.

agent_pool.go
iuri-slywitch-hashicorp: I'll make a separate PR for those. You're right, it will definitely make it easier to review my code. I initially included these changes in my PR because they weren't big enough, but I agree.

hyok_configuration.go
iuri-slywitch-hashicorp: Fixed it.

errors.go
iuri-slywitch-hashicorp: I'm checking with Jarrett. This is what I have so far:
Vault:
address -- 𝑪𝓤
role -- 𝑪𝓤
auth_path
encoded_cacert
Azure
client-id -- 𝑪𝓤
subscription-id -- 𝑪𝓤
tenant-id -- 𝑪𝓤
GCP:
service-account-email -- 𝑪𝓤
project-number -- 𝑪𝓤
workload-provider-name -- 𝑪𝓤
AWS:
role-arn -- 𝑪𝓤

𝑪: I have validations in create options.
𝓤: I have validations in update options.

Just checking for **auth_path** and **encoded_cacert** if any validations would be necessary.

hyok_configuration.go
iuri-slywitch-hashicorp: Fixed it.

hyok_configuration.go
helenjw: Hmm, gave this a thought. Just because an attribute is missing doesn't mean it's bad. It could just mean they didn't want to change it. I'm fine with leaving it as-is.

aws_oidc_configuration.go
```diff
@@ -0,0 +1,136 @@
+package tfe
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+const OIDCConfigPathFormat = "oidc-configurations/%s"
```
helenjw, line 9: Added this because I was getting dinged by the `goconst` lnter, since this string shows up 12 times.

hyok_configuration_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+
+// These tests are intended for local execution only, as HYOK requires specific conditions.
+// To run them locally, set skipTest to false. The default value is true.
+const skipTest = true
+
+// organizationName is the name of an organization that can use HYOK.
+const organizationName = ""
+
+// agentPoolID represents an agent pool with running agents that have HYOK capabilities turned on.
+const agentPoolID = ""
```
helenjw: In every other integration test i used `skipIfEnterprise`, but here I used a different approach. My reasoning is that HYOK requires:
| Organization 	| Agents 	|
|---	|---	|
| GA Feature flag 	| -hyok flag on 	|
| premium 	|  	|

While other tests can create and delete orgs on the fly, and also agent pools, it's not going to work for these tests. It's hard to turn on a FF for a temp org, and also have running agents. This is important because if the agent isn't running, then we can't revoke and delete the HYOK config, leaving dangling resources each time these tests are run.

Please let me know if there's a better way of approaching this!

Edit: actually it seems like the tests with `skipIfEnterprise` are not being skipped in the CI. Might have to skip it forcefully

helenjw: Also, I opted not to test the `Test` function, because these are dummy values and we're expecting it to return an error anyways. There's really no point of expecting an error since we don't know if the error means the call worked or not 🤷‍♀️

aws_oidc_configuration.go
```diff
@@ -0,0 +1,136 @@
...
+
+type AWSOIDCConfiguration struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleARN string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AWSOIDCConfigurationCreateOptions struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
```
JarrettSpiker: It looks like we [most commonly use `Type` ](https://github.com/hashicorp/go-tfe/pull/190/files)as the primary field in these Create and Update options (though I'll admit that Im a bit confused as to why it needs to be specified at all)

aws_oidc_configuration.go
```diff
@@ -0,0 +1,136 @@
...
+	RoleARN string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AWSOIDCConfigurationUpdateOptions struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleARN string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker: I dont think that we want the `Organization` here: the update options should only include fields which can actually be updated by the user, and we dont allow you to re-specify the organization relationship on the oidc configuration.

```suggestion
```

aws_oidc_configuration.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+type AWSOIDCConfigurationCreateOptions struct {
+	ID      string `jsonapi:"primary,aws-oidc-configurations"`
+	RoleARN string `jsonapi:"attr,role-arn"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker: Similarly, I dont think we want to specify the organization in the create options: the organization gets determined by the organization name in the URL path (specified as one of the arguments to `Create` in go-tfe). Including it here could be misleasing
```suggestion
```

aws_oidc_configuration_integration_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestAWSOIDCConfigurationCreate(t *testing.T) {
+	skipIfEnterprise(t)
+
+	client := testClient(t)
+	ctx := context.Background()
+
+	orgTest, orgTestCleanup := createOrganization(t, client)
```
JarrettSpiker: I would expect that these tests would fail against an organization which is not premium tier. Against TFC (which must be the case, because `skipIfEnterprise`) I think this creates a free tier organization, so I would expect the following tests to fail?

Aside from that, another barrier is that the feature flag would need to be enabled for oidc configuration APIs to work. This typically isn't an issue in go-tfe tests b/c we run them against TFE for CI, but again I would expect these to always fail because the organization is being created right here, and so cannot have the feature flag enabled.

I was expecting that we would need to use something more strict than `skipIfEnterprise`, and allow the user to hard code in an organization name when running this test. Or we would need to create an organization, upgrade it to premium, and then rely on the user defaulting feature flags to true in the environment they run this tests against (which would mean they could only run against local development instances)

aws_oidc_configuration_integration_test.go
```diff
@@ -0,0 +1,126 @@
...
+		opts := AWSOIDCConfigurationUpdateOptions{
+			RoleARN: "arn:aws:iam::123456789012:role/some-role-2",
+			Organization: &Organization{
+				Name: orgTest.Name,
+			},
+		}
+		updated, err := client.AWSOIDCConfigurations.Update(ctx, oidcConfig.ID, opts)
+		require.NoError(t, err)
+		require.NotEmpty(t, updated)
+		assert.NotEqual(t, oidcConfig.RoleARN, updated.RoleARN)
```
JarrettSpiker: I would recommend checking that the new value of `updated.RoleARN` is what you set it to in the `UpdateOptions`

azure_oidc_configuration.go
```diff
@@ -0,0 +1,156 @@
...
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type AzureOIDCConfigurationCreateOptions struct {
+	ID             string `jsonapi:"primary,azure-oidc-configurations"`
+	ClientID       string `jsonapi:"attr,client-id"`
+	SubscriptionID string `jsonapi:"attr,subscription-id"`
+	TenantID       string `jsonapi:"attr,tenant-id"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

azure_oidc_configuration.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type AzureOIDCConfigurationUpdateOptions struct {
+	ID             string `jsonapi:"primary,azure-oidc-configurations"`
+	ClientID       string `jsonapi:"attr,client-id"`
+	SubscriptionID string `jsonapi:"attr,subscription-id"`
+	TenantID       string `jsonapi:"attr,tenant-id"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

azure_oidc_configuration.go
```diff
@@ -0,0 +1,156 @@
...
+		return ErrRequiredClientID
+	}
+
+	if o.SubscriptionID == "" {
+		return ErrRequiredSubscriptionID
+	}
+
+	if o.TenantID == "" {
+		return ErrRequiredTenantID
+	}
```
JarrettSpiker: I believe that this is going to enforce that all options (client, subscription, and tenant) are specified during update: that is not what we want! The user should be able to specify only the options which they want to update.

I'd recommend looking at something like the `ProjectUpdateOptions`
```
type ProjectUpdateOptions struct {
	Name *string `jsonapi:"attr,name,omitempty"`
	Description *string `jsonapi:"attr,description,omitempty"`
	TagBindings []*TagBinding `jsonapi:"relation,tag-bindings,omitempty"`
	AutoDestroyActivityDuration jsonapi.NullableAttr[string] `jsonapi:"attr,auto-destroy-activity-duration,omitempty"`
}
```

the types of the attributes are pointers, which allows them to be `nil` and unspecified. We also `omitempty` to prevent the serialized request to HCP Terraform from containing `description = ""` if the user doesn't specify anything

gcp_oidc_configuration.go
```diff
@@ -0,0 +1,156 @@
...
+	Organization *Organization `jsonapi:"relation,organization"`
+}
+
+type GCPOIDCConfigurationCreateOptions struct {
+	ID                   string `jsonapi:"primary,gcp-oidc-configurations"`
+	ServiceAccountEmail  string `jsonapi:"attr,service-account-email"`
+	ProjectNumber        string `jsonapi:"attr,project-number"`
+	WorkloadProviderName string `jsonapi:"attr,workload-provider-name"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

gcp_oidc_configuration.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type GCPOIDCConfigurationUpdateOptions struct {
+	ID                   string `jsonapi:"primary,gcp-oidc-configurations"`
+	ServiceAccountEmail  string `jsonapi:"attr,service-account-email"`
+	ProjectNumber        string `jsonapi:"attr,project-number"`
+	WorkloadProviderName string `jsonapi:"attr,workload-provider-name"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

azure_oidc_configuration.go
JarrettSpiker: I won't comment on each instance, but this is also true for the other OIDC configuration `Update` functions and `UpdateOptions`...minus maybe AWS which only has one attribute

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+)
+
+type HYOKConfigurations interface {
+	List(ctx context.Context, organization string, options *HYOKConfigurationsListOptions) (*HYOKConfigurationsList, error)
+
+	Create(ctx context.Context, organization string, options HYOKConfigurationsCreateOptions) (*HYOKConfiguration, error)
+
+	Read(ctx context.Context, hyokID string, options *HYOKConfigurationsReadOptions) (*HYOKConfiguration, error)
+
+	Update(ctx context.Context, hyokID string, options HYOKConfigurationsUpdateOptions) (*HYOKConfiguration, error)
```
JarrettSpiker, line 16: ❓ why are the Create and Update options structs, but the List and Read options pointers?

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+	HYOKConfigurationUntested   HYOKConfigurationStatus = "untested"
+	HYOKConfigurationTesting    HYOKConfigurationStatus = "testing"
+	HYOKConfigurationTestFailed HYOKConfigurationStatus = "test_failed"
+	HYOKConfigurationAvailable  HYOKConfigurationStatus = "available"
+	HYOKConfigurationErrored    HYOKConfigurationStatus = "errored"
+	HYOKConfigurationRevoking   HYOKConfigurationStatus = "revoking"
+	HYOKConfigurationRevoked    HYOKConfigurationStatus = "revoked"
+)
+
+type OIDCConfigurationType struct {
```
JarrettSpiker:
```suggestion
type OIDCConfigurationTypeChoice struct {
```

by convention we call these `polyrelation` structs `Choice` types, so the the user knows that they can expect this to have one populated field, corresponding to the actual type of the relationship

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+
+type HYOKConfiguration struct {
+	ID string `jsonapi:"primary,hyok-configurations"`
+
+	// Attributes
+	KEKID      string                  `jsonapi:"attr,kek-id"`
+	KMSOptions *KMSOptions             `jsonapi:"attr,kms-options,omitempty"`
+	Name       string                  `jsonapi:"attr,name"`
+	Primary    bool                    `jsonapi:"attr,primary"`
+	Status     HYOKConfigurationStatus `jsonapi:"attr,status"`
```
JarrettSpiker, line 68: ❓ what happens if `attr,status` doesnt resolve to one of the `HYOKConfigurationStatus` constants when we try to deserialize? I like the idea of the constants, but I want to make sure that we dont break go-tfe versions if we ever add a new status to hyok configurations.

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+type HYOKConfigurationsCreateOptions struct {
+	ID string `jsonapi:"primary,hyok-configurations"`
+
+	// Attributes
+	KEKID      string      `jsonapi:"attr,kek-id"`
+	KMSOptions *KMSOptions `jsonapi:"attr,kms-options"`
+	Name       string      `jsonapi:"attr,name"`
+
+	// Relationships
+	Organization      *Organization          `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+	ID string `jsonapi:"primary,hyok-configurations"`
+
+	// Attributes
+	KEKID      string      `jsonapi:"attr,kek-id"`
+	KMSOptions *KMSOptions `jsonapi:"attr,kms-options"`
+	Name       string      `jsonapi:"attr,name"`
+	Primary    bool        `jsonapi:"attr,primary"`
+
+	// Relationships
+	Organization      *Organization          `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+	OIDCConfiguration *OIDCConfigurationType `jsonapi:"polyrelation,oidc-configuration"`
+	AgentPool         *AgentPool             `jsonapi:"relation,agent-pool"`
+}
+
+func (h *HYOKConfigurationsListOptions) valid() error {
+	return nil
+}
```
JarrettSpiker: ❓ do we need to specify an empty `valid`, or can we just not perform validation?

hyok_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+
+	Create(ctx context.Context, organization string, options HYOKConfigurationsCreateOptions) (*HYOKConfiguration, error)
+
+	Read(ctx context.Context, hyokID string, options *HYOKConfigurationsReadOptions) (*HYOKConfiguration, error)
+
+	Update(ctx context.Context, hyokID string, options HYOKConfigurationsUpdateOptions) (*HYOKConfiguration, error)
+
+	Delete(ctx context.Context, hyokID string) error
+
+	Test(ctx context.Context, hyokID string) error
```
JarrettSpiker, line 22: It could be nice to add a comment explaining how this function behaves: i.e if there is no error, than a successful test was performed.

hyok_configuration_integration_test.go
JarrettSpiker: > if the agent isn't running, then we can't revoke and delete the HYOK config

I wonder if that is a bug in the HCP Terraform implementation? I would expect that you could revoke and delete an hyok configuration without an agent running as long as it doesnt have any key versions.

>While other tests can create and delete orgs on the fly

I'm not sure that's true...I could be missing something, but I think the other tests added in this PR are going to have the same problem, since they also require premium tier and the feature flag, right?

hyok_configuration_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+
+		// Must revoke and delete HYOK config or else agent pool and OIDC configs cannot be cleaned up
+		err = client.HYOKConfigurations.Revoke(ctx, created.ID)
+		if err != nil {
+			require.NoError(t, err)
+		}
+
+		fetched, err := client.HYOKConfigurations.Read(ctx, created.ID, nil)
+		require.NoError(t, err)
+		assert.True(t, fetched.Status == HYOKConfigurationRevoked || fetched.Status == HYOKConfigurationRevoking)
```
JarrettSpiker: a bit of an unfortunate detail: the status could be Revoking, which would make the delete fail. Maybe we need the cleanup for these tests to wait (with some max timeout) for the configuration to become revoked before deleting it?

vault_oidc_configuration.go
```diff
@@ -0,0 +1,154 @@
...
+
+type VaultOIDCConfigurationCreateOptions struct {
+	ID               string `jsonapi:"primary,vault-oidc-configurations"`
+	Address          string `jsonapi:"attr,address"`
+	RoleName         string `jsonapi:"attr,role"`
+	Namespace        string `jsonapi:"attr,namespace"`
+	JWTAuthPath      string `jsonapi:"attr,auth-path"`
+	TLSCACertificate string `jsonapi:"attr,encoded-cacert"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

vault_oidc_configuration.go
```diff
@@ -0,0 +1,154 @@
...
+
+type VaultOIDCConfigurationUpdateOptions struct {
+	ID               string `jsonapi:"primary,vault-oidc-configurations"`
+	Address          string `jsonapi:"attr,address"`
+	RoleName         string `jsonapi:"attr,role"`
+	Namespace        string `jsonapi:"attr,namespace"`
+	JWTAuthPath      string `jsonapi:"attr,auth-path"`
+	TLSCACertificate string `jsonapi:"attr,encoded-cacert"`
+
+	Organization *Organization `jsonapi:"relation,organization"`
```
JarrettSpiker:
```suggestion
```

aws_oidc_configuration.go
helenjw: Removed all the org relations in 6f8cc7ba22947e3509571248efcb65a8b3c20eda!

hyok_configuration.go
helenjw: ✅ Done in 71baeacd0abcfea6119e9112f6abe1cbf9bdfaac

hyok_configuration.go
helenjw: I don't think we do, but I think I've seen a few other places with empty validaiton. Nonetheless, I've removed it! 99af76e08096140e05b64f8e7d12937a2806814a

hyok_configuration.go
helenjw: Doneso -> 2af481df2f0d986362d46d63af66db0ed21111cf

aws_oidc_configuration_integration_test.go
helenjw: That's so valid 5c3db3321872763fe147e4a4b310a8869f2177a1

azure_oidc_configuration.go
helenjw: Ohh thanks for catching this! I was wondering if the fields inside `KMSOptions` should similarly all be pointers to allow for `nil` as well?

Edit: actually i tried this and it was making tests fail and i had no idea how to fix it so i reverted specifically the `KMSOptions` changes in 0bf5344d7e8ea4ecce457f6bdd0c453bfea31f6a

helenjw: These changes were made in 5d94db18cdb68a629750f9a521900eec00eb9022

hyok_configuration_integration_test.go
helenjw: > While other tests can create and delete orgs on the fly

Oh yes sorry I meant this is the pattern i see for other tests in the repo.

> I would expect that you could revoke and delete an hyok configuration without an agent running as long as it doesnt have any key versions.

~actually let me double check on this and report back~

Running these tests, it fails on revoke since there's no running agents
image

aws_oidc_configuration_integration_test.go
helenjw: Yeah, all valid points. I think I'm going to do the same thing that I did in `hyok_configuration_integration_test.go` for all oidc tests.

See 0bf5344d7e8ea4ecce457f6bdd0c453bfea31f6a

hyok_configuration.go
helenjw: I actually have no clue 😭 It is something that's seen in the example in `CONTRIBUTING.md`

hyok_configuration.go
helenjw: I think it should be okay - I tested by retrieving an HYOK configuration that's `available`, but commented out the `available` status. I was still able to get it to pass this test:
```go
assert.Equal(t, HYOKConfigurationStatus("available"), fetched.Status)
```

I think any string in the `status` field just becomes of type `HYOKConfigurationStatus`

hyok_configuration_integration_test.go
helenjw: that is unfortunate 😭 fortunately it has been addressed f5a0fdda0cf1c2f9eaed938cf82fb237dd3aa47d

aws_oidc_configuration.go
helenjw: Oh yes, thanks for catching this. I updated all of them in db238370ecff7b2f794e4f2bba58fe0e7d006d5b

helper_test.go
```diff
@@ -569,6 +569,337 @@ func createGPGKey(t *testing.T, client *Client, org *Organization, provider *Reg
...
+		if err := client.HYOKConfigurations.Delete(ctx, hyokConfig.ID); err != nil {
+			t.Errorf("Error removing HYOK Configuration! WARNING: Dangling resources\n"+
+				"may exist! The full error is shown below.\n\n"+
+				"HYOKConfigurations: %s\nError: %s", hyokConfig.ID, err)
+		}
+
+		if orgCleanup != nil {
+			orgCleanup()
+		}
+	}
```
JarrettSpiker: I like the idea behind having separate `createHYOKConfiguration` functions per OIDC config type, but this non-trivial cleanup function gets duplicated in each one

What do you think about creating a shared `cleanupHyokConfig` function which could get returned here instead?

hyok_configuration.go
JarrettSpiker: huh, weird. I would probably make them all consistent, but it is not a big deal...we dont seem to have a consistent pattern across the codebase anyway

aws_oidc_configuration_integration_test.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("with valid options", func(t *testing.T) {
+		opts := AWSOIDCConfigurationCreateOptions{
+			RoleARN: "arn:aws:iam::123456789012:role/some-role",
+		}
+
+		oidcConfig, err := client.AWSOIDCConfigurations.Create(ctx, orgTest.Name, opts)
```
JarrettSpiker, line 32: General suggestion for all of the OIDC `create` tests: It doesnt look like we clean up the oidc configurations created by the test. Most -go-tfe tests get away with this by deleting the org at the end of the test, but we dont have that option available.

What do you think about adding a helper to `t.Cleanup` an oidc configuration based on its ID, and using it in these tests?

aws_oidc_configuration_integration_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	t.Run("delete existing configuration", func(t *testing.T) {
+		err := client.AWSOIDCConfigurations.Delete(ctx, oidcConfig.ID)
+		require.NoError(t, err)
+	})
+
+	t.Run("fetching non-existing configuration", func(t *testing.T) {
+		err := client.AWSOIDCConfigurations.Delete(ctx, "awsoidc-notreal")
+		require.ErrorIs(t, err, ErrResourceNotFound)
+	})
```
JarrettSpiker: General suggestion for all of these Delete tests: try to keep in mind whether we are testing go-tfe functionality, or HCP Terraform functionality.

For example, what we are actually testing is that HCP Terraform 404s if we fetch a non-existent configuration. We're not really testing any go-tfe functionality (other than its ability to return the error).
I think it's fine to leave this test here, because it is pretty basic HCP Terraform functionality, but it is probably not necessary to add this sort of test in the future.

aws_oidc_configuration_integration_test.go
JarrettSpiker: Alternatively, could combine the Create and Delete tests so that we delete the configuration which gets created, as part of the test

JarrettSpiker (commented):
Overall looks good, thanks for all the work on this! HYOK definitely take a bit extra...

I've just left some comments with testing suggestions, but Ill approve, and I'd recommend reaching out to the core-cloud team to request a code owners review as well

hyok_configuration_integration_test.go
JarrettSpiker: I am going to file a bug for that for the HCP Terraform side: we shouldnt require that you have an agent running in order to revoke a configuration with no key versions

JarrettSpiker: https://hashicorp.atlassian.net/browse/TF-28619

JarrettSpiker: I actually see several errors running the tests, and I believe they are all due to configurations not being revoked during cleanup
CleanShot 2025-08-25 at 16 22 36@2x

It would be nice to figure out why the retryable revocation isnt working for me, but Id also suggest that a lot of the issue goes away if we just fix that HCP Terraform bug

helper_test.go
helenjw: Great idea! Done!

aws_oidc_configuration_integration_test.go
helenjw: Oh good catch, I combined the delete and create so everything is cleaned up ab0100e14895fa47533586702e823d53526056d1

aws_oidc_configuration_integration_test.go
helenjw: Gotcha, I deleted those tests in ab0100e14895fa47533586702e823d53526056d1

hyok_configuration_integration_test.go
JarrettSpiker: Tests have been fixed with Helen's latest change

Maed223 (approved):
Looks good!

Issue #1061, opened 2025-02-20
ERROR: invalid value "cost_estimated" for run status


Olgoetz:
#### go-tfe version

```plaintext
1.75.0
...
```

## Description

Getting runs with the status `cost_estimated` leads to 
"invalid value "cost_estimated" for run status"
altough it is a supported state:
https://pkg.go.dev/github.com/hashicorp/go-tfe#RunStatus

## Testing plan

```plaintext
Executing this function leads to the above mentioned error:


log.Println("Listing all 'cost_estimated' runs...")
adminRunsListOptions.RunStatus = "cost_estimated"
costEstimatedRuns, err := tfe_client.Admin.Runs.List(context.Background(), &adminRunsListOptions)
if err != nil {
	log.Fatal(err)
}
log.Printf("Found %s runs in 'cost_estimated' state", strconv.Itoa(len(costEstimatedRuns.Items)))
...
```

#### Expected Behavior

Getting all runs with 'cost_estimated' state

#### Actual Behavior

invalid value "cost_estimated" for run status

KenCox-Hashicorp:
This will be fixed by #1171 in the next release.

PR #1170, opened 2025-08-01
Fix validation of run status; add unit test [issue 1061]

KenCox-Hashicorp:
## Description

Fixes https://github.com/hashicorp/go-tfe/issues/1061.  Note, that was reported in v1.75, the bug is still in head of code v1.88, so this may need backport.

`validateAdminRunFilterParams` in admin_run.go was using `RunCostEstimate`, a `RunIncludeOpt`. It should be using `RunCostEstimated`, a `RunStatus`. Adds unit test.

Some of the `RunStatus` values are not accepted by `validateAdminRunFilterParams` - this PR assumes that is intentional.

**Note:** Merge either this PR or #1171, which updates the validate function to add the missing RunStatus values.

## Testing plan

Code given in the issue should run, since "cost_estimated" is a RunStatus. This now works:

```
func main() {
                Address:           "...",
                RetryServerErrors: true,
                Token:             "...",
        })
        if err != nil {
                log.Fatal("Error %v\n", err)
                os.Exit(1)
        }
        costEstimatedRuns, err := client.Admin.Runs.List(context.Background(), &tfe.AdminRunsListOptions{
                RunStatus: "cost_estimated",
        })
        if err != nil {
                log.Fatal(err)
        }
        log.Printf("Found %v runs in 'cost_estimated' state", len(costEstimatedRuns.Items))
}
```

## Related Links

- ILM team [JIRA](https://hashicorp.atlassian.net/browse/TF-24176)

## Output from tests

New unit test does not require TFE instance.

## Rollback Plan

Revert the PR.

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 * Adds `CreatedAt` and `UpdatedAt` fields to `StackConfiguration` by @Maed223 [#1168](https://github.com/hashicorp/go-tfe/pull/1168)
 * Adds endpoint for advancing `StackDeploymentStep` by @hwatkins05-hashicorp [#1166](https://github.com/hashicorp/go-tfe/pull/1166)
 
+## Bug Fixes
+* Fixes issue [1061](https://github.com/hashicorp/go-tfe/issues/1061), validation accepts `"cost_estimated"`, by @KenCox-Hashicorp [#1170](https://github.com/hashicorp/go-tfe/pull/1170)
+
 # v1.88.0
 
 ## Enhancements
--- a/admin_run.go
+++ b/admin_run.go
@@ -167,7 +167,7 @@ func validateAdminRunFilterParams(runStatus string) error {
 				string(RunApplying),
 				string(RunCanceled),
 				string(RunConfirmed),
-				string(RunCostEstimate),
+				string(RunCostEstimated),
 				string(RunCostEstimating),
 				string(RunDiscarded),
 				string(RunErrored),
--- a/admin_run_test.go
+++ b/admin_run_test.go
@@ -1,0 +1,73 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tfe
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_validateAdminRunFilterParams(t *testing.T) {
+	// RunStatus values accepted by validate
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Adds `Completed` status to `StackConfiguration` by @hwatkins05-hashicorp [#1163](https://github.com/hashicorp/go-tfe/pull/1163)
 * Adds `CreatedAt` and `UpdatedAt` fields to `StackConfiguration` by @Maed223 [#1168](https://github.com/hashicorp/go-tfe/pull/1168)
 
+## Bug Fixes
+* Fixes issue [1061](https://github.com/hashicorp/go-tfe/issues/1061), RunStatus value `"cost_estimated"` by @KenCox-Hashicorp [#1170](https://github.com/hashicorp/go-tfe/pull/1170)
+
```
brandonc: Can you move this to the "unreleased" section?

brandonc (commented):
Thanks for the fix!!

CHANGELOG.md
KenCox-Hashicorp: Will do.  

Is this the PR you want, or #1171 - that one adds all the missing RunStatus values to the validation.

admin_run.go
```diff
@@ -167,7 +167,7 @@ func validateAdminRunFilterParams(runStatus string) error {
 				string(RunApplying),
 				string(RunCanceled),
 				string(RunConfirmed),
-				string(RunCostEstimate),
+				string(RunCostEstimated),
```
notchairmk, line 170: can we remove the `RunCostEstimate` const if it's not being used anywhere?

notchairmk (approved):
one minor question, but overall looks good to me

admin_run.go
KenCox-Hashicorp: It's defined as a RunIncludeOpt [here](https://github.com/hashicorp/go-tfe/blob/b35229217714b43cd438597eecb785fdfefd8fe1/run.go#L219-L230). I believe those options are used when filtering runs, at least, they (mostly) match the list in `mirage/routes/runs.js` [here](https://github.com/hashicorp/atlas/blob/5e8659c8005466f7fc5b905983b211072f03a13e/frontend/atlas/mirage/routes/runs.js#L9-L20). 

Possibly RunIncludeOpt needs to be updated to add the missing value `workspace.project`, in a separate PR.

notchairmk: ah yep makes sense. thanks for the explanation!

PR #1171, opened 2025-08-01
Fix validation of run status; add unit test [issue 1061]

KenCox-Hashicorp:
## Description

Fixes https://github.com/hashicorp/go-tfe/issues/1061.  Note, that was reported in v1.75, the bug is still in head of code v1.88, so this may need backport.

`validateAdminRunFilterParams` in admin_run.go was using `RunCostEstimate`, a `RunIncludeOpt`. It should be using `RunCostEstimated`, a `RunStatus`. Adds unit test.

Some of the RunStatus values were not accepted by validateAdminRunFilterParams - this PR assumes that was a mistake. The validate function has been fixed to accept all status values.

**Note:** Merge either this PR or #1170, which does _not_ update the validate function with the missing RunStatus values.

## Testing plan

Code given in the issue should run, since "cost_estimated" is a RunStatus. This now works:

```
func main() {
                Address:           "...",
                RetryServerErrors: true,
                Token:             "...",
        })
        if err != nil {
                log.Fatal("Error %v\n", err)
                os.Exit(1)
        }
        costEstimatedRuns, err := client.Admin.Runs.List(context.Background(), &tfe.AdminRunsListOptions{
                RunStatus: "cost_estimated",
        })
        if err != nil {
                log.Fatal(err)
        }
        log.Printf("Found %v runs in 'cost_estimated' state", len(costEstimatedRuns.Items))
}
```

## Output from tests

New unit test does not require TFE instance.

## Rollback Plan

Revert the PR.

None.


brandonc (commented):
Thank you for adding these. Before I merge it, it looks like the CHANGELOG entry is still in the wrong place (under 1.89.0) -- Can you move it to underneath # Unrelelased?

PR #1179, opened 2025-08-11
Add support for searching modules in private registry

gautambaghel:
* search modules in private registry by name
* Filter modules in private registry based on `provider`, `registry_name` and `organization_name`

## Description

* Currently the `RegistryModuleListOptions` only includes limit options for pagination, this would allow users to query modules better
* This is needed for the Terraform MCP server tools for private modules

## External links

* [API Docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/private-registry/modules#query-parameters)

## Output from tests

```

...
```

### DIFF
--- a/registry_module.go
+++ b/registry_module.go
@@ -277,7 +277,7 @@ type RegistryModuleListOptions struct {
 	Provider string `url:"filter[provider],omitempty"`
 
 	// RegistryName filters results by registry name (public or private)
-	RegistryName string `url:"filter[registry_name],omitempty"`
+	RegistryName RegistryName `url:"filter[registry_name],omitempty"`
 
 	// OrganizationName filters results by organization name
 	OrganizationName string `url:"filter[organization_name],omitempty"`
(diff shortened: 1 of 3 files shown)


registry_module.go
```diff
@@ -269,6 +269,18 @@ type RegistryModuleListOptions struct {
...
 	Include []RegistryModuleListIncludeOpt `url:"include,omitempty"`
+
+	// Search is a search query string. Modules are searchable by name, namespace, provider fields.
+	Search string `url:"q,omitempty"`
+
+	// Provider filters results by provider name
+	Provider string `url:"filter[provider],omitempty"`
+
+	// RegistryName filters results by registry name (public or private)
+	RegistryName string `url:"filter[registry_name],omitempty"`
```
shwetamurali: I noticed the `RegistryName` field in `RegistryModuleListOptions` uses `string` type instead of the `RegistryName` type used in other areas of the file. This may be intentional for query parameter compatibility, but wanted to confirm in case you weren't aware we had the attribute already!

shwetamurali (dismissed):

registry_module.go
gautambaghel: yup, you're correct

shwetamurali (approved):

PR #1184, opened 2025-08-13
StackConfiguration's `CreateAndUpload` 

Maed223:
## Description

For Stacks GA the `CreateAndUpload` endpoints are now on the stack configuration, this PR makes the needed addition to support this in go-tfe.

## Testing plan

Added new test for associated endpoint.

## Output from tests

Screenshot 2025-08-13 at 4 03 20 PM

### DIFF
--- a/stack_configuration.go
+++ b/stack_configuration.go
@@ -195,7 +195,7 @@ func (s stackConfigurations) CreateAndUpload(ctx context.Context, stackID, path 
 		return nil, err
 	}
 
-	err = s.UploadTarGzip(ctx, uploadURL, body)
+	err = s.uploadTarGzip(ctx, uploadURL, body)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


stack_configuration.go
```diff
@@ -155,3 +161,95 @@ func (s stackConfigurations) List(ctx context.Context, stackID string, options *
...
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving upload URL for stack configuration %q: %w", sc.ID, err)
+	}
+
+	body, err := packContents(path)
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.UploadTarGzip(ctx, uploadURL, body)
```
brandonc: Do you want this to be a part of the public interface? Because it's public but not an interface method

brandonc (approved):

stack_configuration.go
markdecrane: Ah shoot

markdecrane: fixed

PR #1193, opened 2025-08-21
add sorting to agent list and agent pool list

twitnithegirl:
## Output from tests

```

...
```


brandonc:
We're moving toward a code-generated v2 package. This query parameter can be added to the OpenAPI spec in the HCP Terraform platform code and it will be correctly generated for the relevant interfaces.

PR #1199, opened 2025-08-26
fix naming

shwetamurali:
## Output from tests

```

...
```


stack_deployment_steps.go
```diff
@@ -29,7 +29,7 @@ type StackDeploymentStep struct {
 	ID        string    `jsonapi:"primary,stack-deployment-steps"`
 	Status    string    `jsonapi:"attr,status"`
 	CreatedAt time.Time `jsonapi:"attr,created-at,iso8601"`
-	UpdatedAt time.Time `jsonapi:"attr,created-at,iso8601"`
+	UpdatedAt time.Time `jsonapi:"attr,updated-at,iso8601"`
```
markdecrane, line 32: Nice catch there, I was just about to look into a problem related to this

Maed223 (approved):

PR #1212, opened 2025-09-08
[TFDN-1189] tool version arch validation

kelsi-hoyle:
## Description

After testing [provider updates](https://github.com/hashicorp/terraform-provider-tfe/pull/1762) to support archs on tool versions, a bug was discovered in go-tfe that caused validation to fail if using top level `url` and `sha` with multiple archs. 

This change fixes the validation to allow multiple `archs` and top level `url` and `sha`, validating the `url` and `sha` against the `amd64` arch _only_.

## Output from tests

(Using a admin version maintenance token)
Screenshot 2025-09-08 at 9 09 34 AM
Screenshot 2025-09-08 at 9 10 24 AM
Screenshot 2025-09-08 at 9 11 22 AM

### DIFF
--- a/admin_opa_version.go
+++ b/admin_opa_version.go
@@ -215,10 +215,20 @@ func (o AdminOPAVersionCreateOptions) validArch() bool {
 
 	emptyToolVersionFields := o.URL == "" && o.SHA == ""
 
+	var amdArch *ToolVersionArchitecture
 	for _, a := range o.Archs {
-		if !validArch(a) || !emptyToolVersionFields && (o.URL != a.URL || o.SHA != a.Sha) {
+		if !validArch(a) {
 			return false
 		}
+
+		if a.Arch == "amd64" {
+			amdArch = a
+		}
 	}
+
+	if !emptyToolVersionFields && (amdArch.URL != o.URL || amdArch.Sha != o.SHA) {
+		return false
+	}
+
 	return true
 }
--- a/admin_opa_version.go
+++ b/admin_opa_version.go
@@ -213,21 +213,10 @@ func (o AdminOPAVersionCreateOptions) validArch() bool {
 		return true
 	}
 
-	emptyToolVersionFields := o.URL == "" && o.SHA == ""
-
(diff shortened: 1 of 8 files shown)


admin_sentinel_version_integration_test.go
```diff
@@ -102,24 +102,75 @@ func TestAdminSentinelVersions_CreateDelete(t *testing.T) {
 	client := testClient(t)
 	ctx := context.Background()
 	version := createAdminSentinelVersion()
+	url := "https://www.hashicorp.com"
+	amd64Sha := String(genSha(t))
 
-	t.Run("with valid options including archs", func(t *testing.T) {
+	t.Run("with valid options including top level url & sha and archs", func(t *testing.T) {
```
kelsi-hoyle, line 108: adding test for this specific case 😅

admin_opa_version.go
```diff
@@ -215,10 +215,20 @@ func (o AdminOPAVersionCreateOptions) validArch() bool {
...
+		if !validArch(a) {
 			return false
 		}
+
+		if a.Arch == "amd64" {
+			amdArch = a
+		}
 	}
+
+	if !emptyToolVersionFields && (amdArch.URL != o.URL || amdArch.Sha != o.SHA) {
```
brandonc: This could panic if there is no AMD architecture and the top level is empty...

brandonc: Do you feel confident in the backend's ability to validate the architectures business logic? It seems complicated so we could remove all client-side validation and let the backend return an error if it's incorrect.

kelsi-hoyle: Yeah, I was just realizing this could be nil.

Yes, the API handles this logic and we could just rely on that. @natalie-todd thoughts? you had originally added this validation logic.

natalie-todd (dismissed):
LGTM approved pending @brandonc's review

brandonc (approved):

brandonc (dismissed):
Works for me

PR #1223, opened 2025-09-19
Remove `ForceCancel` from `QueryRuns`

sebasslash:
## Description

This endpoint does not exist, queries are canceled by a single endpoint `Cancel`. 

## Output from tests

```

...
```


brandonc:
Removed from final v1 package in https://github.com/hashicorp/go-tfe/pull/1364

PR #1225, opened 2025-09-23
Add UserTokensEnabled field for Organizations

JarrettSpiker:
## Description

Adds `UserTokensEnabled` for `Organizations`.

This new HCP Terraform setting defaults to true. When disabled by an organization owner, user tokens will no longer be permitted to access the organization's resources through the API.

## Testing plan

1.  Validate that the setting can be read from existing orgs
1.  Test creating orgs with the setting enabled/disabled
1.  Test updating an org to enable/disable the setting
2. Validate that the setting is not included in update request serialization if not specified

## External links

- [Documentation PR]()https://github.com/hashicorp/web-unified-docs/pull/952

## Output from tests

```
/usr/local/bin/go test -timeout 3000s -run ^TestOrganizationsUpdate$ github.com/hashicorp/go-tfe

=== RUN   TestOrganizationsUpdate
=== RUN   TestOrganizationsUpdate/with_HCP_Terraform-only_options
--- PASS: TestOrganizationsUpdate/with_HCP_Terraform-only_options (2.83s)
=== RUN   TestOrganizationsUpdate/with_new_AggregatedCommitStatusEnabled_option
--- PASS: TestOrganizationsUpdate/with_new_AggregatedCommitStatusEnabled_option (5.25s)
=== RUN   TestOrganizationsUpdate/with_new_SpeculativePlanManagementEnabled_option
--- PASS: TestOrganizationsUpdate/with_new_SpeculativePlanManagementEnabled_option (3.65s)
=== RUN   TestOrganizationsUpdate/with_new_UserTokensEnabled_option
--- PASS: TestOrganizationsUpdate/with_new_UserTokensEnabled_option (5.66s)
=== RUN   TestOrganizationsUpdate/with_valid_options
--- PASS: TestOrganizationsUpdate/with_valid_options (3.72s)
=== RUN   TestOrganizationsUpdate/with_invalid_name
--- PASS: TestOrganizationsUpdate/with_invalid_name (0.00s)
=== RUN   TestOrganizationsUpdate/with_agent_pool_provided,_but_remote_execution_mode
--- PASS: TestOrganizationsUpdate/with_agent_pool_provided,_but_remote_execution_mode (3.75s)
=== RUN   TestOrganizationsUpdate/when_only_updating_a_subset_of_fields
--- PASS: TestOrganizationsUpdate/when_only_updating_a_subset_of_fields (2.16s)
=== RUN   TestOrganizationsUpdate/with_different_default_execution_modes
--- PASS: TestOrganizationsUpdate/with_different_default_execution_modes (3.93s)
--- PASS: TestOrganizationsUpdate (31.59s)
PASS
ok      github.com/hashicorp/go-tfe
...
```

## Rollback Plan

If we need to revert this change before a go-tfe release, we will.

no

### DIFF
--- a/organization.go
+++ b/organization.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright IBM Corp. 2018, 2025
 // SPDX-License-Identifier: MPL-2.0
 
 package tfe
--- a/organization_integration_test.go
+++ b/organization_integration_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright IBM Corp. 2018, 2025
 // SPDX-License-Identifier: MPL-2.0
 
 package tfe
--- a/META.d/_summary.yaml
+++ b/META.d/_summary.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) HashiCorp, Inc.
+# Copyright IBM Corp. 2018, 2025
 # SPDX-License-Identifier: MPL-2.0
 
 ---
--- a/META.d/data.yaml
+++ b/META.d/data.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) HashiCorp, Inc.
+# Copyright IBM Corp. 2018, 2025
 # SPDX-License-Identifier: MPL-2.0
 
 data_summary:
--- a/admin_opa_version.go
+++ b/admin_opa_version.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright IBM Corp. 2018, 2025
 // SPDX-License-Identifier: MPL-2.0
 
 package tfe
--- a/admin_opa_version_integration_test.go
+++ b/admin_opa_version_integration_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright IBM Corp. 2018, 2025
 // SPDX-License-Identifier: MPL-2.0
 
 package tfe
--- a/admin_organization.go
+++ b/admin_organization.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright IBM Corp. 2018, 2025
 // SPDX-License-Identifier: MPL-2.0
 
 package tfe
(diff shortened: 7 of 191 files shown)


organization.go
```diff
@@ -118,6 +118,7 @@ type Organization struct {
 	RemainingTestableCount                            int                      `jsonapi:"attr,remaining-testable-count"`
 	SpeculativePlanManagementEnabled                  bool                     `jsonapi:"attr,speculative-plan-management-enabled"`
 	EnforceHYOK                                       bool                     `jsonapi:"attr,enforce-hyok"`
+	UserTokensEnabled                                 *bool                    `jsonapi:"attr,user-tokens-enabled"`
```
ctrombley, line 121: Why do we use a bool pointer type? `false` and `nil` are both logically equivalent, correct?

organization_integration_test.go
```diff
@@ -320,6 +320,45 @@ func TestOrganizationsUpdate(t *testing.T) {
...
+		ownerClient.token = ownerToken.Token
+
+		options := OrganizationUpdateOptions{
+			UserTokensEnabled: Bool(false),
+		}
+
+		org, err := ownerClient.Organizations.Update(ctx, orgTest.Name, options)
+		require.NoError(t, err)
+		assert.False(t, *org.UserTokensEnabled, "user tokens disabled")
+
```
ctrombley, line 354: Consider including assertions that validate the change had the intended effect, i.e. try and look up some resources and expect an error here. (& the inverse when the setting is off)

organization.go
JarrettSpiker: Good question: it is because this setting defaults to `true`. Making this a bool pointer allows us to distinguish between "this is unspecified on the organization" (ie, because it is an older version of TFE) and "this is explicitly false" (ie, because this is an up-to-date TFE and the user has disabled the setting)

That is important because in the provider [when we read an organization](https://github.com/hashicorp/terraform-provider-tfe/pull/1860/files#diff-20e671524a420e4c3d892ee6da80b7c556b5d89671f4e6f7be930d974029a0adR180-R183) from an old TFE version, we dont want the the value to appear as `false`. That could lead to the provider always showing drift for the org and/or be misleading since the behaviour of the old org is that user tokens are enabled.

organization_integration_test.go
JarrettSpiker: thanks for the suggestion, I have added some more assertions!

organization.go
JarrettSpiker: A similar example would be like `OrganizationScoped` for agent pools

PR #1229, opened 2025-10-02
TFDN-1107 add sort option for agent list and agent pool list

twitnithegirl:
## Description

A sort param was added to the agent-pool list a while back but was never added here. I am adding it now. I am also adding the recently added agent list sort option. There is an issue with the CreateAgent helper used int he agent list integration tests so I have not added an integration test for the agent sort option. There is a ticket to debug this helper in our backlog and once that ticket is in flight the test will be added. If you are not comfortable with us adding this option without a test here, let me know and I can just remove that line but I would like to at least get the agent pool option in.

## External links
Here is a link to the Draft PR with the broken test helper for reference: https://github.com/hashicorp/go-tfe/pull/1193/
Here is a link to the ticket for the follow up work on investigating these broken tests: https://hashicorp.atlassian.net/browse/TFDN-1240

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,15 @@
 # Unreleased
 
+## Enhancements
+
+* Add `Sort` option to `Agents` and `AgentPools`, by @twitnithegirl [#1229](https://github.com/hashicorp/go-tfe/pull/1229)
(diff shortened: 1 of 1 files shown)


ctrombley (approved):
Looks good, thanks @twitnithegirl!

twitnithegirl:
@ctrombley can you please re-approve. Fixing the merge conflict negated your approval.

ctrombley (approved):

lafentres (dismissed):

ctrombley (approved):

PR #1233, opened 2025-10-10
tests: attempt to fix the persistent nightly failures

ctrombley:
This PR updates several tests due to persistent nightly failures.

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -1363,6 +1363,17 @@ func createOrganizationMembership(t *testing.T, client *Client, org *Organizatio
 	}
 }
 
+func createPremiumOrganization(t *testing.T, client *Client) (*Organization, func()) {
+	org, orgCleanup := createOrganizationWithOptions(t, client, OrganizationCreateOptions{
+		Name:  String("tst-" + randomString(t)),
+		Email: String(fmt.Sprintf("%s@tfe.local", randomString(t))),
+	})
+
+	newSubscriptionUpdater(org).WithPremiumPlan().Update(t)
+
+	return org, orgCleanup
+}
+
 func createOrganizationToken(t *testing.T, client *Client, org *Organization) (*OrganizationToken, func()) {
 	var orgCleanup func()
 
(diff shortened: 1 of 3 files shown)


helper_test.go
```diff
@@ -1363,6 +1363,17 @@ func createOrganizationMembership(t *testing.T, client *Client, org *Organizatio
 	}
 }
 
+func createPremiumOrganization(t *testing.T, client *Client) (*Organization, func()) {
+	org, orgCleanup := createOrganizationWithOptions(t, client, OrganizationCreateOptions{
+		Name:  String("tst-" + randomString(t)),
+		Email: String(fmt.Sprintf("%s@tfe.local", randomString(t))),
+	})
+
+	newSubscriptionUpdater(org).WithPremiumPlan().Update(t)
```
brandonc, line 1372: It's curious to me that this would be the cause of nightlies failures since it's supposed to be testing against ON_PREM mode-- no subscriptions in TFE!

helper_test.go
```diff
@@ -2741,7 +2752,7 @@ func createWorkspaceWithGithubApp(t *testing.T, client *Client, org *Organizatio
 
 	githubIdentifier := os.Getenv("GITHUB_POLICY_SET_IDENTIFIER")
 	if githubIdentifier == "" {
-		t.Fatal("Export a valid GITHUB_POLICY_SET_IDENTIFIER before running this test!")
+		t.Skip("Export a valid GITHUB_POLICY_SET_IDENTIFIER before running this test")
```
brandonc, line 2755: 🤔 maybe we should just set one? I think it's meant to be a repository but I don't know much more

helper_test.go
ctrombley: Yeah, I'm not sure I understand how to do this then. The error says:
```
 the organization does not have private VCS enabled
 ```
 
 I've seen in docs & elsewhere that that's only available for premium plans.

helper_test.go
ctrombley: Yep, that value should be `hashicorp/test-policy-set`. I'll drop it in the GHA job definition.

ctrombley: Nevermind, it's already there.

PR #1237, opened 2025-10-14
Parallelization of tests

Maed223:
## Description
Introduces parallelization to our CI tests. This was done through an added naming convention `_RunDependent`, and running/skipping by said convention to isolate and sequentially run, run dependent tests.

Build tags were considered as a solution but a notable downside is that build tags are additive, meaning we can only explicitly exclude them and not include them. This would mean that all non run dependent testing files have to be marked to exclude the build tags.

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -1710,6 +1710,7 @@ func readQueryRun(t *testing.T, client *Client, ctx context.Context, r *QueryRun
 
 // applyRun will apply the given run.
 func applyRun(t *testing.T, client *Client, ctx context.Context, r *Run) {
+	runDependentTestNameValidator(t)
 	t.Logf("Applying run %q", r.ID)
 
 	if err := client.Runs.Apply(ctx, r.ID, RunApplyOptions{}); err != nil {
@@ -1758,6 +1759,7 @@ func fatalDumpRunLog(t *testing.T, client *Client, ctx context.Context, run *Run
 }
 
 func createRun(t *testing.T, client *Client, w *Workspace) (*Run, func()) {
+	runDependentTestNameValidator(t)
 	var wCleanup func()
 
 	if w == nil {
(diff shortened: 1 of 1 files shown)


.github/actions/test-go-tfe/action.yml
```diff
@@ -108,7 +109,7 @@ runs:
         GO111MODULE: "on"
         ENABLE_TFE: ${{ inputs.enterprise }}
       run: |
-        gotestsum --junitfile summary.xml --format short-verbose -- -parallel=1 -timeout=59m -coverprofile cover.out -run "${{ steps.test_split.outputs.run }}"
```
ctrombley, line 111: I had a thought, take it or leave it - we might be able to introduce a new env var here that would cause go-tfe tests to fail if they tried to start a run without the `_RunDependent` convention.

.github/workflows/ci.yml
```diff
@@ -92,10 +163,10 @@ jobs:
 
       - name: Merge coverage reports
         run: |
-          gocovmerge junit-test-summary-0/cover.out > merged-coverage.out
+          gocovmerge junit-test-summary-0/cover.out junit-test-summary-1/cover.out junit-test-summary-2/cover.out junit-test-summary-3/cover.out junit-test-summary-4/cover.out junit-test-summary-5/cover.out junit-test-summary-6/cover.out junit-test-summary-7/cover.out junit-run-dependent-test-summary/cover.out > merged-coverage.out
```
ctrombley: Brace expansion might do the job also: `junit-test-summary-{0..7}/cover.out`

.github/workflows/ci.yml
```diff
@@ -92,10 +163,10 @@ jobs:
...
 
       - name: Merge junit reports
-        run: jrm ./ci-summary.xml "junit-test-summary-0/*.xml"
+        run: jrm ./ci-summary.xml "junit-test-summary-0/*.xml" "junit-test-summary-1/*.xml" "junit-test-summary-2/*.xml" "junit-test-summary-3/*.xml" "junit-test-summary-4/*.xml" "junit-test-summary-5/*.xml" "junit-test-summary-6/*.xml" "junit-test-summary-7/*.xml" "junit-run-dependent-test-summary/*.xml"
```
ctrombley: Same here: `junit-test-summary-{0..7}/*.xml`.

ctrombley (dismissed):
Nice work @Maed223! This looks great.

.github/workflows/ci.yml
markdecrane: Ou that's pretty slick, included https://github.com/hashicorp/go-tfe/pull/1237/commits/a957b805739a48c949a73f8d40fa337b6fd3efb3

barrettclark:
Chefs-kiss

.github/actions/test-go-tfe/action.yml
markdecrane: Mind expanding a bit on what you mean by the ENV var here 😅 

Though reading this gave me another idea that I think might be along the lines of what you're thinking.  [In this commit](https://github.com/hashicorp/go-tfe/pull/1237/commits/cefa6e0f82199ab381fd114dd4600e2f572d1d0b) I added guards to the helper funcs that create runs in order to enforce the naming convention.

helper_test.go
```diff
@@ -1758,6 +1759,7 @@ func fatalDumpRunLog(t *testing.T, client *Client, ctx context.Context, run *Run
 }
 
 func createRun(t *testing.T, client *Client, w *Workspace) (*Run, func()) {
+	runDependentTestNameValidator(t)
```
ctrombley, line 1762: ❤️

PR #1248, opened 2025-11-17
Add project-remote-state attribute to Workspace

hs26gill:
## Description

This PR adds a `project-remote-state` attribute to Workspaces. This attribute allows the workspace to share its remote state with other workspaces within the project. 

## Testing plan

Added unit tests: `TestWorkspacesProjectRemoteState`

- [API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#create-a-workspace)
- [Jira](https://hashicorp.atlassian.net/browse/TF-31708)

## Output from tests

```
$ TFE_ADDRESS="https://example" TFE_TOKEN="example" go test ./... -v -run TestWorkspacesProjectRemoteState
```

Screenshot 2025-11-17 at 7 26 02 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 ## Enhancements
-* Add `ProjectRemoteState` attribute to `Workspace` by @hs26gill [#1248](https://github.com/hashicorp/go-tfe/pull/1248)
+* Adds `ProjectRemoteState` field to `Workspace` to support state sharing at the project level, by @hs26gill [#1248](https://github.com/hashicorp/go-tfe/pull/1248)
 
 # v1.98.0
 
--- a/workspace_integration_test.go
+++ b/workspace_integration_test.go
@@ -3425,18 +3425,11 @@ func TestWorkspacesProjectRemoteState(t *testing.T) {
 
 	// create workspace in first project
 	wTest, wTestCleanup := createWorkspaceWithOptions(t, client, orgTest, WorkspaceCreateOptions{
-		Name:    String(randomString(t)),
-		Project: projTest,
+		Name:              String(randomString(t)),
+		Project:           projTest,
+		GlobalRemoteState: Bool(false),
 	})
 	t.Cleanup(wTestCleanup)
-
-	// Update workspace to allow no remote state sharing
-	options := WorkspaceUpdateOptions{
-		ProjectRemoteState: Bool(false),
-		GlobalRemoteState:  Bool(false),
-	}
-	wTest, err := client.Workspaces.Update(ctx, orgTest.Name, wTest.Name, options)
-	require.NoError(t, err)
 
 	t.Run("successfully returns remote state consumer list", func(t *testing.T) {
 		// create consumer workspace in the test project
@@ -3472,7 +3465,6 @@ func TestWorkspacesProjectRemoteState(t *testing.T) {
 		// Update workspace to allow project remote state sharing
 		options := WorkspaceUpdateOptions{
 			ProjectRemoteState: Bool(true),
-			GlobalRemoteState:  Bool(false),
 		}
 		wTest, err := client.Workspaces.Update(ctx, orgTest.Name, wTest.Name, options)
 		require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Remove org settings validation in RegistryModulesCreateMonorepo tests, by @jillirami ([#1236](https://github.com/hashicorp/go-tfe/pull/1236))
+* Add `ProjectRemoteState` attribute to `Workspace` by @hs26gill [#1248](https://github.com/hashicorp/go-tfe/pull/1248)
```
markdecrane: We'll want to call out that this is in beta here

hs26gill: Updated, thanks!

workspace_integration_test.go
```diff
@@ -3407,3 +3407,90 @@ func TestWorkspaces_effectiveTagBindingsInheritedFrom(t *testing.T) {
...
+			Name:    String(randomString(t)),
+			Project: projTest,
+		})
+		t.Cleanup(wTestCleanupConsumer1)
+
+		// create another project
+		projTest2, projTest2Cleanup := createProject(t, client, orgTest)
+		t.Cleanup(projTest2Cleanup)
+
+		// create consumer workspace in the other project
```
netramali, line 3446: Should we also test what happens if
1. On the project we already had a list of remote state consumers (maybe includes workspaces that are outside the project)
2. User now sets it to project remote state

Expectation: the list should now only have workspaces in the project.

workspace_integration_test.go
```diff
@@ -3407,3 +3407,90 @@ func TestWorkspaces_effectiveTagBindingsInheritedFrom(t *testing.T) {
 		}
 	})
 }
+
+func TestWorkspacesProjectRemoteState(t *testing.T) {
+	skipUnlessBeta(t)
```
jbonhag: Could we please change this, since it's not in beta -- but only TFE-only for now?

```suggestion
	skipUnlessEnterprise(t)
```

jbonhag (changes requested):
Thank you!! The attribute changes look good, but I think we should drop the "beta" labels since this is available in TFE, and we're not calling it beta there.

workspace_integration_test.go
hs26gill: Thanks for the suggestion, added it in.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Add `ProjectRemoteState` attribute to `Workspace` by @hs26gill [#1248](https://github.com/hashicorp/go-tfe/pull/1248)
```
jbonhag: I'm struggling to author a CHANGELOG entry that is both accurate and useful given the timing around merging the ultimate terraform-provider-tfe changes that depend on this PR. What do you think about this?

```suggestion
* The Workspace `ProjectRemoteState` attribute is now available in Terraform Enterprise, and will soon be available in HCP Terraform, by @hs26gill [#1248](https://github.com/hashicorp/go-tfe/pull/1248)
```

jbonhag (changes requested):
I don't know how specific we need to get with the CHANGELOG entry, but it's _accurate_ to say that this is only available on Terraform Enterprise (for now).

workspace_integration_test.go
```diff
@@ -3409,3 +3409,110 @@ func TestWorkspaces_effectiveTagBindingsInheritedFrom(t *testing.T) {
...
+		Name:    String(randomString(t)),
+		Project: projTest,
+	})
+	t.Cleanup(wTestCleanup)
+
+	// Update workspace to allow no remote state sharing
+	options := WorkspaceUpdateOptions{
+		ProjectRemoteState: Bool(false),
+		GlobalRemoteState:  Bool(false),
+	}
```
ctrombley: Isn't the default value already false? It's hard to assert that this worked as expected without any prior value to compare against. Should we instead update to true and assert that the update was successful?

hs26gill: Removed the update call for GlobalRemoteState to false and moved it to [workspace creation](https://github.com/hashicorp/go-tfe/pull/1248/files#diff-6a758ca73359be1bc23f2276ed511980c3c7157055bc9ace202af6a54d0fcbe7R3427-R3430)
The default value seems to be set to true for GlobalRemoteState in the test but we do not need the update call

ctrombley (approved):
Thanks @hs26gill !

PR #1287, opened 2026-03-06
Added Print for Policy Outcome

straubt1:
## Description

The policy-set-outcomes API response includes an `output` array on each outcome object containing `print` entries emitted by Sentinel policies. The `Outcome` struct does not expose this field.

There are additional fields that can be added, but "print" is the output of the the policy logs.

## Changes

- Added `OutcomeOutput` struct with a `Print string` field (`jsonapi:"attr,print"`)
- Added `Output []OutcomeOutput` field to `Outcome` struct (`jsonapi:"attr,output,omitempty"`)
- Added unit tests in `policy_evaluation_test.go` covering field present and field absent cases
- Updated `policy_evaluation_beta_test.go` with an assertion and comment for the `Output` field
- Added CHANGELOG entry under Enhancements

## Testing

- `go build ./...` passes
- `go vet ./...` passes (pre-existing warning in `examples/backing_data/main.go` unrelated to this change)
- Unit tests `TestPolicySetOutcomes_OutcomeOutputDeserialization` and `TestPolicySetOutcomes_OutcomeOutputEmpty` pass


ctrombley:
Superseded by #1330 .

PR #1300, opened 2026-03-27
Token test takes account of default token expiration date

Uk1288:
## Description

Token test now takes account of default token expiration date

## Rollback Plan

- Revert PR

None


ctrombley (approved):
Thanks!

PR #1303, opened 2026-04-01
Added `ProviderType` attribute in saml settings

skj-skj:
## Description

This PR introduces support for configuring the `provider-type` SAML setting. It allows administrators to specify the identity provider type (e.g., `okta`, `entra`, or generic `saml`) directly in their TFE admin SAML configuration.

## Testing plan

1. Added validation for the `ProviderType` field inside `TestAdminSettings_SAML_Read` to confirm the value is successfully read.
2. Added a new sub-test ("with provider type defined") inside `TestAdminSettings_SAML_Update` to verify updates pass for valid provider types (`okta`, `entra`, `saml`, `unknown`) and appropriately error out on invalid ones.

## External links

- [JIRA](https://hashicorp.atlassian.net/browse/TF-35670)
- [RFC](https://hermes-sharepoint.hashicorp.services/document/01XOO7K4IWFH44KZAPDJC3QXNKIRWEZMBA) 

## Output from tests
image

## Rollback Plan

we can revert this pr

NA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Enhancements
+* Adds the `ProviderType` field to `AdminSAMLSetting` and `AdminSAMLSettingsUpdateOptions` to support the `provider-type` SAML setting.
+
 # v1.103.0
 
 ## Enhancements
--- a/admin_setting_saml.go
+++ b/admin_setting_saml.go
@@ -28,30 +28,41 @@ type adminSAMLSettings struct {
 	client *Client
 }
 
+// SAMLProviderType represents the SAML identity provider type.
+type SAMLProviderType string
+
+// SAMLProviderType constants define the supported SAML identity provider types.
+const (
+	SAMLProviderTypeOkta    SAMLProviderType = "okta"
+	SAMLProviderTypeEntra   SAMLProviderType = "entra"
+	SAMLProviderTypeGeneric SAMLProviderType = "saml"
+	SAMLProviderTypeUnknown SAMLProviderType = "unknown"
+)
+
 // AdminSAMLSetting represents the SAML settings in Terraform Enterprise.
 type AdminSAMLSetting struct {
-	ID                        string `jsonapi:"primary,saml-settings"`
-	Enabled                   bool   `jsonapi:"attr,enabled"`
-	Debug                     bool   `jsonapi:"attr,debug"`
-	AuthnRequestsSigned       bool   `jsonapi:"attr,authn-requests-signed"`
-	WantAssertionsSigned      bool   `jsonapi:"attr,want-assertions-signed"`
-	TeamManagementEnabled     bool   `jsonapi:"attr,team-management-enabled"`
-	OldIDPCert                string `jsonapi:"attr,old-idp-cert"`
-	IDPCert                   string `jsonapi:"attr,idp-cert"`
-	SLOEndpointURL            string `jsonapi:"attr,slo-endpoint-url"`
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Enhancements
+* Adds the `ProviderType` field to `AdminSAMLSetting` and `AdminSAMLSettingsUpdateOptions` to support the `provider-type` SAML setting.
```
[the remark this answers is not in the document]
skj-skj: not needed

admin_setting_saml_integration_test.go
```diff
@@ -40,6 +40,7 @@ func TestAdminSettings_SAML_Read(t *testing.T) {
 	assert.NotNil(t, samlSettings.PrivateKey)
 	assert.NotNil(t, samlSettings.SignatureSigningMethod)
 	assert.NotNil(t, samlSettings.SignatureDigestMethod)
+	assert.NotNil(t, samlSettings.ProviderType)
```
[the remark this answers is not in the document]
skj-skj: not applicatble here, using NoNil is existing pattern.

admin_setting_saml.go
```diff
@@ -90,6 +91,7 @@ type AdminSAMLSettingsUpdateOptions struct {
 	WantAssertionsSigned      *bool   `jsonapi:"attr,want-assertions-signed,omitempty"`
 	SignatureSigningMethod    *string `jsonapi:"attr,signature-signing-method,omitempty"`
 	SignatureDigestMethod     *string `jsonapi:"attr,signature-digest-method,omitempty"`
+	ProviderType              *string `jsonapi:"attr,provider-type,omitempty"`
```
Maed223: I think having a type with a set of consts would be a helpful user hint here

Such as:
```go
type ProviderType string

const (
   ProviderTypeOkta    ProviderType = "okta"
   ...
)

File: admin_setting_saml_integration_test.go [hunk 4]
```diff
@@ -130,4 +131,23 @@ func TestAdminSettings_SAML_Update(t *testing.T) {
 		require.NoError(t, err)
 		assert.NotNil(t, samlSettings.IDPCert)
 	})
+
+	t.Run("with provider type defined", func(t *testing.T) {
+		providerTypesForTesting := []string{"okta", "entra", "saml", "unknown", "error"}
```
Maed223: The inclusion of the assertion on "unknown" here feels a bit off, though I'm not very familiar with the backing API. If the intention is to  have a constrained set, then the way the test is currently setup could hide regressions. Open question for me is whether the provider type is intentionally open ended to allow non-standard values.

Nit: this test could be better expressed being table driven. Just so the expected intentions/outcomes are easily parseable.

admin_setting_saml.go
Maed223: Though if the value here isn't being constrained in any way, then ignore this point.

Maed223 (commented):
Smoke tested and works as expected. Just an open question below.

admin_setting_saml.go
skj-skj: thx for the suggestion, I have added the type.

admin_setting_saml_integration_test.go
skj-skj: backend can only have these 4 values: "okta", "entra", "saml" and "unknown".
"unknown" is for backward compatibility, so it will be the default value.

I have update the test to table driven, thx for the suggestion. 
and I have also added validtion as well.

Maed223 (approved):

PR #1314, opened 2026-04-21
Adds `AdminSCIMGroups` to Fetch List of SCIM Groups

skj-skj:
## Description

This PR adds support for fetching SCIM groups provisioned in Terraform Enterprise via an IdP, implemented in `admin_setting_scim_groups.go`. It also includes integration tests, mocks, and a small helper function.

## Testing plan

Adds integration test for `List` for Admin SCIM Groups

## External links
[JIRA](https://hashicorp.atlassian.net/browse/TF-35673)
[RFC](https://hermes-sharepoint.hashicorp.services/document/01XOO7K4IWFH44KZAPDJC3QXNKIRWEZMBA)

## Output from tests
`ENABLE_TFE=1 envchain scim gotest -v -run "TestAdminSCIMGroups_List"`
image

## Rollback Plan
we can revert this pr

NA


helper_test.go
```diff
@@ -3529,6 +3529,26 @@ func createSCIMGroup(ctx context.Context, t *testing.T, client *Client, groupNam
...
+func deleteSCIMGroup(ctx context.Context, t *testing.T, client *Client, groupID, scimToken string) {
+	t.Helper()
+
+	u := client.BaseURL()
+	u.Path = "/scim/v2/Groups/" + groupID
+
+	req, err := retryablehttp.NewRequest("DELETE", u.String(), nil)
+	require.NoError(t, err)
+	req.Header = client.headers.Clone()
+	req.Header.Set("Authorization", "Bearer "+scimToken)
```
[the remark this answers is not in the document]
skj-skj: `DELETE /scim/v2/Groups/:id` api endpoint in TFE returns 204 `No Content` so this is not required.

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+	client := testClient(t)
+	ctx := context.Background()
+
+	scimClient := client.Admin.Settings.SCIM
+
+	enableSCIM(ctx, t, client, true)
+	defer enableSCIM(ctx, t, client, false)
+
+	scimToken, err := scimClient.Tokens.Create(ctx, "integration-test-token")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
skj-skj: not needed, as scim token gets cleanuped up when scim gets disabled via `defer enableSCIM(ctx, t, client, false)`

CHANGELOG.md
```diff
@@ -5,10 +5,11 @@
 
 ## Enhancements
 * Adds the `ProviderType` field to `AdminSAMLSetting` and `AdminSAMLSettingsUpdateOptions` to support the `provider-type` SAML setting.
-* Adds `AdminSCIMSetting` to support managing site-level SCIM settings by @skj-skj [#1307](https://github.com/hashicorp/go-tfe/pull/1307)
+* Adds `AdminSCIMSettings` to support managing site-level SCIM settings by @skj-skj [#1307](https://github.com/hashicorp/go-tfe/pull/1307)
```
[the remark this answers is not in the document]
skj-skj: fixed

admin_setting.go
```diff
@@ -7,6 +7,7 @@ package tfe
 type SCIMResource struct {
 	SCIMSettings
 	Tokens AdminSCIMTokens
+	Groups AdminSCIMGroups
 }
```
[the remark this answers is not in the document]
skj-skj: i have added a note to the `SCIMResource` documentation advising that the struct should only be constructed with keyed fields or obtained via the client to avoid this breakage.

helper_test.go
```diff
@@ -3529,6 +3529,26 @@ func createSCIMGroup(ctx context.Context, t *testing.T, client *Client, groupNam
 	return res.ID
 }
 
+func deleteSCIMGroup(ctx context.Context, t *testing.T, client *Client, groupID, scimToken string) {
+	t.Helper()
+
+	u := client.BaseURL()
+	u.Path = "/scim/v2/Groups/" + groupID
+
+	req, err := retryablehttp.NewRequest("DELETE", u.String(), nil)
```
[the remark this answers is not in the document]
skj-skj: I have updated `deleteSCIMGroup` to use `url.PathEscape(groupID)` to ensure the path is always built safely.

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	t.Run("simple", func(t *testing.T) {
+		t.Run("no groups", func(t *testing.T) {
+			scimGroups, err := scimClient.Groups.List(ctx, nil)
+			require.NoError(t, err)
+			assert.Len(t, scimGroups.Items, 0)
+			assert.Equal(t, 0, scimGroups.TotalCount)
+		})
```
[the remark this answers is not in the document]
skj-skj: Not required, the specific intention of this subtest is to verify that immediately after SCIM is enabled, the true global list of SCIM groups is returned completely empty (0 items). If we filter by a random query prefix, it defeats the purpose of asserting the blank slate state of the endpoint upon first initialization.

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+				id := createSCIMGroup(ctx, t, client, groupName, scimToken.Token)
+				groupIDs = append(groupIDs, id)
+				expectedGroups = append(expectedGroups, AdminSCIMGroup{ID: id, Name: groupName})
+			}
+
+			scimGroups, err := scimClient.Groups.List(ctx, nil)
+			require.NoError(t, err)
+			assert.Len(t, scimGroups.Items, 2)
+			assert.Equal(t, 2, scimGroups.TotalCount)
+
```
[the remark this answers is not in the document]
skj-skj: in our case, it shouldn't be an issue. The TFE test cases run once every day on a nightly instance, meaning no two CI runs will be executing at the exact same time.

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+			groupName := prefix + randomStringWithoutSpecialChar(t)
+			id := createSCIMGroup(ctx, t, client, groupName, scimToken.Token)
+			groupIDs = append(groupIDs, id)
+		}
+
+		// Create 6 matching groups with same suffix "-idp-group"
+		for range 6 {
+			groupName := fmt.Sprintf("%s-idp-group-%s", prefix, randomStringWithoutSpecialChar(t))
+			id := createSCIMGroup(ctx, t, client, groupName, scimToken.Token)
+			groupIDs = append(groupIDs, id)
```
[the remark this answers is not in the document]
skj-skj: I have simplified it by simply removing the trailing `-` from the `prefix` variable definition entirely, which gets rid of the double hyphen formatting down below

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+						for _, exGroup := range excludedGroups.Items {
+							assert.NotEqual(t, g.ID, exGroup.ID)
+						}
+					}
+				}
+			})
+		}
+	})
+
+	t.Run("combine query filtering and pagination", func(t *testing.T) {
```
kkavish, line 215: Is this really required? I am wondering why will pagination fail in query since the URL remains the same only with a filter added.

admin_setting_scim_groups.go
```diff
@@ -0,0 +1,65 @@
...
+	// List all SCIM groups.
+	List(ctx context.Context, options *AdminSCIMGroupListOptions) (*AdminSCIMGroupList, error)
+}
+
+// adminSCIMGroups implements AdminSCIMGroups
+type adminSCIMGroups struct {
+	client *Client
+}
+
+// AdminSCIMGroupList represents a list of SCIM groups
```
kkavish, line 26: Redundant comment, let's remove explanations like these. Does linter require comments for exported data types? If not, we can remove such comments without losing clarity.

kkavish (commented):
@skj-skj A very nicely drafted PR. Thanks! I have a left a couple of comments after which this should be good to go.

admin_setting_scim_groups.go
```diff
@@ -0,0 +1,65 @@
...
+type AdminSCIMGroupListOptions struct {
+	ListOptions
+	Query string `url:"q,omitempty"`
+}
+
+func (o *AdminSCIMGroupListOptions) valid() error {
+	return nil
+}
+
+// List all SCIM groups.
```
nikolasrieble, line 55: **Small bug worth fixing:** `valid()` is a method on `*AdminSCIMGroupListOptions`, and your integration tests call `List(ctx, nil)`. That works *today* because the `valid()` body doesn't touch the receiver — but the moment someone adds a real check like `if o.PageSize < 0 { ... }`, `List(ctx, nil)` will panic with a nil pointer dereference.

Two common ways to fix:

```go
// Option A: nil-safe valid()
func (o *AdminSCIMGroupListOptions) valid() error {
    if o == nil {
        return nil
    }
    return nil
}
```

```go
// Option B: guard at the call site
if options != nil {
        return nil, err
    }
}
```

I'd lean toward **Option A** — it keeps the contract ("this options struct knows how to validate itself, including the nil case") in one place, so future callers don't have to remember the guard.

💡 **Why this matters generally:** in Go, calling a method on a nil pointer is legal as long as the method body doesn't dereference the receiver. That's a sharp edge — code that "works" can silently break when the method body changes. Whenever you add a `valid()`-style helper that accepts an optional pointer, it's worth deciding up-front whether nil is a valid input and encoding that in the method itself.

nikolasrieble (changes requested):
One latent bug worth fixing before merge — details inline. Otherwise nice work on this PR.

admin_setting_scim_groups.go
skj-skj: it got auto generated, when I had ran `RESOURCE=admin_setting_scim_groups make generate`

admin_setting_scim_groups_integration_test.go
skj-skj: yeah its highly unlikely for pagination to fail in query, I have added this to cover all scenarios and ways the api will be used. just a sanity check.

admin_setting_scim_groups.go
skj-skj: locally `make lint` passed but got `exported type AdminSCIMGroupList should have comment or be unexported go-revive` warning in the VScode

admin_setting_scim_groups.go
skj-skj: I have updated `valid()` to be nil-safe so `List(ctx, nil)` remains safe even if we add receiver-based checks later. I have chose the option A to keep nil handling centralized.

admin_setting_scim_groups_integration_test.go
kkavish: Will one of them give us 100% coverage? Like just having the filter and pagination?

admin_setting_scim_groups.go
kkavish: Makes sense, I suspected this to be an exported data type issue. Thanks

admin_setting_scim_groups.go
skj-skj: lint got failed with error `(error) is always nil (unparam)`, so I have added a simple page validation to get it passed.

admin_setting_scim_groups_integration_test.go
skj-skj: I have [removed](https://github.com/hashicorp/go-tfe/pull/1314/changes/3caf57cdcd76d4886484dbb6dbcee5a3bde53ca5) the custom size and out of bound test cases which are been redundant

kkavish (dismissed):

admin_setting.go
```diff
@@ -1,12 +1,15 @@
-// Copyright IBM Corp. 2018, 2025
```
shwetamurali, line 1: nit: not sure why this was changed in a few files but it is not needed/consistent with the repo

CHANGELOG.md
```diff
@@ -5,11 +5,12 @@
 
 ## Enhancements
 * Adds the `ProviderType` field to `AdminSAMLSetting` and `AdminSAMLSettingsUpdateOptions` to support the `provider-type` SAML setting.
-* Adds `AdminSCIMSetting` to support managing site-level SCIM settings by @skj-skj [#1307](https://github.com/hashicorp/go-tfe/pull/1307)
+* Adds `SCIMSettings` / `AdminSCIMSetting` to support managing site-level SCIM settings by @skj-skj [#1307](https://github.com/hashicorp/go-tfe/pull/1307)
```
shwetamurali, line 8: it's typically not best to rewrite old entries, it would be better to address the renaming in the new entry. but since this is unreleased it is not blocking

admin_setting_scim_groups_integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	t.Run("paginate through groups", func(t *testing.T) {
+		var groupIDs []string
+		t.Cleanup(func() {
+			for _, id := range groupIDs {
+				deleteSCIMGroup(ctx, t, client, id, scimToken.Token)
+			}
+		})
+
+		prefix := randomStringWithoutSpecialChar(t) + "-"
+		// Create 30 groups to test default page size of 20
```
shwetamurali: If the default page size is ever changed, this breaks. I recommend explicitly setting `PageSize` to 20 in the options to make this test self contained

shwetamurali (dismissed):

admin_setting.go
skj-skj: since the file was being touched in 2026, I figured it made sense to bump the copyright end year to match. happy to revert across the affected files if the repo convention is to leave the header alone.

CHANGELOG.md
skj-skj: fair point on not rewriting old entries in general. since these are unreleased though I wanted to keep the changelog accurate, the original entries only mentioned the struct names (`AdminSCIMSetting`, `AdminSCIMToken`) and missed the interfaces (`SCIMSettings`, `AdminSCIMTokens`). since consumers would interact with the interfaces, I felt it was worth correcting in-place.

admin_setting_scim_groups_integration_test.go
skj-skj: Good point, but I explicitly want to test the API default pagination behaviour here, hardcoding `PageSize` in the options would defeat that purpose.

to make it less brittle I pulled the expected size into a `defaultPageSize` constant and added a clear error message. This way, if the API default ever changes, the test will fail cleanly and only require a one-line update.

shwetamurali (approved):

nikolasrieble (approved):
🎉

PR #1316, opened 2026-04-27
setup + first skips

aaabdelgany:
## Description

This PR removes the beta skip flags in all of the stacks integration tests to ensure that they start running in CI. It also removes any references to stack_diagnostics#acknowledge which is something that was removed as part of the Stacks GA release and we just happened to miss it in go-tfe.

## Output from tests

```

...
```


brandonc:
We're transitioning away from the traditional integration tests in this package. Sorry to have to close this unmerged!

PR #1322, opened 2026-04-30
feat: add TerraformVersion to RegistryNoCodeModuleCreateWorkspaceOptions

chrisarcand:
## Description

The no-code module workspace creation API accepts a terraform-version attribute, but the Go client struct was missing the field, making it impossible to specify a Terraform version at creation time.

Add the field with an accurate doc comment (the fallback is the platform's latest available Terraform version, not an org-level default) and a corresponding integration test subcase that verifies the value roundtrips correctly through the API.

### DIFF
--- a/registry_no_code_module_integration_test.go
+++ b/registry_no_code_module_integration_test.go
@@ -417,6 +417,23 @@ func TestRegistryNoCodeModulesCreateWorkspace(t *testing.T) {
 		)
 		r.Error(err)
 	})
+
+	t.Run("test creating a workspace with a specified terraform version", func(t *testing.T) {
+		wn := fmt.Sprintf("foo-%s", randomString(t))
+		tfVersion := "1.3.0"
+		w, err := client.RegistryNoCodeModules.CreateWorkspace(
+			ctx,
+			ncm.ID,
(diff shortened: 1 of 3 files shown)


chrisarcand:
Replaces #1320

bbasata (approved):
As an exceptional case: I approve the use of this change via psuedoversion (i.e. Git ref).

```
$ go get -u github.com/hashicorp/go-tfe@7f37f4e330185da50318cbbb5e7fb46f71e338ba
```

I have confirmed that any failing tests are known flakes.

registry_no_code_module_integration_test.go
```diff
@@ -417,6 +417,23 @@ func TestRegistryNoCodeModulesCreateWorkspace(t *testing.T) {
 		)
 		r.Error(err)
 	})
+
+	t.Run("test creating a workspace with a specified terraform version", func(t *testing.T) {
```
bbasata, line 421: For thoroughness: I noticed that this test was skipped in the checks that have reported to this PR so far.

I think it's likely that integration tests such as this one are exercised only in the nightly build.

That said, I believe we can move forward & that we are exercising the right balance of speed and safety.

Uk1288 (approved):

PR #1323, opened 2026-04-30
Adding Post Apply Status Constants in Run

jose-kunnel:
## Description
This PR introduces a new constant for run status, which needs to be included in the terraform-provider-tfe repository. Currently, the RUN status does not include RunPostApplyCompleted and RunPostApplyRunning, and these statuses are also missing from the applyPendingStatuses list in resource_tfe_workspace_run.go.

As a result, when creating a run with a post-apply run task configured, the status post_apply_running is not recognized and leads to an error due to it being treated as an invalid status.

To resolve this issue, these new run status constants must be added to terraform-provider-tfe.

## Testing plan
Create a Terraform run with either a global run task configured for post-apply or a run task configured at the workspace level.

Before the fix:
This setup results in an error, as the post_apply_running status is not recognized and is treated as invalid.

Screenshot 2026-04-30 at 11 02 40 AM

After the Fix
The run is created successfully, and no error occurs, as the post_apply_running status is now properly recognized.

## External links

[JIRA] https://hashicorp.atlassian.net/browse/IPL-9456

## Rollback Plan

Revert the PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Bug Fixes
+* Adds `RunPostApplyRunning`,`RunPostApplyCompleted` run status by @jose-kunnel [#1323](https://github.com/hashicorp/go-tfe/pull/1323)
+
 ## Enhancements
 * Adds `CanReadStateVersions` and `CanReadVariable` fields to `WorkspacePermissions` by @jondavidjohn [#1325](https://github.com/hashicorp/go-tfe/pull/1325)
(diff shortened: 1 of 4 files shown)


christian-doucette (commented):
could you add the new states to the test here? 
https://github.com/hashicorp/go-tfe/blob/4adc65619021ddbfe94f7ae65c854275302ece6f/admin_run_test.go#L16

jose-kunnel:
> could you add the new states to the test here?
> 
> https://github.com/hashicorp/go-tfe/blob/4adc65619021ddbfe94f7ae65c854275302ece6f/admin_run_test.go#L16

Added changes  for the new Run Status for validating in admin_run.go and admin_run_test.go

christian-doucette (dismissed):

christian-doucette (approved):

PR #1327, opened 2026-05-04
TF-36033: Add GenerateConfigOut to QueryRun struct

mjyocca:
## Description

Adds `GenerateConfigOut` support to the `QueryRun` struct and `QueryRunCreateOptions`.

`GenerateConfigOut` controls whether the Terraform query CLI execution passes the `-generate-config-out=` flag. When `true`, Terraform generates resource configuration output as a side effect of the query run. The field defaults to `true` when omitted from the create request.

Changes:
- Added `GenerateConfigOut bool` to `QueryRun` (response struct) to reflect the value returned by the API
- Added `GenerateConfigOut *bool` to `QueryRunCreateOptions` with `omitempty`, so callers can explicitly set it to `false` or omit it entirely to accept the API default of `true`

> **Note:** This field is in BETA and subject to change.

## Testing plan

Integration tests are gated behind `skipUnlessBeta(t)` and require `ENABLE_BETA=1` to run locally. See `TESTS.md` for setup instructions.

Two subtests cover the expected behavior in `TestQueryRunsCreate_RunDependent`:

1. **Default (`true`) when omitted** — verifies that omitting `GenerateConfigOut` from `QueryRunCreateOptions` results in the API returning `true`:
```
ENABLE_BETA=1 TFE_ADDRESS="..." TFE_TOKEN="..." go test ./... -v -run TestQueryRunsCreate_RunDependent/without_a_configuration_version
```

2. **Explicit `false`** — verifies that passing `GenerateConfigOut: Bool(false)` results in the API returning `false`:
```
ENABLE_BETA=1 TFE_ADDRESS="..." TFE_TOKEN="..." go test ./... -v -run TestQueryRunsCreate_RunDependent/with_generate-config-out_set_to_false
```

## External links

- [Queries API reference](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/queries)

## Output from tests

> Tests for this change require a beta-enabled TFE environment. Output to be added once verified against a compatible instance.

## Rollback Plan

If this change needs to be reverted, a follow-up PR removing the `GenerateConfigOut` fields from `QueryRun` and `QueryRunCreateOptions` can be opened and merged within 7 days.

None. This change only adds a new optional attribute to an existing API request/response struct.


lilincmu (dismissed):
Looks good to me!

query_runs_integration_test.go
```diff
@@ -116,6 +116,8 @@ func TestQueryRunsCreate_RunDependent(t *testing.T) {
 		assert.NotNil(t, qr.CreatedAt)
 		assert.NotNil(t, qr.Source)
 		require.NotNil(t, qr.StatusTimestamps)
+		// GenerateConfigOut defaults to true when omitted from the create options.
+		assert.True(t, qr.GenerateConfigOut)
```
ctrombley, line 120: The API always returns `true` in this case?

ctrombley (approved):

query_runs_integration_test.go
mjyocca: Yes! To keep backwards compatibility with versions of TF

PR #1344, opened 2026-05-19
Add `FollowAPIRedirect` method for streaming redirect endpoint responses

shwetamurali:
## Description
Several API endpoints return a 302 redirect to Archivist. The generated client can't handle these because it uses `SendNoContent` (the OpenAPI spec doesn't define a body on 302 responses). This method gives SDK callers a supported way to follow those redirects and stream the content.

- Adds `Client.FollowAPIRedirect` which follows the redirect chain from TFE API endpoints that return 302 → Archivist (307) → S3 (200), returning the final body as an `io.ReadCloser` without buffering into memory
- Handles the case where Archivist proxies content directly (redirects disabled) as well as multi-hop presigned URL redirects
- Detects redirect loops and returns `tfe.ErrRedirectLoop` and returns an error on non-200 final status

## Testing plan
Tests use `httptest.NewServer` to simulate:
- Single redirect (TFE → Archivist proxied response)
- Real redirect chain (TFE 302 → Archivist 307 → S3 200)
- Redirect loop detection
- Non-200 final response error
- Streaming body (only reads 10 bytes of 1MB to prove no buffering)
- No Authorization header leaked to redirect targets

## Output from tests

```
=== RUN   TestFollowAPIRedirect_SingleRedirect
--- PASS: TestFollowAPIRedirect_SingleRedirect (0.00s)
=== RUN   TestFollowAPIRedirect_TFEToArchivistToS3
--- PASS: TestFollowAPIRedirect_TFEToArchivistToS3 (0.00s)
=== RUN   TestFollowAPIRedirect_ArchivistNoRedirect
--- PASS: TestFollowAPIRedirect_ArchivistNoRedirect (0.00s)
=== RUN   TestFollowAPIRedirect_RedirectLoop
--- PASS: TestFollowAPIRedirect_RedirectLoop (0.00s)
=== RUN   TestFollowAPIRedirect_Non200FinalResponse
--- PASS: TestFollowAPIRedirect_Non200FinalResponse (0.00s)
=== RUN   TestFollowAPIRedirect_StreamingBody
--- PASS: TestFollowAPIRedirect_StreamingBody (0.00s)
=== RUN   TestFollowAPIRedirect_NoAuthHeaderSent
--- PASS: TestFollowAPIRedirect_NoAuthHeaderSent (0.00s)
PASS
ok  	github.com/hashicorp/go-tfe	0.465s
```

### DIFF
--- a/follow_redirect.go
+++ b/follow_redirect.go
@@ -13,11 +13,33 @@ import (
 
 const maxRedirectHops = 10
 
+// ErrRedirectLoop is returned when a redirect cycle is detected or the
+// maximum number of redirect hops is exceeded.
 var ErrRedirectLoop = errors.New("redirect loop detected")
 
+// FollowAPIRedirect follows the redirect chain from an API response until it
+// reaches the final response body, returning it as an io.ReadCloser. The caller
+// is responsible for closing the returned body and for limiting the amount of
+// data read (the response may be arbitrarily large).
+//
+// This is intended for API endpoints that return a 302 redirect to an Archivist
+// presigned URL, which may in turn issue a 307 redirect to a storage backend
+// (e.g., S3). The method intentionally does NOT reuse the client's own HTTP
+// transport or send the Authorization header on follow-up requests, because the
+// redirect targets are presigned URLs that are self-authenticating and must not
+// receive the bearer token.
+//
+// Returns ErrRedirectLoop if a cycle is detected or more than 10 hops occur.
+// Returns an error if the final response status is not 200.
 func (c *Client) FollowAPIRedirect(ctx context.Context, resp *http.Response) (io.ReadCloser, error) {
 	if resp == nil {
 		return nil, errors.New("response must not be nil")
+	}
+
+	httpClient := &http.Client{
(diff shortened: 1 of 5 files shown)


follow_redirect.go
```diff
@@ -0,0 +1,90 @@
...
+			return currentResp.Body, nil
+		}
+
+		if currentResp.StatusCode != http.StatusFound &&
+			currentResp.StatusCode != http.StatusMovedPermanently &&
+			currentResp.StatusCode != http.StatusTemporaryRedirect &&
+			currentResp.StatusCode != http.StatusPermanentRedirect &&
+			currentResp.StatusCode != http.StatusSeeOther {
+			currentResp.Body.Close() //nolint:errcheck
+			return nil, fmt.Errorf("unexpected response status: %d", currentResp.StatusCode)
```
brandonc, line 59: Even though we're returning an error, let's also return the currentResp here so the caller can inspect

brandonc (commented):
Can you add a little example code to the README? This will be a very common pattern

brandonc:
@shwetamurali went this direction because of some kiota limitations  f5586dc67608915908468d63a94f5a7802154499

PR #1353, opened 2026-06-02
Update error.go

shwetamurali:
## Description

Fixes a bug where `ErrorMiddleware` prevented `RetryHandler` from retrying 5xx responses. 
`NewErrorMiddleware` was appended to the end of the middleware pipeline, converting 4xx/5xx responses into Go errors before `RetryHandler` could inspect the status code. Since Kiota checked the error before evaluating retry logic, retries never fired.

Fix: Remove `NewErrorMiddleware` from the middleware pipeline entirely. Instead, apply error conversion in `GetStream`. If the error code >= 400, call `APIErrorFactory(resp, nil)` and return the error. This ensures that the handler sees the raw response and can retry retryable status codes, error conversion only happens after retries are exhausted, and callers still receive the same API Error for 4xx/5xx responses

## Testing plan

1. Verify that 5xx responses are retried up to MaxRetries before returning an error
2. Verify that 4xx responses (non-retryable) still return *APIError immediately
3. Verify existing tests pass with no regressions
4. Run `go build ./...`

## Output from tests

```
=== RUN   TestGetStream
=== RUN   TestGetStream/With_base_URL
=== RUN   TestGetStream/Without_base_URL
=== RUN   TestGetStream/With_query_parameters
=== RUN   TestGetStream/With_host_and_scheme
=== RUN   TestGetStream/No_leading_slash,_no_base_path
=== RUN   TestGetStream/With_headers
--- PASS: TestGetStream (0.00s)
    --- PASS: TestGetStream/With_base_URL (0.00s)
    --- PASS: TestGetStream/Without_base_URL (0.00s)
    --- PASS: TestGetStream/With_query_parameters (0.00s)
    --- PASS: TestGetStream/With_host_and_scheme (0.00s)
    --- PASS: TestGetStream/No_leading_slash,_no_base_path (0.00s)
    --- PASS: TestGetStream/With_headers (0.00s)
PASS
ok  	github.com/hashicorp/go-tfe	0.459s

...
```


brandonc:
It looks like you just moved error middleware to the very top, so 4xx and 5xx are still always converted to errors. (Different from PR description)

brandonc (approved):

PR #1356, opened 2026-06-06
feat(explorer): add query and CSV export support

### LINKED ISSUE #909: Add Support for Explorer API
#### Use-cases
the go-tfe library should support the new explorer API https://developer.hashicorp.com/terraform/cloud-docs/api-docs/explorer so users can run the same queries utilizing the go-tfe library

#### Attempted Solutions
I can write my own raw http calls to the endpoints, but this should be supported by go-tfe as its an official Terraform Cloud API.

#### Proposal
Add support for the endpoint, its response and query params/filters

a9logic:
## Description

Adds an `Explorer` client for read access to the HCP Terraform / Terraform Enterprise Explorer API. Resolves #909.

It exposes the two generally available read endpoints:
- **`Query`** — query a view (`workspaces`, `tf_versions`, `providers`, `modules`) with filters, sort, field selection, and pagination.
- **`ExportCSV`** — export the same data as CSV.

The design is intentionally generic: view types, fields, and operators are passed as strings and validated by the backend rather than hardcoded in the client, so there's no baked-in field/type schema to maintain or unwind later. The exported `ExplorerViewType` / `ExplorerOperator` constants are conveniences only; the option fields accept any string, so values the backend adds later work without a client release.

### Prior art

This builds on the groundwork in #1018 by @sebasslash, which first mapped the Explorer endpoints, established the `filter[i][field][operator][j]` query-parameter encoding, and introduced the `EXPLORER_TEST_ORGANIZATION` integration-test gating that this PR reuses. The main difference here is the approach to typing: rather than hardcoding per-view structs and methods, this client keeps view types, fields, and operators as backend-validated strings — which is what addresses the discovery/lock-in concern that paused #1018.

## Testing plan

- **Unit tests** cover filter query-parameter encoding (`filter[i][field][operator][j]`) and option validation.
- **Mock-server (`httptest`) tests** exercise the full request path — URL and query-parameter encoding — and response decoding (the polymorphic JSON:API records and `meta.pagination`) for both `Query` and `ExportCSV`.
- **Integration tests** `TestExplorerQuery` and `TestExplorerExportCSV` exercise the methods end-to-end. Because the Explorer aggregates organization-wide data with no API to seed it synchronously, they're gated on the `EXPLORER_TEST_ORGANIZATION` environment variable (an org that already contains workspaces) and skip otherwise.

I was unable to run the integration tests against a live Explorer-enabled backend, so they're included but verified only through local build/vet/lint and the unit + mock-server tests, which pass.

## External links

- [Explorer API documentation](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/explorer)
- Resolves #909

## Output from tests

```
$ go build ./...
OK

$ gofmt -l explorer.go explorer_test.go explorer_integration_test.go examples/explorer/main.go
(no output — clean)

$ go vet ./...        # clean for changed files
$ golangci-lint run   # 0 issues for changed files

$ go test -run TestExplorer -v .
--- SKIP: TestExplorerQuery (0.00s)               # needs EXPLORER_TEST_ORGANIZATION
--- SKIP: TestExplorerExportCSV (0.00s)           # needs EXPLORER_TEST_ORGANIZATION
--- PASS: TestExplorerQueryOptions_filterParams (0.00s)
--- PASS: TestExplorerQueryOptions_valid (0.00s)
--- PASS: TestExplorerExportCSV_returnsRawBytes (0.00s)
--- PASS: TestExplorerQuery_requestEncodingAndDecoding (0.00s)
PASS
ok  	github.com/hashicorp/go-tfe	0.084s
```

## Rollback Plan

Revert the commit. The change is purely additive — a new `explorer.go` plus its registration on the client — and there are no callers of the new methods inside the SDK.

None.

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -4716,6 +4716,12 @@ var (
 	ErrInvalidSAMLProviderType = errors.New("invalid SAML provider type")
 
 	ErrInvalidTFPolicyEvaluationID = errors.New("invalid value for tfpolicy evaluation ID")
+
+	ErrInvalidExplorerViewType = errors.New("explorer query type is required")
+
+	ErrInvalidExplorerFilterField = errors.New("explorer filter field is required")
+
+	ErrInvalidExplorerFilterOperator = errors.New("explorer filter operator is required")
 )
 
 var (
@@ -21119,6 +21125,7 @@ type Client struct {
 	Comments                        Comments
 	ConfigurationVersions           ConfigurationVersions
 	CostEstimates                   CostEstimates
+	Explorer                        Explorer
 	GHAInstallations                GHAInstallations
 	GPGKeys                         GPGKeys
 	NotificationConfigurations      NotificationConfigurations
@@ -21467,6 +21474,7 @@ func NewClient(cfg *Config) (*Client, error) {
 	client.Comments = &comments{client: client}
 	client.ConfigurationVersions = &configurationVersions{client: client}
 	client.CostEstimates = &costEstimates{client: client}
+	client.Explorer = &explorer{client: client}
 	client.GHAInstallations = &gHAInstallations{client: client}
 	client.GPGKeys = &gpgKeys{client: client}
 	client.RegistryNoCodeModules = &registryNoCodeModules{client: client}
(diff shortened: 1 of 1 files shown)


brandonc:
I'm sorry to report that your contribution to go-tfe has come at a strange time. We're transitioning from a human-coded API interface with integration tests to a code-generated API interface with unit tests.

You have two choices from here:

1. You can rebase your new explorer interface on main, moving your code to v1.go and deleting the test code.
2. You can close this PR and define the new operations as OpenAPI specification in the platform code and it will be code-generated automatically and released shortly after in the `v2` package

a9logic:
Went with option 1 - rebased on main (it was already current), dropped the unit/mock-server/integration tests and the generated mocks. The implementation stays in `explorer.go` to match the per-feature convention on the v1 line. Happy for you to merge as-is if it's useful; I'm not planning further investment on the v2/OpenAPI path since that needs the explorer upstream spec.

brandonc:
I'm not sure your fork is up to date with this repo's main branch. I consolidated all v1 package files into a single file v1.go to reduce the surface area of the legacy package. But I haven't cut a release yet and you're welcome to add your changes (and your changes to tfe.go and errors.go) to that file. You can drop the CHANGELOG changes altogether. Thanks and sorry for the unfortunate timing.

a9logic:
Updated - rebased onto the latest main with the `v1.go` consolidation, and folded everything into `v1.go`: the Explorer impl, the Client field + NewClient registration (was `tfe.go`), and the error vars (was `errors.go`). Dropped the CHANGELOG changes as you suggested; tests/mocks are gone too. Thanks for the heads-up on the consolidation.

brandonc (approved):

brandonc:
@a9logic thanks for the contribution! It'll go out as 2.0.0 and will be the last feature added to the v1 package 🏆

PR #1357, opened 2026-06-08
Don't block s390x tool versions in go-tfe

JenGoldstrich:
## Description

go-tfe currently blocks s390x tool versions, which prevents the usage of s390x agents.

## Testing plan

Try and upload a linux s390x Terraform version.

## Output from tests

N/A

## Rollback Plan

N/A

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Enhancements
 * Adds support for `NotificationConfigurationSubscribableChoice` type `Project` to support Project Level Notifications by @jillirami [#1350](https://github.com/hashicorp/go-tfe/pull/1350)
 
+## Bug Fixes
+* Don't block s390x tool versions
+
 # v1.108.0
 
 ## Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Adds support for `NotificationConfigurationSubscribableChoice` type `Project` to support Project Level Notifications by @jillirami [#1350](https://github.com/hashicorp/go-tfe/pull/1350)
 
 ## Bug Fixes
-* Don't block s390x tool versions
+* Adds support for s390x tool versions, which were previously blocked by @JenGoldstrich [#1357](https://github.com/hashicorp/go-tfe/pull/1357)
 
 # v1.108.0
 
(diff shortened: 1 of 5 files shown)


admin_terraform_version.go
```diff
@@ -237,5 +238,5 @@ func (o AdminTerraformVersionCreateOptions) validArchs() bool {
 }
 
 func validArch(a *ToolVersionArchitecture) bool {
-	return a.URL != "" && a.Sha != "" && a.OS == linux && (a.Arch == amd64 || a.Arch == arm64)
+	return a.URL != "" && a.Sha != "" && a.OS == linux && (a.Arch == amd64 || a.Arch == arm64 || a.Arch == s390x)
```
shwetamurali, line 241: This function is also called by `admin_sentinel_version.go` and `admin_opa_version.go` . Should we add s390x entries to those integration tests too for coverage?

shwetamurali:
The current `validArch` function uses a strict allowlist approach, so any new architecture requires a code change. This is fine for a controlled set, but if more architectures are expected in the future, it might be worth considering whether the validation should be loosened (e.g., just check `a.URL != "" && a.Sha != "" && a.OS != "`") or made configurable. That said, this is a broader architectural question and not something this PR needs to solve.

admin_terraform_version.go
JenGoldstrich: I added those tests

JenGoldstrich:
@shwetamurali yeah I get where you're coming from, I could see the value of just removing this client side check for the architecture, definitely would appreciate it being made in a follow up PR though

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Enhancements
 * Adds support for `NotificationConfigurationSubscribableChoice` type `Project` to support Project Level Notifications by @jillirami [#1350](https://github.com/hashicorp/go-tfe/pull/1350)
 
+## Bug Fixes
+* Don't block s390x tool versions
```
shwetamurali: Please follow the same format as the other entries! Thank you

JenGoldstrich: Whoops, good call, done!

shwetamurali (approved):

PR #1364, opened 2026-06-15
Prep v2 release

brandonc:
## Description

- Integrated open PRs to v1 when possible
- `make api`
- Nightly automation to `make api` and merge the results
- Fix README configuration reference
- Remove User-Agent configuration in favor of simple header configuration
- New PR Template
- Security fixes in v2/examples


.github/workflows/merge-nightly-api-update.yml
```diff
@@ -0,0 +1,104 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  merge-latest:
+    name: Merge Latest Nightly API PR
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' && startsWith(github.event.workflow_run.head_branch, 'automation/nightly-api-update/')) }}
```
ctrombley, line 18: Consider gating on service user as well in case a user manually creates a matching branch by accident?

Something like:
`pr.user.login == "github-actions[bot]"`

v1.go
```diff
@@ -11913,9 +11913,6 @@ type QueryRuns interface {
 
 	// Cancel a query run by its ID.
 	Cancel(ctx context.Context, runID string) error
-
-	// Force-cancel a query run by its ID.
-	ForceCancel(ctx context.Context, runID string) error
```
ctrombley, line 11918: Was this removed intentionally?

v1.go
```diff
@@ -12165,23 +12162,6 @@ func (r *queryRuns) Cancel(ctx context.Context, queryRunID string) error {
...
-	}
-
-	u := fmt.Sprintf("queries/%s/actions/force-cancel", url.PathEscape(queryRunID))
-	req, err := r.client.NewRequest("POST", u, nil)
-	if err != nil {
-		return err
-	}
-
-	return req.Do(ctx, nil)
-}
```
ctrombley, line 12180: Was this removed intentionally?

v2/tfe.go
```diff
@@ -51,10 +49,6 @@ type Config struct {
 
 	// RetryMaxRetries sets the maximum number of retries for a request before giving up.
 	RetryMaxRetries int
-
-	// UserAgent is the User-Agent header value sent with each request.
-	// If not set, a default value ("go-tfe") will be used.
-	UserAgent string
```
ctrombley, line 57: Intentional breaking change?

v1.go
brandonc: Yes, it was part of an open PR that was not merged. I'll double check with Sebastian.

v2/tfe.go
brandonc: Not breaking because there's no release yet. I think I prefer the unambiguity of setting a user-agent default header, rather than an explicit one

v1.go
brandonc: Confirmed we want to remove this.

v1.go
sebasslash: Yes this is desired, `ForceCancel` doesn't do anything.

Maed223 (approved):
Just to double check on [Chris' point here](https://github.com/hashicorp/go-tfe/pull/1364#discussion_r3417310427), a malicious actor trying to merge code would be stopped since they don't have CI privileges correct?

brandonc:
@Maed223 I think I should double check on this myself. I need to get the workflow merged to start testing it and will reconsider the security implications. We may have to merge these manually.

PR #1367, opened 2026-06-17
Fix access token allowed host configuration, prep v2.0.0 release

brandonc:
## Description

Finalizing v2.0.0 release.

- Modifies the token provider to only send the token to the configured hostname. This is important because you can use the client on arbitrary URLs, some of which could be adversarial.

- Adds `make api_internal` which uses a local copy of the internal-beta spec to generate SDK, but also adds a git pre-commit hook to prevent you from accidentally committing these to this repo.

- Removes nightly merge workflow, but keep the "close older PRs" step

- Docs revisions

## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.


ctrombley (changes requested):
Looks really good @brandonc, I found some issues using valid/self-hosted URLs for the hostname configuration.

- IPv4 + port (`127.0.0.1:9000`)
- IPv6 (`[::1]:8080`)
- `localhost:8080`

in each case I'm getting failures. Here are some test cases you can drop into `TestNormalizeHostname`:

```go
{
        Name:     "ipv4 hostname with port",
        Input:    "127.0.0.1:9000",
        Expected: "127.0.0.1:9000",
},
{
        Name:     "localhost with port",
        Input:    "localhost:8080",
        Expected: "localhost:8080",
},
{
        Name:     "ipv6 hostname with port",
        Input:    "[::1]:8080",
        Expected: "[::1]:8080",
},
```

docs/UPGRADING.md
```diff
@@ -1,23 +1,18 @@
...
+- The v2 module no longer makes an initial request in order to decode and store platform configuration.
 
-The v2 package does not include any up-to-date mocks. Old mocks may be found in the v1 `mocks` package.
+- The v2 module does not include any up-to-date mocks. Old mocks may be found in the v1 `mocks` module.
 
-To upgrade, get the module
+To upgrade, get the module and import as `github.com/hashicorp/go-tfe/v2`
 
 ```bash
 $ go get github.com/hashicorp/go-tfe/v2@v2.0.0-beta1
```
ctrombley: Consider making this a little less specific for anyone who needs the extra hint. We're already moving to v2.0.1 which makes this example outdated.

```suggestion
$ go get github.com/hashicorp/go-tfe/v2@<version>
```

.github/workflows/ci.yml
```diff
@@ -9,33 +9,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
-    name: Lint
+  lint-test:
+    name: Lint/Test
     runs-on: ubuntu-latest
     timeout-minutes: 5
```
ctrombley, line 15: Wonder if this needs a bump since we are consolidating? Linting is usually pretty quick. Wdyt?

ctrombley (dismissed):
Looks good! no major issues found. Approving unless you want to make any more changes based on my minor comments.

.github/workflows/ci.yml
brandonc: It should be very quick. I renamed back to just "Lint" because that is a required check and I can't seem to change the required checks anymore!

brandonc: I was able to change the branch protections and further consolidate the CI workflow

ctrombley (approved):
Thanks!

PR #1376, opened 2026-06-23
[TF-37921] Add VCS connection fields to registry module update options

hashimoon:
## Description

Adds `Identifier`, `OAuthTokenID`, and `GHAInstallationID` to `RegistryModuleVCSRepoUpdateOptions` so callers can update a registry module's VCS connection — its repository identifier and the OAuth token or GitHub App installation — via `RegistryModules.Update`, not just the publishing mechanism (branch/tags). Setting both `OAuthTokenID` and `GHAInstallationID` in the same call returns the new `ErrMutuallyExclusiveOAuthTokenAndGHAInstallation`.

This matches the corresponding HCP Terraform / Atlas registry-module API change, which is now generally available.

## Testing plan

1. The mutual-exclusion validation runs before any network call, so it is covered by a self-contained unit test (an `httptest` server stands in for the API):

```
go test -run TestRegistryModules_Update_VCSRepoMutualExclusionValidation .
```

It asserts both IDs set → `ErrMutuallyExclusiveOAuthTokenAndGHAInstallation`, and that only-one / neither set does not return that validation error.

## External links

- [Related Atlas PR](https://github.com/hashicorp/atlas/pull/29638)

## Output from tests

```
$ go test -run TestRegistryModules_Update_VCSRepoMutualExclusionValidation .
ok  	github.com/hashicorp/go-tfe	0.605s
```

## Rollback Plan

If this change needs to be reverted, revert the pull request. The change is additive (new optional option fields plus one validation error) with no data migration or stateful component.

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Enhancements
+* Adds `Identifier`, `OAuthTokenID`, and `GHAInstallationID` fields to `RegistryModuleVCSRepoUpdateOptions` so callers can update a registry module's VCS repository identifier and connection (OAuth token or GitHub App installation) via the `Update` method; setting both `OAuthTokenID` and `GHAInstallationID` in the same call returns `ErrMutuallyExclusiveOAuthTokenAndGHAInstallation` [#1376](https://github.com/hashicorp/go-tfe/pull/1376)
+
 # v2.0.0
 
 The go-tfe v2 package has been added to this repository and contains substantial breaking changes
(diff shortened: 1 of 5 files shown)


hashimoon:
Closing in favor of the OpenAPI→v2 path. go-tfe v1 is deprecated, and the registry-module VCS-connection update is being added to the HCP Terraform OpenAPI spec in hashicorp/atlas#29638 — the go-tfe v2 client is generated from that spec (nightly `make api`), so no hand-written go-tfe change is needed.

PR #1416, opened 2026-07-30
fix: concurrency-related panic in v2 pipeline

ctrombley:
When using Terraform built against go-tfe/v2 to create multiple resources in parallel, the provider binary crashes with:

```
fatal error: concurrent map read and map write
github.com/microsoft/kiota-abstractions-go.(*header).Add 
github.com/microsoft/kiota-http-go.HeadersInspectionHandler.Intercept
```

To reproduce: Using a go-tfe/v2-enabled Teraform build, run `terraform apply` with any config containing ≥2 tfe_team resources.

Root cause: `GetForKiota` creates one `khttp.HeadersInspectionHandler` per `NewClient` call. The upstream `Intercept` method falls back to `&middleware.options` (which is shared handler state) when options are not already set per request. Multiple goroutines all write into the same `ResponseHeaders` map, which causes a fatal concurrent map write.

Fix: Implement a `safeHeadersInspectionHandler` in the middleware package that creates a fresh `HeadersInspectionOptions` per request when options are not already set per request, eliminating any shared mutable state.

### DIFF
--- a/v2/middleware/middleware.go
+++ b/v2/middleware/middleware.go
@@ -101,12 +101,6 @@ func GetForKiota(tfeSDKVersion string, options ...MiddlewareOption) ([]khttp.Mid
 			ProductName:    "go-tfe",
 			ProductVersion: tfeSDKVersion,
 		}),
-		// safeHeadersInspectionHandler replaces khttp.HeadersInspectionHandler
-		// to fix a data race: the upstream handler shares a single
-		// HeadersInspectionOptions (and its ResponseHeaders map) across all
-		// concurrent requests that don't carry a per-request option, causing
-		// fatal concurrent map writes. Our implementation allocates fresh
-		// options per-request when none is provided by the caller.
-		newSafeHeadersInspectionHandler(false, true),
+		NewHeadersInspectionHandler(false, true),
 	}, nil
 }
--- a/v2/middleware/middleware.go
+++ b/v2/middleware/middleware.go
@@ -101,11 +101,12 @@ func GetForKiota(tfeSDKVersion string, options ...MiddlewareOption) ([]khttp.Mid
 			ProductName:    "go-tfe",
 			ProductVersion: tfeSDKVersion,
 		}),
-		khttp.NewHeadersInspectionHandlerWithOptions(func() khttp.HeadersInspectionOptions {
-			opts := *khttp.NewHeadersInspectionOptions()
-			opts.InspectRequestHeaders = false
-			opts.InspectResponseHeaders = true
-			return opts
-		}()),
+		// safeHeadersInspectionHandler replaces khttp.HeadersInspectionHandler
+		// to fix a data race: the upstream handler shares a single
+		// HeadersInspectionOptions (and its ResponseHeaders map) across all
+		// concurrent requests that don't carry a per-request option, causing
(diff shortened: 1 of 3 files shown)


v2/middleware/middleware.go
```diff
@@ -101,11 +101,12 @@ func GetForKiota(tfeSDKVersion string, options ...MiddlewareOption) ([]khttp.Mid
...
-			opts.InspectRequestHeaders = false
-			opts.InspectResponseHeaders = true
-			return opts
-		}()),
+		// safeHeadersInspectionHandler replaces khttp.HeadersInspectionHandler
+		// to fix a data race: the upstream handler shares a single
+		// HeadersInspectionOptions (and its ResponseHeaders map) across all
+		// concurrent requests that don't carry a per-request option, causing
+		// fatal concurrent map writes. Our implementation allocates fresh
+		// options per-request when none is provided by the caller.
```
brandonc: Too much temporaneous comment/context/naming. Let's just call this NewHeadersInspectHander

v2/middleware/safe_headers_inspection.go
```diff
@@ -0,0 +1,98 @@
...
+// embedded ResponseHeaders map, producing:
+//
+//	fatal error: concurrent map read and map write
+//	github.com/microsoft/kiota-abstractions-go.(*header).Add
+//	github.com/microsoft/kiota-http-go.HeadersInspectionHandler.Intercept
+//
+// This implementation allocates a fresh HeadersInspectionOptions for every
+// request that does not already carry one, so each goroutine writes into its
+// own private map. Per-request options provided by callers (the intended
+// consumption pattern) are honoured unchanged.
```
brandonc: Same here. The only information we need is that we try to get a NewHeadersInspectionOptions from the context, or create a fresh one with the configured options.

v2/middleware/safe_headers_inspection.go
```diff
@@ -0,0 +1,98 @@
...
+// headersInspectionReqKey mirrors the unexported headersInspectionKeyValue in
+// kiota-http-go (headers_inspection_handler.go) so that context lookups are
+// wire-compatible with any per-request HeadersInspectionOptions that a caller
+// may have stored via khttp.NewHeadersInspectionOptions().GetKey().
+var headersInspectionReqKey = abs.RequestOptionKey{Key: "nethttplibrary.HeadersInspectionOptions"}
+
+// headersInspectionOptsIface mirrors the unexported headersInspectionOptionsInt
+// interface in kiota-http-go so that we can type-assert per-request options
+// stored in the context without importing an internal type.
+type headersInspectionOptsIface interface {
```
brandonc: Hungarian notation?

ctrombley: Addressed in b7b2d4ed.

v2/middleware/safe_headers_inspection.go
ctrombley: Addressed in b7b2d4ed.

v2/middleware/middleware.go
ctrombley: Addressed in b7b2d4ed.

brandonc (approved):
Much improved! The only optional cleanup is to rename the file to something shorter, inspect_headers.go or something

PR #1422, opened 2026-08-04
middleware: set Content-Type for POST/PATCH/DELETE requests without a body

ctrombley:
## Summary

The Atlas API requires `Content-Type: application/vnd.api+json` on all mutating requests, even when the request body is empty (e.g. action endpoints like `/workspaces/{id}/actions/safe-delete`). When this header is absent, Atlas returns `415 Unsupported Media Type`.

The go-tfe v1 client set this header unconditionally for POST, PATCH, and DELETE in `NewRequestWithAdditionalQueryParams`. The Kiota-based v2 client only sets `Content-Type` when `SetContentFromParsable` (or similar) is called, so requests with no body omit it.

## Change

Add `ContentTypeMiddleware` to the default Kiota middleware pipeline. It sets `Content-Type: application/vnd.api+json` on POST, PATCH, and DELETE requests that do not already carry a `Content-Type` header. Pre-existing values (e.g. `application/octet-stream` for binary uploads) are preserved unchanged.

## Testing

Unit test in `content_type_test.go` covers all six cases: POST/PATCH/DELETE (header set), GET/PUT (header not set), and POST with a pre-existing Content-Type (header preserved).

### DIFF
--- a/v2/middleware/content_type.go
+++ b/v2/middleware/content_type.go
@@ -1,0 +1,34 @@
+// Copyright IBM Corp. 2018, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package middleware
+
+import (
+	"net/http"
+
+	khttp "github.com/microsoft/kiota-http-go"
+)
+
+const contentTypeJSONAPI = "application/vnd.api+json"
+
+// ContentTypeMiddleware ensures that POST, PATCH, and DELETE requests always
+// carry a Content-Type: application/vnd.api+json header, even when the Kiota
+// request builder did not set one because the request has no body.
+//
+// The Atlas API requires this header on all mutating requests regardless of
+// whether a body is present; omitting it results in a 415 Unsupported Media
+// Type response.  The go-tfe v1 client set this header unconditionally for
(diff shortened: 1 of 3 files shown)


v2/middleware/content_type.go
```diff
@@ -0,0 +1,34 @@
...
+
+const contentTypeJSONAPI = "application/vnd.api+json"
+
+// ContentTypeMiddleware ensures that POST, PATCH, and DELETE requests always
+// carry a Content-Type: application/vnd.api+json header, even when the Kiota
+// request builder did not set one because the request has no body.
+//
+// The Atlas API requires this header on all mutating requests regardless of
+// whether a body is present; omitting it results in a 415 Unsupported Media
+// Type response.  The go-tfe v1 client set this header unconditionally for
```
brandonc, line 20: This seems like an API bug but we need to fix it client side anyway for TFE. ¯\_(ツ)_/¯

brandonc (approved):

brandonc:
Add a CHANGELOG for this one

PR #1454, opened 2026-08-26
Return error for unsupported cost-estimation Logs function

KenCox-Hashicorp:
## Description

The `/api/v2/cost-estimates/:id/output` endpoint used by cost-estimation `Logs` is no longer supported. This changes the `Logs` function to return an unsupported error.

I regenerated the cost-estimation mocks using uber mockgen, but there was no change in the mock file. Should it be edited to always return `nil, errors.New("...")`, as the real implementation does?

The `/api/v2/cost-estimates/:id/output` route is still defined in Atlas but has no controller. The route will be removed in an Atlas PR.

Jira: https://hashicorp.atlassian.net/browse/TF-40447, https://hashicorp.atlassian.net/browse/TF-40504

## PCI review checklist

- [N/A] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [N/A] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -4238,7 +4238,8 @@ type CostEstimates interface {
 	Read(ctx context.Context, costEstimateID string) (*CostEstimate, error)
 
 	// Logs retrieves the logs of a costEstimate.
-	// Warning: This method is deprecated and will be removed from a future version of go-tfe. There is no replacement.
+
+	// Deprecated: Cost estimate logs are unsupported. There is no replacement.
 	Logs(ctx context.Context, costEstimateID string) (io.Reader, error)
 }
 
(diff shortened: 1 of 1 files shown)


v1.go
```diff
@@ -4305,45 +4306,9 @@ func (s *costEstimates) Read(ctx context.Context, costEstimateID string) (*CostE
 }
 
 // Logs retrieves the logs of a costEstimate.
+// Warning: This method is deprecated and will be removed from a future version of go-tfe. There is no replacement.
```
shwetamurali: Could we use Go’s recognized Deprecated: format here? Also, since keeping this method avoids breaking v1 consumers, I’d also avoid promising its future removal:

`// Deprecated: Cost estimate logs are unsupported. There is no replacement.`

v1.go
```diff
@@ -4963,6 +4928,11 @@ var (
 	ErrRequiredSCIMGroupMappingUpdateOps = errors.New("Update Options are required to update SCIM Group Mapping")
 )
 
+// functionality which has been removed since v1
+var (
+	ErrCostEstimatesLogsUnsupported = errors.New("CostEstimates Logs() is now unsupported")
```
shwetamurali: One small suggestion since this error will be publicly available to users of go-tfe: would you mind renaming it to `ErrUnsupportedCostEstimateLogs`? That follows the naming pattern of other errors in this package, such as `ErrUnsupportedOperations`, and will be easier for callers to discover. Go error messages also typically begin with lowercase text, so the full declaration could look like:

// ErrUnsupportedCostEstimateLogs is returned when cost estimate logs are requested.
ErrUnsupportedCostEstimateLogs = errors.New("cost estimate logs are unsupported")

shwetamurali:
Sorry, I missed that this repo’s `gocritic` configuration requires deprecation notices to be in a separate comment paragraph. Adding a blank // line before Deprecated: in both locations should fix the lint failures.

shwetamurali (approved):

PR #1469, opened 2026-09-09
[TF-40114] Remove omitempty from TagSelectorMatchingLogic in PolicySetUpdateOptions

anubhav-goel:
## Description
With omitempty set, a nil TagSelectorMatchingLogic pointer is omitted from
the PATCH body entirely, Atlas never receives an instruction to clear the
field, so the DB retains the previous value and the provider sees perpetual
drift.

Removing omitempty means a nil pointer serialises as JSON null. Atlas
assign_attributes sets tag_selector_matching_logic to nil on receipt,
the serializer returns nil, and the provider writes "" to state. Plan
is clean with no second apply required.

This change will be used by https://github.com/hashicorp/terraform-provider-tfe/pull/2197

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.


brandonc:
This is one of those rare cases where a present attribute with a null value causes rails to clear the field. In go-tfe, removing omitempty means that the update clears the field by default.

We've got a typed solution for this issue: replace *string with `jsonapi.NullableAttr[string]`, keeping  `omitempty` (Search for "jsonapi.NullableAttr[string]" to see existing usage)

This allows callers to explicitly set the value to `tfe.NullString()` when they want to clear the value, and otherwise omitempty continues to work in a way they would expect

brandonc (changes requested):
See comment- this is a dangerous breaking change because it changes all patch requests to clear the attribute, which we likely don't want.

brandonc (approved):

